PR #2, opened 2015-04-10
Add two DHT commands.

NodeGuy:


travisperson:
This all looks good, thought I noticed when playing with it that the `Type` value is just a constant. We may want to export a lookup table for this.

The values come from https://github.com/ipfs/go-ipfs/blob/master/notifications/query.go#L14. We will start to write some tests and we can keep this value in check with them. For now though, I think just defining an export that can be used would be fine till we figure out testing.

NodeGuy:
Well, I think a better long-term approach would be for the IPFS API to be RESTful and not have these weird quirks.

travisperson:
What quirks? Issue 1116 is a bug, and the issue with the `Type` value is just IPFS sending constants instead of strings. The lookup table just allows for a single source of truth when dealing with magic numbers.

Testing can be done in other places to make sure though values do not change over time by forcing new values to be added to the end.

travisperson:
Ah, after some thought and looking at the PR again, are you referring to having to perform the `Type` check on line 221 as a quirk?

travisperson:
Ping @NodeGuy

NodeGuy:
Sorry, I just got back from a workshop.

What I mean is that if we use HTTP status codes instead of something custom then we can use an existing Node request library to automatically translate them to Node conventions such as passing an error as the first argument when something goes wrong.  Much cleaner all around and will also be less cognitive overhead for every other language library writer who knows what a REST API looks like but has to ignore that knowledge and learn the semantics of IPFS's `Type` variable instead.

whyrusleeping:
if we arent returning sensible status codes that sounds more like a bug than a feature.

travisperson:
I know @jbenet talked about getting a repo setup for API development so we could have a unified spec for each implementation to work off of as well as tests that can be run against any implementation.

travisperson:
As far as the PR I think we are good to go unless anyone has anything else to add.

jbenet:
hang on, @NodeGuy has a good point here. the relevant part:

```js
return send('dht/get', key, null, null, function(err, res) {
  if (res.Type === 5)
    cb(null, res.Extra);
  else
    cb(res);
});
```

what do we need go-ipfs side to make this have proper HTTP errors? is the problem that the page says `200 OK` because it is streaming updates in as the dht is traversed, and the outcome isn't known until the end?

travisperson:
I revise my previous statement. The DHT code returns an array of of records corresponding to the DHT entry. The above will always take the second path to call `cb(res)`, as `res.Type` will never be defined.

NodeGuy:
I added two tests and would like to get this merged.  Please tell me what you think.

jbenet:
@NodeGuy this LGTM.

please note that dht put/get will change in the coming versions (0.4.0 or a bit after). as we'll be following the new records spec. records on the dht will need to be signed by the author and include a validity function. this will solve a lot of questions re: lifetime / correctness, etc.

jbenet:
@travisperson @krl LGTU? if so, let's merge

travisperson:
:+1:

PR #8, opened 2015-04-21
'name' and 'value' are required for setHeader().

travisperson:
As of node v0.12.0 (and v0.11.x), the http client uses `setHeader` on
each header key, this means if we have a key with an `undefined` or
`null` value it will try to set the header value, and not just skip it,
or use the default.

This update defaults us to `application/json`.


jbenet:
LGTM. @travisperson Ready For Merge?

PR #9, opened 2015-04-22
 Implement opts into pin remove/add for recursive

travisperson:
This add the ability to pass an opts object into `pin.add` and
`pin.remove` so the recursive option can be used.

Example:

```javascript
ipfs.pin.remove(<hash>, {r:true}, console.log);
```

Addresses #9


jbenet:
LGTM

PR #10, opened 2015-04-27
Remove json type from code blocks

harlantwood:
These code blocks are not literally json, so they get "invalid"-ish syntax highlighting, eg on the npm site & github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ curl 'http://localhost:5001/api/v0/add?stream-cannels=true' \
 ```
 
 **Response**
-```json
+```
 [{
     Hash: string,
     Name: string
@@ -146,7 +146,7 @@ curl 'http://localhost:5001/api/v0/object/get?arg=QmYEqnfCZp7a39Gxrgyv3qRS4MoCTG
 ```
 
 **Response**
-```json
+```
 {
     Links: [{
         Name: string,


whyrusleeping:
whats the reasoning for this?

whyrusleeping:
maybe the better solution would be to change the examples to be valid json? syntax highlighting is nice to have

jbenet:
> maybe the better solution would be to change the examples to be valid json? syntax highlighting is nice to have

:+1:

harlantwood:
+1

travisperson:
We can really do either, the reason for the invalid json is due to trying to explain the format and types that are returned via the API, you can't see it in the diff of this PR unfortunately so you'd have to go look at the file right now, but I'll copy it here for ease of access.

This isn't the code being changed, but a larger example.
```
{
    Objects: [
        { 
            Hash: string,
            Links: [{
                Name: string,
                Hash: string,
                Size: number
            }, ...]
        },
        ....
    ]
}
```

While in Javascript types have little importance, this documentation serves for the http api in general. The fact that objects are contained in arrays is probably enough to let the user know that there are multiple of them, and we could fill in the values with some fake information to bring it all up to valid JSON. The above format is just an easy way to spec out things and it's just my own style, and completely up for changing it.

jbenet:
yeah, i also tend to use `...` to denote repeated items. I think it's probably understood by most js users just from the array. (could have two objects if ever really ambiguous.) i think in this case matching the syntax highlighting is useful. not as much because highlighting is nice, but because angry highlighting is annoying:

travisperson:
Right now I'm going to merge this in, we can address this further if we want to have better examples, but for now I think just getting rid of the format will make things look nicer.

PR #14, opened 2015-05-06
add linting

krl:
I hope i'm not pissing anyone off too bad here.

Semicolon fan


jbenet:
+1 for linting, though is this using feross/standard? we should use that.

krl:
It is, i forgot to include the dev-dependency though. fixed

jbenet:
:+1:

PR #17, opened 2015-05-09
Add a way to create a Promisified client

nchabra:
Offers a new constructor to create an ipfs object which work with
promises instead of callbacks.
The arguments before the callback are unchanged.
No extra dependency is required - it optionally uses native promises,
es6-promise package or custom.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,3 +162,27 @@ curl 'http://localhost:5001/api/v0/object/get?arg=QmYEqnfCZp7a39Gxrgyv3qRS4MoCTG
 ### Pin
 
 ### Gateway
+
+## Promise support
+In order to use a promisified client, instantiate the object as follow:
+
+```javascript
+var ipfsApi = require('ipfs-api')
+var ipfs = new ipfsApi.IpfsClient('localhost', '5001')
+ipfs.add(new Buffer('Hello world!')).then(function (result) {
+    console.log('Added: ', result)
+}).catch(function (error) {
+    console.log('Something went wrong: ', error)
+})
+```
+
+By default, it will try to use native promises. If not available it will try to use
+[es6-promise](https://www.npmjs.com/package/es6-promise) package.
+
+In order to use a custom Promise implementation, do as follow:
+
+```javascript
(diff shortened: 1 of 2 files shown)


krl:
I'm all for promises, but could this not be done as a separate, more general lib?

I'm  not exactly sure where this kind of thing belongs.

jbenet:
Yeah I don't think this belongs here. We don't want to get into that debate, and this can be done by generalized libs



—

krl:
@nchabra https://github.com/petkaantonov/bluebird/blob/master/API.md#promisification

This seems to be a library capable of this, if it can be done generally, there's no point in including it in this repo, i'd rather keep it simple and uniform.

nchabra:
You guys are right. The promisification can be done outside. Sorry for the wasted time.

nchabra:
Yeah, after looking into bluebird lib. I'm starting to bite my tongue.
1) There is no way to recursively promisify nested objects like 'object', 'object', etc...
Which make the promisification code looks like this:

```javascript
bluebird.promisifyAll(ipfs);
['object', 'config', 'log', 'block', 'swarm', 'pin'].forEach(function(o) {
     bluebird.promisifyAll(ipfs[o]);
})
```

2) All methods are suffixed with 'Async' - maybe that's not a big deal
3) It's almost half a meg of dependency on a full Promise library that is not needed.

Would that home made promisify function in a separate file cut it? Or should I just shut up? :smile:
That being said, I do understand your concerns.

krl:
@nchabra One alternative would be writing your own small wrapper lib, that could just import node-ipfs-api and export a promisified version, seems if your code works, you colud make something smaller and nicer than bluebird, which other promised people might find useful!

nchabra:
Got it. I found thenify-all which works with only tiny dependencies.
I apologize if I was a pain... I just felt too strongly that support for Promises should be built in at this point. But I guess that was not the right way to show it.
Anyway, I'll submit some tests soon using this thenify-all lib.
Hopefully, I can be helpful this time :pray:

PR #18, opened 2015-05-10
release version 1.1.4

krl:
Just the fix that tperson posted about.

Don't want to keep this around broken, published it to npm as well.


jbenet:
Perfect, thanks!

PR #20, opened 2015-05-29
add slurp command

krl:
I found myself re-implementing this in multiple places, i think it deserves its own place as an api helper.

### DIFF
--- a/index.js
+++ b/index.js
@@ -154,6 +154,16 @@ module.exports = function (host, port) {
       return send('add', null, null, file, cb)
     },
     cat: argCommand('cat'),
+    slurp: function (path, cb) {
+      this.cat(path, function (err, stream) {
+        if (err) return cb(err)
+        var buf = ''
+        stream
+          .on('error', cb)
+          .on('data', function (data) { buf += data })
+          .on('end', function () { cb(null, buf) })
+      })
+    },
     ls: argCommand('ls'),
 
     config: {


index.js
```diff
@@ -154,6 +154,16 @@ module.exports = function (host, port) {
...
+    slurp: function (path, cb) {
+      this.cat(path, function (err, stream) {
+        if (err) return cb(err)
+        var buf = ''
+        stream
+          .on('error', cb)
+          .on('data', function (data) { buf += data })
+          .on('end', function () { cb(null, buf) })
+      })
+    },
```
jbenet: why not https://github.com/maxogden/concat-stream

krl: I'm not sure if it's really necessary to include a dependency when this is pretty straightforward code that doesn't have to deal with all the complexity that concat-stream handles.

jbenet: fair enough :)

jbenet: what i'm worried about is dedicating a whole command in the top level api for this. people will expect it elsewhere. i want to be able to claim that the API looks the same _everywhere_ (node-ipfs-api, the actual http api on go-ipfs, the cli, etc)

i'd define it like `_slurp` or something.

travisperson: I wouldn't put this on the "main" scope of the API like @jbenet said. It's not really part of IPFS and is more a helper function. Also, it does not respect buffers as it's going to call `toString` due to the plus operator concatenation. I don't know if this is really an issue due to the way javascript works but something we should at least look into.

krl:
ok, i'll keep this out of main.

PR #23, opened 2015-06-04
CI: add basic Travis config

harlantwood:
Adding Travis (in addition to Circle CI), so we can try both side by side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: node_js
+node_js:
+  - "0.12"
+  - "0.10"
+  - "0.8"


travisperson:
Will be interesting to see if things will work in `0.8`. I'm not sure if I'm really using anything special to streams2, but they were introduced in `0.10`.

harlantwood:
Someone wiht admin needs to go to https://travis-ci.org/profile/ipfs and "turn on" builds for this repo.

jbenet:
travis should be enabled!

PR #25, opened 2015-06-04
Look for "standard" package in node_modules

harlantwood:
(Don't require that it be globally installed)


travisperson:
Totally down with this.

PR #28, opened 2015-06-09
typos

Fil:


jbenet:
@krl @travisperson what's up with travis here?

jbenet:
thanks @Fil

travisperson:
Not sure about the sudo, I would assume that's a change to their system to which they are posting a message. 

As far as the failures I would assume its due to the ipfs node not starting. Sudo isn't used (also that error came before). I'll take a peak in a bit to see.

PR #30, opened 2015-07-14
make tests pass

### LINKED ISSUE #26: Test config should use different ports
would be nice if test config randomized ports or used different ports than default cause i cant run tests when my node is running

krl:
bumped ipfsd-ctl version

also fixes #26


jbenet:
@krl did you publish the new one?

PR #31, opened 2015-07-14
make vinyl play nice with browserify

krl:
Vinyl complained about content not being a Buffer when browserified, thus breaking ipfs.add.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,6 +7,7 @@ var multiaddr = require('multiaddr')
 var File = require('vinyl')
 var MultipartDir = require('./multipartdir.js')
 var stream = require('stream')
+var streamifier = require('streamifier')
 
 try {
   var pkg = JSON.parse(fs.readFileSync(__dirname + '/package.json'))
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -123,17 +124,24 @@ module.exports = function (host_or_multiaddr, port) {
...
+          contents: streamifier.createReadStream(file)
+        })
+      } else if (file instanceof stream.Stream) {
         file = new File({
           cwd: '/',
           base: '/',
           path: '/',
           contents: file
         })
-      }
```
jbenet: `} else {` ?

krl: goes on like:
```
else if (!file instanceof File) {
        return null
      }

      files[i] = file
    }
```

Just passthrough, strings, like local paths, should just passthrough.

jbenet: actually no, you want this to be a separate part altogether. the `} else if {` breaks it i thik

index.js
```diff
@@ -123,17 +124,24 @@ module.exports = function (host_or_multiaddr, port) {
...
-      if (file instanceof stream.Stream || Buffer.isBuffer(file)) {
+
+      if (Buffer.isBuffer(file)) {
+        file = new File({
+          cwd: '/',
+          base: '/',
+          path: '/',
+          contents: streamifier.createReadStream(file)
+        })
+      } else if (file instanceof stream.Stream) {
```
jbenet: maybe

maybe:
```js

var contents = files[i]

if (Buffer.isBuffer(contents))
    contents = streamifier.createReadStream(contents)

if (contents instanceof stream.Stream) {
  file = new File({
    cwd: '/',
    base: '/',
    path: '/',
    contents: contents
  })
}

if (!(file instanceof File)) {
  return null
}
```

PR #35, opened 2015-07-18
multipart dir now takes a stream of vinyl files

jbenet:
the test doesn't work. there's a very odd error, something
is interfering with vinylfs.src...

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,11 +1,11 @@
 'use strict'
 
 var fs = require('fs')
-var path = require('path')
 var http = require('http')
 var qs = require('querystring')
 var multiaddr = require('multiaddr')
 var File = require('vinyl')
+var vinylfs = require('vinyl-fs')
 var MultipartDir = require('./multipartdir.js')
 var stream = require('stream')
 var streamifier = require('streamifier')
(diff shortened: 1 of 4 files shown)


jbenet:
cc @travisperson @krl

index.js
```diff
@@ -124,42 +124,41 @@ module.exports = function (host_or_multiaddr, port) {
...
       } else if (file instanceof stream.Stream) {
-        file = new File({
+        fstream.push(new File({
           cwd: '/',
           base: '/',
           path: '/',
           contents: file
-        })
-      } else if (!file instanceof File) {
-        return null
```
jbenet: this seems to be so wrong. why was there a return null here? totally silent failure.

test/test.js
```diff
@@ -62,18 +62,21 @@ describe('ipfs node api', function () {
...
+  //   ipfs.add(fileName, function (err, res) {
+  //     if (err) throw err
+  //     filePathAdded = res
+  //     done()
+  //   })
+  // })
+  //
+  // it('add path', function () {
+  //   assert.equal(filePathAdded[0].Hash, 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP')
+  // })
```
jbenet: can someone with vinylfs experience figure this one out?

travisperson: `vinylfs.src` is async, so before it has ready anything to put into the fstream it's already calling multipartdir.

jbenet: @travisperson shouldn't the `pipe` and subsequent `.read()` wait until there is stuff read? i guess not.

jbenet: (yeah makes sense why not)

diasdavid: .read() in a Node.js stream returns `null` if no bytes are available. To solve this (when reading binary frames specially) I made [`async-buffered-reader`](https://www.npmjs.com/package/async-buffered-reader). But I see that in this case, there is no expectation of what is the length. Are you sure to want to use .read()? It will return any number of bytes that it has available. That is why we see a lot of times code with `.on('data', chunkAppend)` and then `.on('end', doSomething)`

index.js
krl: Looked at this again, definitely wrong, i was parsing the for as a forEach, so that the return would just keep the loop going. i'll look at it.

jbenet:
closed in favor of #43

jbenet:
(it's fixed now)

PR #39, opened 2015-08-04
fix for buffer checking failure in browserified code

whyrusleeping:
The fix was to fix vinyls checking of buffered files: https://github.com/wearefractal/vinyl/pull/58

and to also remove the streamifier library.

To get it to let me commit i had to fix some formatting on the dht api stuff.


index.js
```diff
@@ -309,18 +308,20 @@ module.exports = function (host_or_multiaddr, port) {
...
+            cb(res)
+          }
+        })
+      },
+
+      put: function (key, value, cb) {
+        return send('dht/put', [key, value], null, null, cb)
       }
-    }
+	}
```
jbenet: this doesnt seem to be using standard.

jbenet: maybe it's not in the repo yet?

jbenet:
changes LGTm

index.js
diasdavid: standard is there https://github.com/ipfs/node-ipfs-api/blob/master/package.json#L34-L38 didn't it report a error to you @whyrusleeping ?

whyrusleeping: standard is there, it didnt let me commit until i fixed a bunch of other things, not sure how this got through...

whyrusleeping:
vinyl pushed a new package version so i can update the dep from being a git reference

jbenet:
what's up with this test failing in nodejs? 

```
1) puts and gets a key value pair in the DHT
```

diasdavid:
@jbenet what I believe that means is that the DHT is slower, had to stretch the timeout to get it to pass.

```
✓ puts and gets a key value pair in the DHT (5425ms)
```

test/test.js
```diff
@@ -238,24 +238,25 @@ describe('ipfs node api', function () {
...
+      ipfs.dht.get('non-existent', function (error, value) {
+        assert(error)
+        done()
+      })
+    })
 
-  it("puts and gets a key value pair in the DHT", function (done) {
-    this.timeout(5 * 1000);
+  it('puts and gets a key value pair in the DHT', function (done) {
+    this.timeout(5 * 1000)
```
diasdavid: To make the test pass, change to 7 or 8 seconds

jbenet: oh maybe it is that the timeout in the daemon is not hitting before the timeout in the test.

we can pass the new --timeout option to the call:

```
ipfs dht put --timeout 100ms scope interplanetary
```

diasdavid: Is that available through the API as well?

jbenet: @diasdavid yeah it is

diasdavid:
Nice! :+1: It seems that it only fails on `0.10`, which I guess that with the Node.js LTS coming up (October), we can certainly leverage that as being the 1st version of Node.js that we will continue supporting in the future

diasdavid:
Woot! \o/ :dancer: just one more :)

PR #42, opened 2015-08-12
Make brfs a non-dev-dependency

krl:
Otherwise:

webui/apps/connections$ browserify -t reactify index.jsx
Error: Cannot find module 'brfs' from '/home/krl/unison/work/ipfs/webui/apps/connections/node_modules/ipfs-web-app/node_modules/ipfs-api'


krl:
i'll assume it's ok to just merge this trivial change.

diasdavid:
:+1:

PR #44, opened 2015-08-14
Code Review

diasdavid:


examples/ipfs-add.js
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-var ipfs = require('./index')('localhost', 5001)
```
jbenet: this file (ipfs-add) is more than an example. it's runnable. (could be a bin to install :) )

src/get-files-stream.js
```diff
@@ -0,0 +1,45 @@
+var File = require('vinyl')
+var vinylfs = require('vinyl-fs-that-respects-files')
+var vmps = require('vinyl-multipart-stream')
+var stream = require('stream')
+var Merge = require('merge-stream')
+
+exports = module.exports = getFilesStream
+
+function getFilesStream (files, opts) {
```
jbenet: @diasdavid did you change anything here, or can we just pull my stuff from #45 45?

src/index.js
```diff
@@ -0,0 +1,210 @@
...
+
+  function argCommand (name) {
+    return function (arg, cb) {
+      return requestAPI(name, arg, null, null, cb)
+    }
+  }
+
+  // -- Interface
+
+  self.send = requestAPI
```
jbenet: maybe we should call it `self.request` ?

src/index.js
```diff
@@ -0,0 +1,210 @@
...
+
+  self.add = function (files, opts, cb) {
+    if (typeof (opts) === 'function' && cb === undefined) {
+      cb = opts
+      opts = {}
+    }
+
+    return requestAPI('add', null, opts, files, cb)
+  }
```
jbenet: so, there's a bunch of commands that take options but that the api doesnt make this available. it may be better if we make _all_ commands capable of taking opts field, and passing it through to go-ipfs? (may be better as a PR on top of this one though.

src/request-api.js
```diff
@@ -0,0 +1,100 @@
...
+      } else {
+        data = objects
+      }
+
+      if (res.statusCode >= 400 || !res.statusCode) {
+        if (!data) data = new Error()
+        return cb(data, null)
+      }
+      return cb(null, data)
+    })
```
jbenet: wish that the cb could give us a stream with the response, because here we're buffering it all in memory first. again this is probably better as a future PR, not on this one.

jbenet:
@diasdavid this LGTM. nice breaking apart of things.

src/index.js
diasdavid: I would like that, just didn't want to break user space.

jbenet: :+1: maybe we can alias?

PR #45, opened 2015-08-14
more vinyl fun

jbenet:
commit b9e5313ab3e8064eba3551604f2974b33ce8e4af
Author: Juan Batiz-Benet <juan@benet.ai>
Date:   Fri Aug 14 22:57:26 2015 +0200

    more vinyl fun: use thing that look like streams too

    - cleaner file construction logic
    - use thing that look like streams too

commit 76068f440b9c9383da72ade8b5ab2aac6dce121d
Author: Juan Batiz-Benet <juan@benet.ai>
Date:   Fri Aug 14 20:31:03 2015 +0200

    use vinyl-fs-browser instead

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "multiaddr": "^1.0.0",
     "multipart-stream": "^2.0.0",
     "vinyl": "^0.5.1",
-    "vinyl-fs-that-respects-files": "^1.0.1",
+    "vinyl-fs-browser": "^0.1.0",
     "vinyl-multipart-stream": "^1.2.5"
   },
   "browserify": {
--- a/src/get-files-stream.js
+++ b/src/get-files-stream.js
@@ -28,18 +28,54 @@ function getFilesStream (files, opts) {
       if (opts.r || opts.recursive) {
         adder.add(vinylfs.src(file + '/**/*', {buffer: false}))
       }
-
-    } else if (Buffer.isBuffer(file)) {
-      single.push(new File({ cwd: '/', base: '/', path: '', contents: file }))
-    } else if (file instanceof stream.Stream) {
-      single.push(new File({ cwd: '/', base: '/', path: '', contents: file }))
-    } else if (file instanceof File) {
-      single.push(file)
     } else {
-      return new Error('unable to process file' + file)
+      // try to create a single vinyl file, and push it.
+      // throws if cannot use the file.
+      single.push(vinylFile(file))
     }
   }
 
   single.end()
   return adder.pipe(vmps())
+}
+
+// vinylFile tries to cast a file object to a vinyl file.
+// it's agressive. If it _cannot_ be converted to a file,
+// it returns null.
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -1,14 +1,10 @@
...
 var fs = require('fs')
-var Merge = require('merge-stream')
 var http = require('http')
 var qs = require('querystring')
 var multiaddr = require('multiaddr')
-var File = require('vinyl')
-var vinylfs = require('vinyl-fs-that-respects-files')
-var vmps = require('vinyl-multipart-stream')
-var stream = require('stream')
+var multipartFileStream = require('./files')
```
diasdavid: Keeping consistency with names helps for code readability.

Since only one function is export with the name `fileStream`:
- 'files.js' -> 'files-stream.js'
- 'var multipartFileStream = require('./files')' -> var filesStream = require('./files-stream')

index.js
```diff
@@ -308,41 +304,3 @@ module.exports = function (host_or_multiaddr, port) {
     }
   }
 }
-
-function getFileStream (files, opts) {
```
diasdavid: nice, I also had moved this function to its file on the CR branch :)

files.js
```diff
@@ -0,0 +1,81 @@
...
+
+function vinylContentsSafe (c) {
+  if (Buffer.isBuffer(c)) return c
+  if (typeof (c) === 'string') return c
+  if (c instanceof stream.Stream) return c
+  if (typeof (c.pipe) === 'function') {
+    // hey, looks like a stream. but vinyl won't detect it.
+    // pipe it to a PassThrough, and use that
+    var s = new stream.PassThrough()
+    return c.pipe(s)
```
diasdavid: Not sure if I understand why piping the stream to a PassThrough here

files.js
```diff
@@ -0,0 +1,81 @@
...
+  f.contents = vinylContentsSafe(f.contents)
+  return new File(f)
+}
+
+function vinylContentsSafe (c) {
+  if (Buffer.isBuffer(c)) return c
+  if (typeof (c) === 'string') return c
+  if (c instanceof stream.Stream) return c
+  if (typeof (c.pipe) === 'function') {
+    // hey, looks like a stream. but vinyl won't detect it.
```
diasdavid: Interesting :)

index.js
jbenet: sounds good! i just kept the name you had in #44

index.js
jbenet: :+1:

files.js
jbenet: I'm not sure, i noticed it did not like: https://www.npmjs.com/package/filereader-stream -- the fix works though.

i think vinyl is doing `stream instanceof stream.Stream`?

jbenet: is there a `Stream.isStream(s)` like `Buffer.isBuffer(buf)` ?

PR #47, opened 2015-08-20
added options + refs command

jbenet:


jbenet:
@harlantwood the node-ipfs-api module now has the refs command

PR #49, opened 2015-08-21
error trap & test for stopped daemon

### LINKED ISSUE #48: Uncaught Error: connect ECONNREFUSED
Happens when trying to use ipfs-api on a stopped daemon. In the browser, with ipfsapi.min.js, it is "Uncaught (in promise) TypeError: Failed to fetch"

Would be nice to trap this so I can prompt user to check their ipfs connection

ckeenan:
This should resolve #48


src/request-api.js
```diff
@@ -90,6 +90,10 @@ function requestAPI (path, args, opts, files, buffer, cb) {
     })
   })
 
+  req.on('error', function (err) {
+    return cb(err, null)
+  })
```
jbenet: ahh yep. :+1: 

is it guaranteed that `http.request` only calls _either_ the callback (i.e. emits `response`) or emits `error` ? (i believe so, but asking others). cc @diasdavid

test/test.js
```diff
@@ -25,12 +25,13 @@ describe('ipfs node api', function () {
...
       log('ipfs init done')
+      ipfsNode = node
 
-      node.startDaemon(function (err, ignore) {
+      ipfsNode.startDaemon(function (err, ignore) {
         if (err) throw err
         log('ipfs daemon running')
 
-        ipfs = ipfsApi(node.apiAddr)
+        ipfs = ipfsApi(ipfsNode.apiAddr)
```
jbenet: are these changes necessary? or just for readability? 

maybe i'm not seeing the js-scoping issue at hand, but afaict it is fine to do as it was?

test/test.js
```diff
@@ -285,4 +286,18 @@ describe('ipfs node api', function () {
...
+    ipfsNode.stopDaemon(function () {
+      if (!nodeStopped) {
+        nodeStopped = true
+        ipfs.id(function (err, res) {
+          assert.equal(err.code, 'ECONNREFUSED')
+          done()
+        })
+      }
+    })
+  })
```
jbenet: :+1:

src/request-api.js
diasdavid: Following https://nodejs.org/api/http.html#http_http_request_options_callback, an error is only emitted if something with the `request` goes wrong and the callback is only called if we get a response (after the request went well). So, yeah, all good :) 

```
If any error is encountered during the request (be that with DNS resolution, TCP level errors, or actual HTTP parse errors) an 'error' event is emitted on the returned request object.
```

test/test.js
ckeenan: Wanted a reference to the node so I could stop the daemon and check for the error in the last test. I can revert back to using `node.*` instead of `ipfsNode.*` for this before() block.. I thought it was more readable to stick with one variable name

jbenet: Ohh nevermind that makes sense. i hadn't connected them

jbenet:
This LGTM. :+1:  thanks for the PR @ckeenan

jbenet:
Published v2.2.1

PR #51, opened 2015-08-27
autoconfigure in browser

krl:
If host is unset, and we're in the browser, use the window.location.host to auto-configure the api.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -25,7 +25,7 @@ function IpfsAPI (host_or_multiaddr, port) {
     window && window.location) {
     var split = window.location.host.split(':')
     config.host = split[0]
-    config.port = split[1] || 80
+    config.port = split[1]
   }
 
   // -- Internal
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,14 @@ function IpfsAPI (host_or_multiaddr, port) {
     config.port = port || config.port
   }
 
+  // autoconfigure in browser
+  if (!config.host &&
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -20,6 +20,14 @@ function IpfsAPI (host_or_multiaddr, port) {
     config.port = port || config.port
   }
 
+  // autoconfigure in browser
+  if (!config.host &&
+    window && window.location) {
+    var split = window.location.host.split(':')
```
diasdavid: This looks like it works when the ipfsAPI is used from a `domain:port` url only. Do we want to use the url of the web application where the domain is being loaded for this?

src/index.js
```diff
@@ -20,6 +20,14 @@ function IpfsAPI (host_or_multiaddr, port) {
...
+    config.host = split[0]
+    config.port = split[1] || 80
```
diasdavid: Why port 80?

krl: If it runs on the gateway, it sholud default to 80, best would probably be to leave out the port somehow though.

src/index.js
krl: 'gateway.ipfs.io'.split(':') -> ['gateway.ipfs.io']

krl:
Left out the port 80, tested localy and in gateway. if port is unset, http.request defaults to local port.

jbenet:
@krl LGTM

PR #53, opened 2015-08-29
expose Buffer

krl:
In order to make add work in the browser with the standalone browser, we need to access the Buffer object, it seems importing the browser-buffer as standalone trips up vinyl somehow.

This is not ideal, but unless we can find a way to make adding files from buffers really portable, it seems ok.

The three cases are:

* node, just works
* browserify, works
* standalone module in browser, not without exposing browserifys Buffer

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -197,6 +197,8 @@ function IpfsAPI (host_or_multiaddr, port) {
     resolve: argCommand('name/resolve')
   }
 
+  self.Buffer = Buffer
+
   self.refs = argCommand('refs')
   self.refs.local = command('refs/local')
 


jbenet:
@diasdavid will know better than me how to handle this

diasdavid:
It is a pity that both Buffers are called Buffers and don't have the implementation.

Since this is not a common practice and also something not so usual to do in Browser applications, I suggest strongly that it has to be documented in the README and in case a user calls the function with a normal Browser Buffer obj, we should throw an error an point to the link in the README that explains that scenario.

One other suggestion, which I believe it would be more clean, but at the same time, less convenient, is instead of exposing the buffer-browserify through self, invite people to use https://github.com/feross/buffer, which makes a very good job explaining the differences, making this step a 'buy in' from developers, ensuring that everyone knows the differences and also that they understand this 'special' Buffer can be used for scenarios than just the node-ipfs-api.

krl:
We could also maybe change the default behavior of interpreting strings as filenames, and explicitly require Vinyl files, this way we could simply add JSON.stringify(object), without any extra headache.

I'm not getting ipfs object add to work with Buffer included from either npm browser-buffer or buffer, at least not in combination with webcomponents.

krl:
For now, i would think that exposing Buffer is not the worst way to go. It works reliably, even in webcomponents.

@diasdavid

diasdavid:
Ok :) just add some documentation to it as other people will try to understand why their browser-buffers aren't working.

jbenet:
> We could also maybe change the default behavior of interpreting strings as filenames, and explicitly require Vinyl files, this way we could simply add JSON.stringify(object), without any extra headache.

no, that changes the interface. we promise that:

```
ipfs add <path>
ipfs.add(<path>) // js
ipfs.add(<path>) // go
```
all work the same: adds the file addressed by `<path>`

diasdavid:
Oh :(, no documentation added to the README?

harlantwood:
Probably worth pushing out an NPM release for this, it's quite useful.

diasdavid:
@harlantwood you are totally right. Pushed 2.3.1 tag to npm https://github.com/ipfs/node-ipfs-api/releases/tag/v2.3.1

krl:
@diasdavid pushed 2.3.2, i think you forgot to run npm run build before!

Issue #62, opened 2015-09-18
When no options are passed, it assumes we are in browser context


diasdavid:
```
/Users/david/Documents/code/ipfs/node-ipfs-api/src/index.js:25
    window && window.location) {
    ^

ReferenceError: window is not defined
```

with: 

```
var ipfsAPI = require('./src')()
```

running on Node.js

VictorBjelkholm:
Closed by https://github.com/ipfs/node-ipfs-api/pull/63

PR #63, opened 2015-09-19
Correctly detect browser vs node environment

VictorBjelkholm:
Change conditional to not try to access the `window` object (which throws error when `window` doesn't exist), instead we check the `typeof` so it doesn't crash if `window` is not defined.

We can also safely assume that `window.location` exists if `window` exists.

This closes issue #62

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,7 +22,7 @@ function IpfsAPI (host_or_multiaddr, port) {
 
   // autoconfigure in browser
   if (!config.host &&
-    window && window.location) {
+    typeof window !== 'undefined') {
     var split = window.location.host.split(':')
     config.host = split[0]
     config.port = split[1]


VictorBjelkholm:
Ok, I've run the tests locally, all of them are passing. In CircleCI, `ls` is stepping over the timeout, so the tests are failing. In Travis, we get some cryptic message `TypeError: Cannot read property 'pkg' of null` which I have no idea why that happens. @diasdavid care to take a look?

VictorBjelkholm:
Ok, was able to reproduce the travis problem after updating dependencies.

It seems that commit 5c12f11d784498add132b3ebedefcd0dbc23ae2b in ipfs/node-ipfsd-ctl broke the module for both Mac and what Travis CI is using. What's weird is that master for this project is passing but only on Travis. Locally the build is broken, wondering if it's some Travis cache. What I can see, ipfs/node-ipfsd-ctl is also passing on master locally for me, but not in this project.

VictorBjelkholm:
Submitted PR #64 to fix the tests which will make this PR green as well.

RichardLitt:
:+1: Good catch.

PR #64, opened 2015-09-25
Add go-ipfs and ipfsd-ctl to fix tests

VictorBjelkholm:
Tests were broken due to change in `node-ipfsd-ctl`. This PR fixes that issue by setting `go-ipfs` as a direct dependency.

See this issue: https://github.com/ipfs/node-ipfsd-ctl/issues/20

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,6 @@
   "main": "src/index.js",
   "dependencies": {
     "brfs": "^1.4.0",
-    "go-ipfs": "^0.3.7",
-    "ipfsd-ctl": "^0.5.0",
     "merge-stream": "^1.0.0",
     "multiaddr": "^1.0.0",
     "multipart-stream": "^2.0.0",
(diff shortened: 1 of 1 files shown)


diasdavid:
Thanks @VictorBjelkholm ! This seems like to be the solution, but also a quick patch

It seems that @bcomnes is checking a solution that doesn't require the user to install a different version of go-ipfs - https://github.com/ipfs/node-ipfsd-ctl/issues/20#issuecomment-142485055

diasdavid:
This should solve the issue: https://github.com/ipfs/node-ipfs-api/pull/65

From IRC:
15:13 <+daviddias> VictorBjelkholm: since ipfsd-ctl PR from bcomnes solves the issue of finding 'go-ipfs', we avoid needing to install it twice
15:13 <+daviddias> take a look at https://github.com/ipfs/node-ipfs-api/pull/65, should solve your problem too

Thank you for catching this! :)

PR #65, opened 2015-09-25
bump ipfsd-ctl version, fixes #64

diasdavid:


diasdavid:
Seems the DHT is kind of slow at this moment, in my machine:  
```
✓ ls (1957ms)
```

VictorBjelkholm:
Retried the tests until they passed, this seems ready to be merged.

diasdavid:
woot! :) released 2.3.3

jbenet:
> Seems the DHT is kind of slow at this moment, in my machine:  
> ```
> ✓ ls (1957ms)
> ```

2s for a dht is not bad. i want to get it down to 1s, but i dont think we'll have lookups faster than 1s in say, a 10M node net. (though yes, with a much smaller networks like ours it shouldn't be 2s)

PR #67, opened 2015-09-26
add opts param to IpfsAPI, allow 'protocol' config

### LINKED ISSUE #66: Configurable http.request protocol option
Running ipfs-api from an on-disk "file:///.../index.html" does not work. Calling http.request without passing a `protocol` in the opts, it will default to `window.location.protocol`. This causes the following error:
`TypeError: Failed to execute 'fetch' on 'Window': Failed to parse URL from file://127.0.0.1:5001/api/...`

see `dist/ipfsapi.js:5837`

Could we add a way to explicitly set this protocol option? It could be passed as an opt in the requestAPI or included as a param of IpfsAPI. The latter might be preferable for this scenario so not to require a protocol opt for every method.. maybe it could be IpfsAPI(host_or_multiaddr, port, opts)?

ckeenan:
Should resolve #66


VictorBjelkholm:
Is there a reason why we don't just make all the arguments `opts`? Instead of having host and port on separate arguments, they would be in opts as well. Ala https://github.com/ipfs/node-ipfs-api/issues/71#issuecomment-149176269

diasdavid:
@VictorBjelkholm I like that idea, but can we make it in a way that doesn't break current usage? (when we release a ipfsAPI 3.0.0 we can make breaking changes like that) 

@ckeenan thank you for the PR, would you mind rebasing master first, so it gets a clean merge? + This seems to solve the problem also pointed out at https://github.com/ipfs/js-ipfs-api/pull/123 //cc @Dignifiedquire ?

ckeenan:
@diasdavid rebased

diasdavid:
LGTM. Just missing the documentation on the README.md.

@Dignifiedquire @VictorBjelkholm want to give your CR?

Dignifiedquire:
For this PR please add docs in readme, a test checking this works and move the build into a separate commit.

---

In general I think this is not the right approach though, what @jbenet mentioned makes much more sense we should start supporting `new API('https://127.0.0.1:5001')` and other forms of of addresses rather than forcing people to pass in the parsed values.

diasdavid:
Currently, ipfsAPI supports multiaddr https://github.com/ipfs/js-ipfs-api/blob/master/src/index.js#L19-L21, however it is expecting a `/ip4/<ip>/tcp/<port>` to use with http. Would it feel more right if it expected  `/ip4/<ip>/http/<port>`? I would tend to say yes, but might be confusing to everyone thinking on the tcp ip:port way.

Dignifiedquire:
I think we should support:

```js
'http://localhost:5001'
'/ip4/127.0.0.1/http/5001'
{ host: 'localhost', port: 5001, protocol: 'http'}
```

ckeenan:
@diasdavid @Dignifiedquire updated. Thoughts?

test/tests.js
```diff
@@ -766,4 +766,49 @@ describe('IPFS Node.js API wrapper tests', () => {
       })
     })
   })
+
+  describe('constructor parameters', () => {
```
Dignifiedquire: Can we put this into a new file, this is getting out of hand

src/index.js
```diff
@@ -20,8 +20,23 @@ function IpfsAPI (host_or_multiaddr, port) {
...
+  while (!opts && lastIndex-- > 0) {
+    opts = arguments[lastIndex]
+    if (opts) break
+  }
+
+  if (typeof opts === 'object') {
+    for (let key in opts)
+      if (config[key]) {
+        config[key] = opts[key]
+      }
```
Dignifiedquire: This could be much simplified if I understand correct

```js
config = Object.assign(config, opts)
```

src/index.js
```diff
@@ -20,8 +20,23 @@ function IpfsAPI (host_or_multiaddr, port) {
...
+      config.host = host_or_multiaddr
+      config.port = port && typeof port !== 'object' ? port : config.port
+    }
+  }
+
+  let lastIndex = arguments.length
+  while (!opts && lastIndex-- > 0) {
+    opts = arguments[lastIndex]
+    if (opts) break
+  }
```
Dignifiedquire: Not entirely clear what you are doing above could you explain please?

Dignifiedquire:
Thanks @ckeenan functionality wise this looks good, left some cr comments

src/index.js
ckeenan: It's preventing config.host or config.port from being set to the opts object, thus preserving the defaults if needed
Setting host or port to opts is not problematic if you set host or port in opts. If those aren't set the defaults are overridden with the opts object and we can't connect
e.g. so `ipfsAPI({protocol: 'http'})` will work

src/index.js
ckeenan: Ah yes thanks

ckeenan:
@Dignifiedquire I updated with your requests. For multiple test files I parse the karma config's files for the mocha task, let me know if you want that handled differently

karma.conf.js
```diff
@@ -47,7 +47,8 @@ module.exports = function (config) {
     basePath: '',
     frameworks: ['mocha'],
     files: [
-      'test/tests.js'
+      'test/tests.js',
+      'test/constructor.js'
```
Dignifiedquire: I suggest to rename the files to all in end `.spec.js`, so we can do `test/*.spec.js`

Dignifiedquire: That way you can pass the same glob to the `mocha` config in `tasks/test.js`

ckeenan: Good idea. Added

Dignifiedquire:
Looks like you need to rebase on master again sorry :(

ckeenan:
No worries @Dignifiedquire

Dignifiedquire:
Thanks a lot. LGTM. So let's wait for @diasdavid to check and then we can ship this :)

diasdavid:
This is awesome, thank you @ckeenan and @Dignifiedquire for reviewing it :) LGTM, going to merge and release

PR #74, opened 2015-10-11
Update to be compatible with IPFS 0.4.0

diasdavid:
WIP:
  - [x] files API
  - [x] ipfs object patch updates
  - [x] more tests
  - [x] works with 0.4.0
  - [ ] waiting on 0.4.0 to be merged


README.md
```diff
@@ -215,3 +215,59 @@ curl 'http://localhost:5001/api/v0/object/get?arg=QmYEqnfCZp7a39Gxrgyv3qRS4MoCTG
...
+ipfs.files.stat(<path>, function (err, res) {})
+```
+
+##### rm
+
+```JavaScript
+ipfs.files.rm(<path>, [<options>],  function (err) {})
+```
+
+For `rm -r` pass a options obj with `r: true`
```
Dignifiedquire: It would be noce to have `recursive: true` as an alias for `r: true` as that is much more readable in code segment. Also is this the only option that can be passed atm?

diasdavid: It should be able to accept any of the options that is available through the cli. 

(ref: https://github.com/ipfs/node-ipfs-api/issues/58)@VictorBjelkholm how is the new README going? Anything I could pull and use the new format to add the remaining options?

Dignifiedquire: Right, but would be nice if that's documented here :)

diasdavid: I definitely agree:)

Dignifiedquire:
@diasdavid this needs some updating I think ;)

diasdavid:
Yeah, I need to rebase, but it works quite nicely, ipfs-blob-store uses it :) Can't wait for 0.4.0 to be released for this to be part of the global set

diasdavid:
Now with all the latest and greatest fancy sparkles from 2.9.3 :)

package.json
```diff
@@ -67,6 +67,11 @@
     "lint": "gulp lint",
     "build": "gulp build"
   },
+  "standard": {
+    "ignore": [
+      "dist/*"
+    ]
+  },
```
Dignifiedquire: Not needed anymore

src/request-api.js
```diff
@@ -34,12 +34,14 @@ function onRes (buffer, cb) {
       const error = new Error(`Server responded with ${res.statusCode}`)
 
       Wreck.read(res, {json: true}, (err, payload) => {
+        console.log('WRECK')
```
Dignifiedquire: that sounds unsetteling :D

src/api/files.js
```diff
@@ -0,0 +1,25 @@
...
+
+const argCommand = require('../cmd-helpers').argCommand
+
+module.exports = send => {
+  return {
+    cp: argCommand(send, 'files/cp'),
+    ls: argCommand(send, 'files/ls'),
+    mkdir: argCommand(send, 'files/mkdir'),
+    stat: argCommand(send, 'files/stat'),
+    rm: function (path, opts, cb) {
```
fazo96: `files.rm` is defined in such a way that requires `opts` to be passed. When recursively deleting is not necessary and `opts` is omitted, the function doesn't work, it makes an http request for rm without the `path` argument, so it's mandatory to pass an empty `{}` as opts.

Now that I think of it, how can you _not_ delete recursively? :confused:

diasdavid: @whyrusleeping can give more input here but if you have

```bash
/folder -> hash Qmabcdef
  fileA

# and then add one more file, the folder MerkleDAG node will update

/folder -> hash Qmewasdajl
  fileA
  fileB

# if you decide to rm Qmabcdef, you might not want to delete it recursively because it might delete fileA that it was pointing to
```

fazo96: Ok, so the difference is that when deleting recursively, the dag links are also deleted from the repo.

anyway, if you agree, I think we should put these lines at the start of the function:

```javascript
if (typeof opts === 'function && !cb) {
  cb = opts
  opts = {}
}
```

This way `opts` can be omitted

diasdavid: oh, for sure! I was just checking the commit history and I had that before, dunno why I removed it meanwhile, sorry and thank you for catching that :)

diasdavid: Was in the process of making this and also pulling latest changes from master to feat/files-api and realised that 0.4.0 doesn't know about ndjson stuff yet, which breaks the tests, will wait for that to land

https://github.com/ipfs/go-ipfs/pull/1381#issuecomment-165997124

diasdavid:
@whyrusleeping, `ipfs add -r` is not returning the root folder hash

**repro:**

1. git clone ...
2. npm i
3. cp $GOPATH/bin/ipfs node_modules/go-ipfs/bin/ipfs
4. npm test

note, use latest dev0.4.0

src/request-api.js
```diff
@@ -28,7 +28,7 @@ function onRes (buffer, cb) {
     const isJson = res.headers['content-type'] === 'application/json'
 
     if (res.statusCode >= 400 || !res.statusCode) {
-      const error = new Error(`Server responded with ${res.statusCode}`)
+      let error = new Error(`Server responded with ${res.statusCode}`)
```
Dignifiedquire: ??

src/request-api.js
```diff
@@ -40,6 +40,7 @@ function onRes (buffer, cb) {
         }
         cb(error)
       })
+      return
```
Dignifiedquire: ???

diasdavid: so that it doesn't continue after the `if`

src/api/files.js
diasdavid: this has been added :)

diasdavid:
woot! so we have files-api using latest 0.4.0 + https://github.com/ipfs/go-ipfs/pull/2046

@nginnever just pushed also a new dist so you can use it on the browser :)

diasdavid:
- [ ] update .add API https://github.com/ipfs/js-ipfs-api/issues/178

nginnever:
Chya! Danke! Now I don't have to learn webpack today :)

diasdavid:
Just rebased master onto this branch so that I can use latests features for ipfs-blob-store, note:

- [ ] `ipfs object patch` has a different behaviour in 0.4.0, js-ipfs-api impl needs to be updated
- [ ] `ls` is timing out with 0.4.0, @whyrusleeping any thoughts why?
- [ ] `refs` is timing out with 0.4.0, @whyrusleeping any thoughts why?

diasdavid:
wooooo it's green! :D 



//cc @haadcode @harlantwood @dignifiedquire @nginnever @RichardLitt @xicombd @VictorBjelkholm and all the people that have contributed or looked for this version for quite some time, it is here! Thank you :)

RichardLitt:
Wooo!

PR #77, opened 2015-10-18
Initial work on #57 with cross-browser + node testing [do not merge]

### LINKED ISSUE #57: Add CI for Node.js and Browser tests

VictorBjelkholm:
This is a PR work in progress, to fix #57 and to discuss the approach using testem.

Basically, I commented out the tests that are not currently working for the browser in one way or another. Removed the daemon control from inside the test and moved it out to a shell script instead.

This is all run with testem ( https://github.com/airportyh/testem ) but that can changed, important part is to make the test cases isolated from the setup the different environments need.

Issues with this: kind of slow startup, after that it works fine due to reloading of tests on filechange. Have to run different command when in CI environment.

Please let me know what you think of this and if it's worth keep working on.

**Todo**
- [ ] make sure the startup script works correctly (some issues with the daemon starting currently)
- [ ] make sure we really want to use testem vs karma (karma seems more popular+better support everywhere)
- [ ] refactor tests to pass in nodejs environment
- [ ] add saucelabs (or similar) integration
- [ ] open issue to fix this library to pass in browser environments

Current status:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 some-test.js
 npm-debug.log
+testem.log
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "merge-stream": "^1.0.0",
     "multiaddr": "^1.0.0",
     "multipart-stream": "^2.0.0",
+    "testem": "^0.9.8",
     "vinyl": "^0.5.1",
     "vinyl-fs-browser": "^2.1.1-1",
     "vinyl-multipart-stream": "^1.2.6"
@@ -33,7 +34,7 @@
     "uglify-js": "^2.4.24"
   },
   "scripts": {
-    "test": "./node_modules/.bin/mocha",
+    "test": "./node_modules/.bin/testem",
     "lint": "./node_modules/.bin/standard",
     "build": "./node_modules/.bin/browserify -t brfs -s ipfsAPI -e ./src/index.js | tee dist/ipfsapi.js | ./node_modules/.bin/uglifyjs -m > dist/ipfsapi.min.js"
   },
--- a/test/build_and_start_daemon.sh
+++ b/test/build_and_start_daemon.sh
@@ -1,0 +1,4 @@
+#! /bin/bash
+API_ORIGIN="*" ipfs daemon&
+sleep 5
+./node_modules/.bin/browserify test/test.js -o browserified.js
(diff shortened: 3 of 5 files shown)


diasdavid:
nice! Looking forward to this :) 

Saw that the tool you picked has support for SauceLabs, which is awesome because we definitely will want CI :)

Dignifiedquire:
noooooooo.. you have to use [karma](https://github.com/karma-runner/karma) :stuck_out_tongue_winking_eye:

Dignifiedquire:
In all seriousness, the benefit of using karma would be that I'm the core maintainer of it and you have direct line to the source so to speak ;)

VictorBjelkholm:
@diasdavid thanks, you think the approach is right? Yeah, that was something I was looking for from the beginning.

@Dignifiedquire Haha, yeah, Karma is good, used it in the past and would be trivial to port it, as long as the tests are isolated and can run in both browser+node. Have some issues with testem regarding starting the ipfs daemon before running the tests so might give Karma a shot instead... We'll see :)

Dignifiedquire:
> the tests are isolated and can run in both browser+node.

That shouldn't be a problem as long as the tests are written in a way that supports that

> with testem regarding starting the ipfs daemon before running the tests so might give Karma a shot instead

Let me know if you have any karma specific questions, happy to help out :)

VictorBjelkholm:
@diasdavid added a todo in the description of the PR, anything else I should do related to the tests?

Note to self regarding using karma
```
[11:23:24]  <dignifiedquire>	VictorBjelkholm: I know how you could solve it with karma, you can tell karma to serve arbitrary files and not include them, so that you can load them via simple http request, see http://karma-runner.github.io/0.13/config/files.html for more details on how to configure that. Then you could write a helper method, `getFile` which detects if it's run in the browser and then uses an http request to fetch the file or if it's run under
[11:23:24]  <dignifiedquire>	and fetch the file va fs.readFile
```

VictorBjelkholm:
PR #78 seems a looks a lot better and in my limited testing, works better as well. We should probably close this and focus on that one instead.

diasdavid:
Thank you for figuring things out and joining forces @VictorBjelkholm and @Dignifiedquire :D :D Closing this one to focus on https://github.com/ipfs/node-ipfs-api/pull/78

PR #81, opened 2015-10-21
Make solid tests for node-ipfs-api

diasdavid:
This endeavour was started and pushed forward by @Dignifiedquire and @VictorBjelkholm (<3 ref: https://github.com/ipfs/node-ipfs-api/pull/78 https://github.com/ipfs/node-ipfs-api/pull/77), but we welcome anyone who wants to contribute.

Goal: Have a complete test suit for node-ipfs-api that test both in Node.js and in the Browser.
Stretch goal: Have code coverage checking

This PR will be focused on making tests for IPFS version 0.3.7 (version 0.3.8 will require an update on the request-api.js parser)

Historically, we have had problems because tests are using the global network to perform the DHT queries, this makes testing slow and somewhat unreliable. Tests should their own IPFS cluster, for testing against the global network, we must have _integration_ tests, but that can be the next step.

### API

- [x] before tests run
 - [x] start 3 disposable IPFS nodes, connect them in a triangle (needs: https://github.com/ipfs/node-ipfsd-ctl/issues/24 https://github.com/ipfs/node-ipfsd-ctl/pull/25)
- [x] .send
  - [x] node.js
  - [x] browser
- [ ] .add
  - [ ] .add a file
    - [x] node.js
    - [ ] browser
  - [ ] .add a buffer
    - [x] node.js
    - [ ] browser
  - [ ] .add a path
    - [x] node.js
    - [ ] browser
  - [ ] .add a folder (nested)
    - [x] node.js
    - [ ] browser
- [ ] .cat
  - [x] node.js
  - [ ] browser
- [ ] .ls
  - [x] node.js
  - [ ] browser
- [ ] .config
  - [ ] .config.get
    - [x] node.js
    - [ ] browser
  - [ ] .config.set
    - [x] node.js
    - [ ] browser
  - [ ] .config.show
    - [x] node.js
    - [ ] browser
  - [ ] .config.replace
    - [x] node.js
    - [ ] browser
- [ ] ~~.update~~ (Currently disabled, wait for version 0.4.0 release)
- [ ] .version
  - [x] node.js
  - [ ] browser
- [ ] .commands
  - [x] node.js
  - [ ] browser
- [ ] ~~.mount~~ (Requires FUSE, not practical for testing in node-ipfs-api)
- [ ] .diag
  - [ ] .diag.net
    - [x] node.js
    - [ ] browser
- [ ] .block
  - [ ] .block.get
    - [x] node.js
    - [ ] browser
  - [ ] .block.put
    - [x] node.js
[...]

PR #83, opened 2015-10-24
[discuss] Switch to superagent instead of raw http.request

### LINKED ISSUE #69: Sometimes ECONNRESET Error is thrown
Only sometimes I get this error on `ipfs.ls(<hash>, cb )` and sometimes on `ipfs.pin`. In special when a set of executions is run.

```
I20151004-17:56:25.538(2)? catched error on ipfs ls { [Error: read ECONNRESET] code: 'ECONNRESET', errno: 'ECONNRESET', syscall: 'read' }
W20151004-17:56:25.539(2)? (STDERR) Possibly unhandled Error: read ECONNRESET
W20151004-17:56:25.539(2)? (STDERR)     at errnoException (net.js:905:11)
W20151004-17:56:25.539(2)? (STDERR)     at TCP.onread (net.js:559:19)
```

Any Idea what this could be?

Dignifiedquire:
@diasdavid Ref #69 

### Benefits

* Fixes #69 for me
* Easier to read code
* Request logging when `$DEBUG=true`

### Changes

* No support for `vinyl` currently


VictorBjelkholm:
Would the usage always be like this? (Re `res[0]`)
```
    ipfs.id(function (err, res) {
       if (err) throw err
       if (err) throw err
      var id = res[0]
```

In that case, I would suggest doing that before returning, so people wouldn't have to duplicate the grabbing of the first element of res in every single api call.

Dignifiedquire:
@VictorBjelkholm that is only temporary, the `res[0]` is very annoying atm but wanted to get sth up to see if it works as expected.

Re tests for request-api itself, I think when we have the tests from #81 we should be fine

Dignifiedquire:
@VictorBjelkholm fixed

PR #92, opened 2015-10-29
[discuss] Switch to superagent instead of raw http.request

Dignifiedquire:
So I have gotten superagent to work finally. But I sadly can not suggest to merge this. The reason being that the superagent browser part is much worse than the node part. 
My general feeling is still that we should not use `http.request` directly but rather do some research into what request modules are out there and do a good job of what we need them to do in the browser as well as in node.

### DIFF
--- a/src/get-files-stream.js
+++ b/src/get-files-stream.js
@@ -8,7 +8,6 @@ exports = module.exports = getFilesStream
 
 function getFilesStream (files, opts) {
   if (!files) return null
-  if (!Array.isArray(files)) files = [files]
 
   // merge all inputs into one stream
   var adder = new Merge()
@@ -31,7 +30,7 @@ function getFilesStream (files, opts) {
       adder.add(vinylfs.src(file, srcOpts))
 
       // if recursive, glob the contents
-      if (opts.r || opts.recursive) {
+      if (opts.recursive) {
         adder.add(vinylfs.src(file + '/**/*', srcOpts))
       }
     } else {
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -1,108 +1,123 @@
-var http = require('http')
-var qs = require('querystring')
-var getFilesStream = require('./get-files-stream')
+'use strict'
+
+const request = require('request')
+const getFilesStream = require('./get-files-stream')
+
+const isNode = !global.window
+
+// -- Internal
+
+function onEnd (buffer, result, cb) {
+  return (err, res, body) => {
+    if (err) {
+      return cb(err)
+    }
+
+    if (res.statusCode >= 400 || !res.statusCode) {
+      return cb(new Error(`Server responded with ${res.statusCode}: ${body}`))
+    }
+
+    if ((result.stream && !buffer) ||
+        (result.chunkedObjects && buffer)) {
+      return cb(null, body)
+    }
+
+    if (result.chunkedObjects) return cb(null, result.objects)
+
+    try {
+      const parsedBody = JSON.parse(body)
+      cb(null, parsedBody)
+    } catch (e) {
+      cb(null, body)
+    }
+  }
+}
+
+function onData (result) {
+  return chunk => {
+    if (!result.chunkedObjects) return
+
+    try {
+      const obj = JSON.parse(chunk.toString())
+      result.objects.push(obj)
+    } catch (e) {
+      result.chunkedObjects = false
+    }
+  }
+}
+
+function onResponse (result) {
+  return res => {
+    result.stream = !!res.headers['x-stream-output']
+    result.chunkedObjects = !!res.headers['x-chunked-output']
+  }
+}
+
+function makeRequest (opts, buffer, cb) {
+  // this option is only used internally, not passed to daemon
+  delete opts.qs.followSymlinks
+
+  const result = {
+    stream: false,
+    chunkedObjects: false,
+    objects: []
+  }
+
+  return request(opts, onEnd(buffer, result, cb))
+    .on('data', onData(result))
+    .on('response', onResponse(result))
+}
+
+function requestAPI (config, path, args, qs, files, buffer, cb) {
+  qs = qs || {}
+  if (Array.isArray(path)) path = path.join('/')
+  if (args && !Array.isArray(args)) args = [args]
+  if (args) qs.arg = args
+  if (files && !Array.isArray(files)) files = [files]
+
+  if (typeof buffer === 'function') {
+    cb = buffer
+    buffer = false
+  }
+
+  if (qs.r) qs.recursive = qs.r
+
+  if (!isNode && qs.recursive && path === 'add') {
+    return cb(new Error('Recursive uploads are not supported in the browser'))
+  }
+
+  qs['stream-channels'] = true
+
+  const opts = {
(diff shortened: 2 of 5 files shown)


VictorBjelkholm:
As I've said before, we should really look into fetch that exists natively
I'm browsers, have good shims for older browsers and node as well.

Dignifiedquire:
`fetch` does look interesting, this looks like a good polyfill https://github.com/matthew-andrews/isomorphic-fetch and also browser support looks quite decent http://caniuse.com/#search=fetch.

The other alternative that I know of that is widely used is https://github.com/request/request.

diasdavid:
IIRC, this discussion started with the goal of fixing https://github.com/ipfs/js-ipfs-api/issues/69. From that, independently of the module or ES6/ES7 feature we use, do we have a new working solution, that doesn't loose any of the properties we have so far, such as recursive folder uploading, and that fixes that error?

If yes, let's move to that one. If not we should look into one. Our rule of thumb should be fixing bugs without breaking user space, and since the `request-api.js` component is abstracted by `js-ipfs-api`, we can change modules internally as we please and want to experiment things, that is super ok and welcome :)

Dignifiedquire:
> that doesn't loose any of the properties we have so far, such as recursive folder uploading, and that fixes that error?

Yes we have, this PR, but it only works with node.js the browser part is not working properly that's why I opened it with a discussion context.

Dignifiedquire:
@jbenet @mappum I need some clarification on the capabilities of what you hacked together with vinyl-fs. As far as I can understand it supports recursive file uploads from nodejs but not the browser, as in the browser there is no filesystem from which we could upload, so there you can upload file objects or buffers, or strings, with the proper metadata, is that correct?

jbenet:
the browser can have a stream of (vinyl or other) files just the same,
representing deep hierarchies.

>

Dignifiedquire:
@diasdavid please need some CR here, I think this is ready to be merged now, it resolves the issues from #69 which was the main goal but keeps full vinyl support. All tests are passing in node and the browser (as far as they were passing before). It's using [request](https://github.com/request/request) as it's the only one of the discussed modules that has proper streaming support so we can continue using `get-files-stream`.

src/get-files-stream.js
```diff
@@ -31,7 +30,7 @@ function getFilesStream (files, opts) {
       adder.add(vinylfs.src(file, srcOpts))
 
       // if recursive, glob the contents
-      if (opts.r || opts.recursive) {
+      if (opts.recursive) {
```
diasdavid: why no more `r`?

src/request-api.js
```diff
@@ -1,108 +1,123 @@
-var http = require('http')
-var qs = require('querystring')
-var getFilesStream = require('./get-files-stream')
+'use strict'
```
diasdavid: I've read that this is required for 'let' to work properly, is it the reason why you added?

Dignifiedquire: yes

src/get-files-stream.js
Dignifiedquire: I'm merging the `r` option in the configuration setup, so later down the line it's only one option to check.

diasdavid: got it, nice :)

diasdavid:
> it resolves the issues from #69 which was the main goal but keeps full vinyl support. All tests are passing in node and the browser (as far as they were passing before).

Sweet! Would you mind rebasing master onto this one? We just updated it to work with 0.3.9

Dignifiedquire:
@diasdavid done

diasdavid:
woot ! :D

diasdavid:
@Dignifiedquire seems that now 'cat' always returns a 'string', even in Node.js, where it should be a stream

jbenet:
it should be a stream in the browser too.

Dignifiedquire:
@jbenet it's not possible to provide a real stream, as all browser requests are based on XHR, which has no concept of real streaming. So we can wrap the finished response in a stream, but not actually stream the data.

diasdavid:
I believe what @jbenet meant is that the interface should be the same in both lands and therefore, if it returns a stream in one, should return a stream in the other.

Dignifiedquire:
It makes sense to have that as a goal, but this would be a breaking api change in the browser.

jbenet:
> I believe what @jbenet meant is that the interface should be the same in both lands and therefore, if it returns a stream in one, should return a stream in the other.

yes i meant this o/.

also, seriously? XHR cannot stream _huge_ requests? does it have to support range queries then? (i believe we may already support these because seeking in videos works fine...

diasdavid:
We were checking if the response was a stream before https://github.com/ipfs/js-ipfs-api/blob/cc9ec5138ab57225ba610434422c95bc65a26bc6/src/request-api.js#L56 and now we are glueing the chunks altogether and returning the thing.

> also, seriously? XHR cannot stream huge requests? does it have to support range queries then? (i believe we may already support these because seeking in videos works fine...

Yeah, AFAIK, you always have to do some chunking, range or pagination magic. There is however, a [stream-browserify](https://www.npmjs.com/package/stream-browserify) module to offer Node.js streams in the browser, so that might have been what you had seen before (since we always browserified the code)

Dignifiedquire:
> XHR cannot stream huge requests? does it have to support range queries then? (i believe we may already support these because seeking in videos works fine...

There are issues at least around streaming, I think XHR2 allows for streaming the response though. Will need to do better research though. 



> now we are glueing the chunks altogether and returning the thing.

Yes, that was an oversight on my side, but should be easily fixable, by piping the response into a stream and returning that.

diasdavid:
@Dignifiedquire I'm trying to understand at the moment where in `request`can I check the header first and decide if I want to use the stream or have access to the body by itself

Dignifiedquire:
@diasdavid this should explain it

```js
  var request = require('request')
  request(
    { method: 'GET'
    , uri: 'http://www.google.com'
    , gzip: true
    }
  , function (error, response, body) {
      // body is the decompressed response body
      console.log('server encoded the data as: ' + (response.headers['content-encoding'] || 'identity'))
      console.log('the decoded data is: ' + body)
    }
  ).on('data', function(data) {
    // decompressed data as it is received
    console.log('decoded chunk: ' + data)
  })
  .on('response', function(response) {
    // unmodified http.IncomingMessage object
    response.on('data', function(data) {
      // compressed data as it is received
      console.log('received ' + data.length + ' bytes of compressed data')
    })
  })
```

diasdavid:
The stream is in here https://github.com/ipfs/js-ipfs-api/blob/master/src/request-api.js#L67, but I only learn at https://github.com/ipfs/js-ipfs-api/blob/master/src/request-api.js#L51 if I should return as a stream or if I should parse the chunks. And it is not even returning, it is calling the cb with cb(null, stream)

Trying to think of a clean solution of how to patch this.

Dignifiedquire:
@diasdavid let me do the patch, I think I know how to do it

diasdavid:
awesome, I need to be absent for the next 1:30h, so feel free to push it :) https://github.com/ipfs/js-ipfs-api/pull/99

diasdavid:
fixed on https://github.com/ipfs/js-ipfs-api/pull/99 :)

PR #95, opened 2015-10-31
test: check + print version

jbenet:
this PR adds a quick check on the version output and prints out the version number (to show what version we're testing against. useful to test new versions)


diasdavid:
:+1:

PR #99, opened 2015-11-02
.cat is not returning a stream anymore

diasdavid:
https://github.com/ipfs/js-ipfs-api/pull/92 introduced a bug where in Node.js, the result no longer comes as a stream. This happened because the .cat test was not properly differentiating between node and browser execution.

This PR is a WIP, the test is now checking correctly the behaviour.

### DIFF
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -2,12 +2,13 @@
 
 const request = require('request')
 const getFilesStream = require('./get-files-stream')
+const stream = require('stream')
 
 const isNode = !global.window
 
 // -- Internal
 
-function onEnd (buffer, result, cb) {
+function onEnd (buffer, result, passThrough, cb) {
   return (err, res, body) => {
     if (err) {
       return cb(err)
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
Sorry 'bout that. Can you check the other commands to be sure we don't have the same issue somewhere else please?

VictorBjelkholm:
We do have the same issue elsewhere, what I remember from the top of my
head, findprovs does the same.

diasdavid:
I found it cause for the files API I was explicitly expecting a stream. Still going through this PR

Dignifiedquire:
@diasdavid thanks, you just did what I wasn't able to in my time :)

diasdavid:
@Dignifiedquire sorry for having gone ahead and make this, I needed for other things and since I remembered you telling me that you would be 'out' for the week, I figured it out. Thank you helping identifying the issue :)

Dignifiedquire:
No worries, happy you did it :) 

> Thank you helping identifying the issue :)

Well I created it in the first place, so I'd better do :P

PR #103, opened 2015-11-04
Add object.patch to the API

haadcode:


diasdavid:
sweet! thank you @haadcode

bergie:
Thanks @haadcode !

PR #102, opened 2015-11-03
Add current work in progress of readme.md and api.md

VictorBjelkholm:
**Closed in favor of https://github.com/ipfs/js-ipfs-api/pull/104 that has the branch upstream**

This is my current WIP of having a better readme and also added the API docs.

Missing currently:
- [ ] Examples of all the calls
- [ ] Better description of how to use the methods
- [ ] Probably update the TOC
- [ ] Check links et al
- [ ] Proofread

At first, tried to automatically generate the documentation with a script combining the output of `go-ipfs` and reading the method fingerprint from `js-ipfs-api` but got a bit stuck and think I tried to overengineer a solution for something that would be simpler to fix manually.

Related to #58
Ping @RichardLitt

Issue #58, opened 2015-09-15
Finish and improve the documentation


diasdavid:

VictorBjelkholm:
I'm currently redoing the readme a bit. You could assign me if you please.

diasdavid:
awesome! Thank you @VictorBjelkholm 

@jbenet is the one able to assign people on orgs/repos

jbenet:
@VictorBjelkholm ok you are part of the github node-ipfs team now. please do not merge things, or push straight to master (unless you maintain the project) please use a branch on the same repo and issue a PR and get the repo maintainer to review and then merge it

jbenet:
Actually, i added you directly to this repo instead, as you need more access will bump it up

jbenet:
@diasdavid you now have admin on all these repos so you should be able to add people.

diasdavid:
Thanks @jbenet :)

diasdavid:
Note on: https://github.com/ipfs/node-ipfs-api/issues/33#issuecomment-136978747 to be added to the docs

RichardLitt:
@VictorBjelkholm How's this going? Is there anything I can help with?

VictorBjelkholm:
@RichardLitt Hey! Yes, of course. I've made the structure to be proper but have yet to fill out all the details with calls. Take a look in my branch (that currently have to be rebased from master but have yet to get time to do that) https://github.com/VictorBjelkholm/node-ipfs-api/tree/better-readme

Basically, the only thing I've done is added a ToC, changed the structure and made the wording a bit better.

Dignifiedquire:
Just a note, might make sense to have a separate `api.md` document with those details and just a link from the readme, to keep readme at a reasonable length.

VictorBjelkholm:
That sounds much more reasonable actually. Maybe just a quick introduction
in the readme and refer to them from it.

Sincerely,
Victor Bjelkholm
(+34) 672 15 90 89

RichardLitt:
@VictorBjelkholm Cool! Maybe rebase ,switch it to `api.md`, make sure no current information is lost, and then form a PR? That way we can hypothetically do it piecemeal and faster.

diasdavid:
@VictorBjelkholm any updates on this? It might help to open a branch on this repo for the docs and then open a PR, so that you and @RichardLitt can collaborate (without having to PR to the other private PR)?

VictorBjelkholm:
[...]

PR #104, opened 2015-11-06
Add current work in progress of readme.md and api.md

VictorBjelkholm:
This is my current WIP of having a better readme and also added the API docs.

Missing currently:
- [ ] Examples of all the calls
- [ ] Better description of how to use the methods
- [ ] Probably update the TOC
- [ ] Check links et al
- [ ] Proofread

At first, tried to automatically generate the documentation with a script combining the output of `go-ipfs` and reading the method fingerprint from `js-ipfs-api` but got a bit stuck and think I tried to overengineer a solution for something that would be simpler to fix manually.

Related to #58 

Was https://github.com/ipfs/js-ipfs-api/pull/102 before but moved branch upstream.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1,0 +1,334 @@
+# js-ipfs-api
+
+## API Reference
+
+- [add](#add)
+- [cat](#cat)
+- [ls](#ls)
+- [ID](#id)
+- [version](#version)
+- [commands](#commands)
+- [ping](#ping)
+- [config](#config)
+- [update](#update)
+- [diag](#diag)
+- [block](#block)
+- [object](#object)
+- [swarm](#swarm)
+- [pin](#pin)
+- [gateway](#gateway)
+- [log](#log-1)
+- [name](#name)
+- [dht](#dht)
+
+## add
+`ipfs_client.add(file/files, callback)`
+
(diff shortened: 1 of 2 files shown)


RichardLitt:
- [ ] List options for each call. (Hope, not a commitment).

diasdavid:
@VictorBjelkholm @RichardLitt @Dignifiedquire Can I get an update on this? What is blocking us from finishing the documentation? Is there a decision between using manual vs generated docs from mocha (https://github.com/ipfs/js-ipfs-api/pull/105) ?

Dignifiedquire:
Looking into [jsdoc@3](https://github.com/jsdoc3/jsdoc) currently. Looks really good so far as I think auto generated docs are much better than manual kept ones as they just tend too much to get out of sync with the code but the test docs generator doesn't give any abilities to do good customizations

diasdavid:
As discussed on the last sprint meeting, we are going with https://github.com/ipfs/js-ipfs-api/pull/105/files which offers really good documentation compared to the inexistent that we have now, and eventually, we will have better docs for IPFS API itself. Closing this one for now :)

RichardLitt:
Thanks @diasdavid  and @VictorBjelkholm

PR #106, opened 2015-11-07
circle config

harlantwood:
Use latest stable node version on Circle CI.  Fixes the build.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,3 @@
+machine:
+  node:
+    version: stable


VictorBjelkholm:
Can we lock this to a specific version instead? I know CircleCI uses 0.10.22 but I'm not sure what stable would be for them. Should be version 5 but CircleCI only comes with version 4 and below pre-installed.

harlantwood:
they use nvm; version 5 at the moment.  what would you like to lock it to?  4.0?  4.1?  5?

VictorBjelkholm:
Ah, did not know that. No, I think we can lock it to version 5. @diasdavid @Dignifiedquire opinions?

harlantwood:
I submitted #108 for travis to test multiple node versions (not so pretty to do that under Circle).  For Circle, I'd suggest we leave it with `stable`, so we're always testing the latest.  Happy to do whatever the more active developers among us prefer though...

Dignifiedquire:
:+1:

Dignifiedquire:
@diasdavid Why are we using circle ci in the first place if it is testing less than travis?

harlantwood:
We set up both circle and Travis to see which we preferred. I originally preferred circle, but now prefer Travis, precisely for its elegance in testing under multiple node versions. I am fine with removing circle, but someone suggested yesterday in an infrastructure issue that it may be useful to test under both, with the idea that sometimes a build fails under one but not the other. I would prefer that we standardize on just one CI system, or at least one per language.

harlantwood:
cc @diasdavid re: https://github.com/ipfs/community/issues/45#issuecomment-154477825

diasdavid:
Should we change that decision? I don't necessarily see any harm of leaving as it is.

harlantwood:
https://github.com/ipfs/community/issues/45#issuecomment-154768152

PR #108, opened 2015-11-07
test under multiple node versions

harlantwood:
I suggest we test under node 4 and 5, and also stable, so when new node versions come out, we would be alerted by a broken build if tests failed there.

Thoughts @diasdavid @Dignifiedquire @VictorBjelkholm?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: node_js
 node_js:
   - "4.0"
+  - '4.2'
+  - '5.0'
+  - stable
 
 before_script:
   - export DISPLAY=:99.0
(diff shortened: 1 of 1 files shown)


harlantwood:
eg: https://travis-ci.org/ipfs/js-ipfs-api/builds/89766352

Dignifiedquire:
I suggest using `4` and `5`. as that way it always uses the latest minor version of that release cycle.

harlantwood:
Nice, will update.

Dignifiedquire:
:+1:

diasdavid:
Looks good:) thank you!

PR #115, opened 2015-11-14
fix: Move build to gulp

### LINKED ISSUE #113: dist/{ipfsapi.js, ipfsapi.min.js} not actually compiled
The files `dist/ipfsapi.min.js` and `dist/ipfsapi.js` are not containing the libraries code anymore, since commit f6a7da682f205b13424dfd6082f3eaae51af7a73. The only thing I can see in the files is the standard browserify code, without anything about ipfs. Seems like we're missing to source the correct files in the build-step.

Dignifiedquire:
Fixes #113


VictorBjelkholm:
Sweet! LGTM :+1:

PR #120, opened 2015-11-16
fix: Handle non json formatted errors.

Dignifiedquire:
Ref https://github.com/ipfs/go-ipfs/issues/1970


diasdavid:
sweet! But @whyrusleeping this doesn't mean that we don't want all of the errors to be JSON :)

Dignifiedquire:
@diasdavid can you pull that into your files-api branch so I can use it in the registry please?

diasdavid:
@Dignifiedquire sure! I was just looking surprised why the heck I was being unable to publish (registry-mirror was the reason :) )

PR #121, opened 2015-11-16
feat: Add urls directly.

### LINKED ISSUE #109: Add URLs directly
ipfs add in the cli can add from the filesystem (makes sense, given the cli context). I think the js api on the browser should be able to add from the web (makes sense, given the web context). 

So, i would like this to work:

```js
ipfs.add('http://i.imgur.com/YTSpZdP.gif', function(err, res) { ... })
```
I think it should work on Node, too, given that node is js and already straddles the boundaries between unix and the web.

There is also an argument to be made re adding this natively to the go-ipfs cli too. (i.e. making the cmds lib accept files on http as just other files)

Dignifiedquire:
Closes #109

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 var multiaddr = require('multiaddr')
 var getConfig = require('./config')
 var getRequestAPI = require('./request-api')
+var request = require('request')
 
 exports = module.exports = IpfsAPI
 
@@ -61,6 +62,10 @@ function IpfsAPI (host_or_multiaddr, port) {
     if (typeof (opts) === 'function' && cb === undefined) {
       cb = opts
       opts = {}
+    }
+
+    if (typeof files === 'string' && files.startsWith('http')) {
+      files = request(files)
     }
 
     return requestAPI('add', null, opts, files, cb)
--- a/src/index.js
+++ b/src/index.js
@@ -65,7 +65,10 @@ function IpfsAPI (host_or_multiaddr, port) {
     }
 
     if (typeof files === 'string' && files.startsWith('http')) {
-      files = request(files)
+      files = request.get({
+        url: files,
+        withCredentials: false
+      })
     }
 
     return requestAPI('add', null, opts, files, cb)
--- a/test/tests.js
+++ b/test/tests.js
@@ -4,6 +4,7 @@ var assert = require('assert')
 var fs = require('fs')
 var path = require('path')
 var File = require('vinyl')
+var Readable = require('stream').Readable
 
 var isNode = !global.window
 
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -1,6 +1,7 @@
 var multiaddr = require('multiaddr')
 var getConfig = require('./config')
 var getRequestAPI = require('./request-api')
+var request = require('request')
```
diasdavid: why adding request here?

src/index.js
```diff
@@ -63,6 +64,10 @@ function IpfsAPI (host_or_multiaddr, port) {
       opts = {}
     }
 
+    if (typeof files === 'string' && files.startsWith('http')) {
+      files = request(files)
```
diasdavid: This sync call will block the loop (or am I missing some ES6 thing?)

test/tests.js
```diff
@@ -145,6 +145,19 @@ describe('IPFS Node.js API wrapper tests', function () {
         }
       })
     })
+    it('adds a url', function (done) {
```
diasdavid: :+1: for adding a test :D

src/index.js
Dignifiedquire: Not sure what your alternative scenario would look like, we can't stream directly to go ipfs anyway

Dignifiedquire: The request is a stream object, so it will not block but will be piped into the new request effectively as far as I understand :)

diasdavid: `23:09 <+daviddias> dignifiedquire: interesting, so our thing understands 'strings', 'paths', 'buffers' and 'streams' too?``

Sweet!

diasdavid: Better add a test and docs for adding with a stream though, otherwise that might change in the future and then we don't realize what broke the add by url was actually the lack of add by stream

src/index.js
VictorBjelkholm: I'm not sure if request works in the browser. But the tests are testing that this actually works no?

diasdavid: Not being able to confirm, but read somewhere that request gets polifilled in browserify. Nevertheless, there is https://www.npmjs.com/package/browser-request for that :)

Dignifiedquire: @diasdavid @VictorBjelkholm no need for all that, all tests are passing in the browser, request is 100% browser compatible :)

Dignifiedquire:
@diasdavid all fixed and done, ready for consumption :)

diasdavid:
woot!! :D @jbenet , you have the thing you requested https://github.com/ipfs/js-ipfs-api/issues/109

jbenet:
Thanks I love you. ❤️❤️❤️❤️❤️❤️❤️

Cool thing coming soon.

PR #123, opened 2015-11-16
fix: Enforce http protocol for all requests

### LINKED ISSUE #66: Configurable http.request protocol option

Dignifiedquire:
Closes #66

### DIFF
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -115,6 +115,7 @@ function requestAPI (config, path, args, qs, files, buffer, cb) {
 
   const opts = {
     method: files ? 'POST' : 'GET',
+    protocol: 'http:',
     uri: `http://${config.host}:${config.port}${config['api-path']}${path}`,
     qs: qs,
     useQuerystring: true,


xicombd:
:+1:

diasdavid:
Interesting, what happens if it is https? Why do we need to enforce?

VictorBjelkholm:
Hm, this would block anyone from using js-ipfs-api on an https site, which, anyways is not working because the API is only served via http. How can we solve this? Should we solve this?

Dignifiedquire:
We are already enforcing `http` now in nodejs, but for some reason in the browser it's not picking that up from the url. We should make this in option in the long run, but for now this at least brings the browser and node on equal footing

diasdavid:
As discussed on IRC. Let's add this as an option to the creation of ipfsAPI :)

jbenet:
Users should be able to put the exact endpoint directly:

Api('http://localhost:5001/v0/api')
Api('https://user:pass@host/api')
Api('//meh/API')
Api('/ip4/1.2.2.4/tcp/5001/https/v0/api')

(We need to add support to multiaddr for DNS, http paths, and more)

This is why traditional `'host', port` params suck. Info is lost and not
passed through.

Dignifiedquire:
Closing in favor of #67

PR #128, opened 2015-11-19
Header work

### LINKED ISSUE #76: findprovs returns object or string
using the method `ipfs.dht.findprovs` in node environment does not work like one would expect. Using the following code:
```
var ipfsAPI = require('ipfs-api')

var ipfs = ipfsAPI('localhost', '5001')

const hash = "QmXFmCEhamPHvpbzQ3HMFABrpAXgE7D9jzp6QtvyC8Xp55"

ipfs.dht.findprovs(hash, (err, res) => {
	if(err) throw err
	console.log(res)
	console.log(typeof res)
	console.log("Streamable?", !!res.readable)
	console.log('So what to do with `res` here?')
})
```

I get the following output sometimes (which works correctly):
```
object
Streamable? false
So what to do with `res` here?
```
but other times, it returns this (which is neither streamable or valid json)
```
string
Streamable? false
So what to do with `res` here?
```

Originally open issue https://github.com/ipfs/go-ipfs/issues/1851 on go-ipfs but seems to be a bug in this library.

Dignifiedquire:
Goal: Fix #76 still not there though..

When https://github.com/ipfs/go-ipfs/issues/1979 is merged we can remove the config settings in the daemon setup

### DIFF
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -16,6 +16,7 @@ function onRes (buffer, cb) {
 
     const stream = !!res.headers['x-stream-output']
     const chunkedObjects = !!res.headers['x-chunked-output']
+    const isJson = res.headers['content-type'] === 'application/json'
 
     if (res.statusCode >= 400 || !res.statusCode) {
       const error = new Error(`Server responded with ${res.statusCode}`)
@@ -33,24 +34,18 @@ function onRes (buffer, cb) {
 
     if (chunkedObjects) {
       const parsed = []
-      res.on('data', chunk => parsed.push(JSON.parse(chunk)))
+      res.on('data', chunk => {
+        try {
+          parsed.push(JSON.parse(chunk))
+        } catch (err) {
+          // Browser quirks emit more than needed sometimes
+        }
+      })
       res.on('end', () => cb(null, parsed))
       return
     }
 
-    Wreck.read(res, null, (err, payload) => {
-      if (err) return cb(err)
-
-      let parsed
-
-      try {
-        parsed = JSON.parse(payload.toString())
-      } catch (err2) {
-        parsed = payload.toString()
-      }
-
-      cb(null, parsed)
-    })
+    Wreck.read(res, {json: isJson}, cb)
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/request-api.js
```diff
@@ -33,24 +34,18 @@ function onRes (buffer, cb) {
 
     if (chunkedObjects) {
       const parsed = []
-      res.on('data', chunk => parsed.push(JSON.parse(chunk)))
+      res.on('data', chunk => {
+        try {
+          parsed.push(JSON.parse(chunk))
+        } catch (err) {
+          // Browser quirks emit more than needed sometimes
```
diasdavid: `Emit more than needed`? What does this exactly mean? :)

tasks/daemons.js
```diff
@@ -57,7 +57,14 @@ function startDisposableDaemons (callback) {
             throw err
           }
 
-          ipfsNodes[key].setConfig('API', '{"HTTPHeaders": {"Access-Control-Allow-Origin": ["*"]}}', err => {
+          const headers = {
+            HTTPHeaders: {
```
diasdavid: Let's put `When ipfs/go-ipfs#1979 is merged we can remove the config settings in the daemon setup` as a TODO here and a issue pointing to it

diasdavid:
LGTM :)

src/request-api.js
Dignifiedquire: if you have 3 chunks, you sometimes get: 
* chunk1
* chunk2
* chunk3
* chunk1+chunk2 => not parsable

do not ask me why, browsers are a weird bunch..

tasks/daemons.js
Dignifiedquire: agreed

src/request-api.js
diasdavid: What I'm thinking is that "stream" and "chunked" should not exist at the same time. "chunked" is a stream made of small parts of the same files, they should not be parsed individually. "stream" in the another hand, it is the ndjson stuff :)

Dignifiedquire: Correct, but "chunked" is represented as a (node) stream where each individual emitted event is equal to one chunk, that's why we are parsing it this way

PR #129, opened 2015-11-20
feat: Add diag/sys command

Dignifiedquire:


diasdavid:
thank you :) - released on https://github.com/ipfs/js-ipfs-api/releases/tag/v2.9.1

src/index.js
```diff
@@ -123,7 +123,8 @@ function IpfsAPI (host_or_multiaddr, port) {
   }
 
   self.diag = {
-    net: command('diag/net')
+    net: command('diag/net'),
```
jbenet: maybe remove this command from the api? we'll be deprecating it

diasdavid: on 0.3.10? We can deprecate it when that IPFS version is released.

Dignifiedquire: Why is it deprecated? :( It has some nice details, is there an alternative to get this data?

PR #130, opened 2015-11-20
feat: Fix log/tail for 0.3.9

Dignifiedquire:
Works in node, will work in the browser with #128

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ const multiaddr = require('multiaddr')
 const getConfig = require('./config')
 const getRequestAPI = require('./request-api')
 const Wreck = require('wreck')
+const ndjson = require('ndjson')
 
 exports = module.exports = IpfsAPI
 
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -203,7 +204,10 @@ function IpfsAPI (host_or_multiaddr, port) {
 
   self.log = {
     tail (cb) {
-      return requestAPI('log/tail', null, {enc: 'text'}, null, true, cb)
+      requestAPI('log/tail', null, {}, null, false, (err, res) => {
```
diasdavid: There is no more need for "enc" param?

test/tests.js
```diff
@@ -640,15 +640,18 @@ describe('IPFS Node.js API wrapper tests', () => {
   })
 
   describe('.log', function () {
-    // TODO news 0.3.9 ndjson stuff
-    it.skip('.log.tail', function (done) {
+    it('.log.tail', function (done) {
```
diasdavid: sweet! more tests :D

src/index.js
Dignifiedquire: nope no need for that in 0.3.9 (it actually is ignored for `log/tail`)

PR #132, opened 2015-11-21
[WIP] jsdoc3 for docs

Dignifiedquire:


src/index.js
```diff
@@ -59,9 +65,19 @@ function IpfsAPI (host_or_multiaddr, port) {
...
-  self.add = (files, opts, cb) => {
+  /**
+   * Add a file/many files to IPFS returning the hash and name. The
+   * name value will only be set if you are actually sending a file.
+   *
+   * @public
+   * @param {} files
+   * @param {object} opts
+   * @param {function} cb
+   */
```
diasdavid: How does everyone feel about this JavaDoc'ization in order to have docs? 

It is nice to have API docs, not sure if, since we have to write them anyway, if we prefer to right them on the readme with examples or add this to all of the function calls.

Dignifiedquire: **Pros**
- With the code
- less chance of out of sync

**Cons**
- looks like Java

Dignifiedquire: ping @VictorBjelkholm @RichardLitt

RichardLitt: Jsdoc is pretty sweet, I just get annoyed at it because you end up with truly massive js files. It definitely helps out documentation, though, and I've used it before. 

The only real con is that updating it can get old. but not as old as old docs.

Dignifiedquire:
**NO* JAVA IN OUR CODE :)

RichardLitt:
ONLY ~~JAVA~~SCRIPT.

src/index.js
jbenet: my opinion: in general, while in other languages I do like godoc, i find that most successful modules I use in javascript are very small, easy to use, and have the docs/examples directly in the readme. i typically run away from jsdoc-ed modules.

PR #136, opened 2015-11-22
refactor: Split api definitions into separate files

Dignifiedquire:
As discussed @diasdavid :)


diasdavid:
one file for you, one file for you...


:D

PR #145, opened 2015-11-24
timeouts per suite

diasdavid:


Dignifiedquire:
Why not a global config in `test/setup.js` or directly passed to mocha?

test/api/add.spec.js
```diff
@@ -21,13 +21,13 @@ if (isNode) {
 }
 
 describe('.add', function () {
+  this.timeout(60000)
+
   it('add file', function (done) {
```
Dignifiedquire: Please move these `function`s to `=>` functions as the only reason I kept the inconsistency was the need for calling `this.timeout`

diasdavid: ok, sounds good

diasdavid:
I took out the local timeouts (I went through it because when I went through docs, it was only specified test and suit level timeouts https://mochajs.org/#more-information)

How can I set timeouts in karma?

diasdavid:
Now only missing:

- [ ] remove `function` key word and make @Dignifiedquire happy :)

Dignifiedquire:
mocha-node pass it to the mocha call in the guplfile as an option https://github.com/sindresorhus/gulp-mocha/blob/master/readme.md

karma: in the karma conf set client.mocha.timeout https://github.com/karma-runner/karma-mocha/blob/master/README.md

karma.conf.js
```diff
@@ -46,6 +46,11 @@ module.exports = function (config) {
   config.set({
     basePath: '',
     frameworks: ['mocha'],
+    client: {
+      mocha: {
+        timeout: 60000
```
Dignifiedquire: Let's pull this into `tasks/config.js` so that it's only defined in one place

tasks/config.js
```diff
@@ -42,7 +42,8 @@ const shared = {
     tls: '{}',
     console: '{}',
     'require-dir': '{}'
-  }
+  },
+  timeout: 60000
```
Dignifiedquire: This should not be namespaced under `webpack`, as it is not a webpack relevant config. Better would be `test = {timeout: 60000}`

PR #150, opened 2015-11-25
added ipfs.ls test spec

fazo96:
While working with the API I noticed weird behavior in `ipfs.ls`. Sometimes it calls the callback twice and some other times when there's an error the error object is undefined and the response object is the actual error.

After talking briefly with @Dignifiedquire I've been trying to write tests to test these and also normal `ipfs.ls` functionality. Please let me know about anything wrong with the PR :+1:

Looks like I couldn't reproduce the issue I was having. I'll try again if I happen to run into it, but it was probably some mistake on my end.

### DIFF
--- a/tasks/daemons.js
+++ b/tasks/daemons.js
@@ -70,11 +70,13 @@ function startDisposableDaemons (callback) {
               throw err
             }
 
-            ipfsNodes[key].startDaemon((err, ignore) => {
+            ipfsNodes[key].startDaemon((err, daemon) => {
               if (err) {
                 throw err
               }
-
+              daemon.add('test/test-folder', {r: true}, (e, r) => {
+                if (e) throw e
+              })
               apiAddrs[key] = ipfsNodes[key].apiAddr
               cb()
             })
--- a/test/api/ls.spec.js
+++ b/test/api/ls.spec.js
@@ -2,18 +2,25 @@
 
 describe('ls', function () {
   it('should correctly retrieve links', function (done) {
-    this.timeout(5000)
     apiClients['a'].ls('QmTDH2RXGn8XyDAo9YyfbZAUXwL1FCr44YJCN9HBZmL9Gj', (err, res) => {
-      if (err) {
-        throw err
-      } else done()
+      if (err === null && res !== undefined) {
+        assert(res.Objects[0].Links[0])
+        done()
+      }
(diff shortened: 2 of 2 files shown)


tasks/daemons.js
```diff
@@ -70,11 +70,13 @@ function startDisposableDaemons (callback) {
               throw err
             }
 
-            ipfsNodes[key].startDaemon((err, ignore) => {
+            ipfsNodes[key].startDaemon((err, daemon) => {
               if (err) {
                 throw err
               }
-
+              daemon.add('test/test-folder', {r: true}, (e, r) => {
```
diasdavid: This folder is added too during the `add` tests, could you use the hash from there?

diasdavid:
Thank you @fazo96 :)

I had in my memory that we had `ls` tests, in fact, it was always our sluggish test, but I see that with the refactoring, it was lost (//cc @Dignifiedquire ). You can find the old test here: https://github.com/ipfs/js-ipfs-api/blob/bdf2e38ec233707d16c6fbd55ce878cb91be1987/test/tests.js#L178-L230 . Your tests are more complete, it could only also confirm that all of the links returned are the ones expected, would you mind adding that?

Thank you!

test/api/ls.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    apiClients['a'].ls('QmTDH2RXGn8XyDAo9YyfbZAUXwL1FCr44YJCN9HBZmL9Gj', (err, res) => {
+      if (err === null && res !== undefined) {
+        assert(res.Objects[0].Links[0])
+        done()
+      }
+    })
+  })
+  it('should correctly handle a nonexisting hash', function (done) {
+    apiClients['a'].ls('surelynotavalidhashheh?', (err, res) => {
+      if (err !== null && res === undefined) {
```
Dignifiedquire: We since switched to `expect`, so you can do something like 

```js
expect(err).to.exist
expect(res).to.not.exist
```

Dignifiedquire:
Upps sorry for removing the `ls` tests :(

fazo96:
Will fix it soon :+1: Thanks!

diasdavid:
@fazo96 still update to do the last changes so that we can merge this PR?  Thank you :)

PR #153, opened 2015-11-27
Replace assert with chai

Dignifiedquire:
- [ ] Fix bug in dht that this exposes


test/api/dht.spec.js
```diff
@@ -4,29 +4,26 @@ describe('.dht', () => {
   it('returns an error when getting a non-existent key from the DHT',
      done => {
        apiClients['a'].dht.get('non-existent', {timeout: '100ms'}, (err, value) => {
-         assert(err)
+         console.log(err, value)
+         expect(err).to.be.an.instanceof(Error)
```
diasdavid: I believe the test fails because the err passed is not in fact an Error object, but just some message.

Dignifiedquire:
LGTM thanks @diasdavid

Issue #159, opened 2015-12-06
Broken if not loaded using webpack due to `require.context`


alexmingoia:
[require.context](https://github.com/ipfs/js-ipfs-api/blob/master/src/load-commands.js#L9) is not available unless transpiled using webpack. This breaks any module that depends on ipfs-api and uses another loader such as browserify, such as webui.

Dignifiedquire:
If you don't want to transpile using webpack, why not just load the prebuilt files in `dist`?

diasdavid:
@Dignifiedquire being browserify compatible, or in another words, maintaining that available for the developers that have been using js-ipfs-api with browserify from the beginning, it is kind of important and nice to have.

Seems like @alexmingoia already fixed it https://github.com/ipfs/js-ipfs-api/pull/160

PR #160, opened 2015-12-06
Replace `require.context` with individual requires.

alexmingoia:
Webpack's `require.context` is transpiled only if ipfs-api is loaded or bundled
using webpack. This broke modules that require ipfs-api and use browserify or
other bundlers (such as webui)

Fixes ipfs/js-ipfs-api#159 by using individual requires. I opted for this because IMO maintaining a list of requires is less headache than dealing with the complexities of directory walking in different runtime contexts (browser, node, webpack, browserify, etc.)

### DIFF
--- a/src/load-commands.js
+++ b/src/load-commands.js
@@ -2,25 +2,34 @@
 
 const isNode = !global.window
 
-function requireDir () {
+function requireCommands () {
   if (isNode) return require('require-dir')('./api')
 
-  // Webpack specific require of a directory
-  const req = require.context('./api', false, /\.js$/)
-
-  const files = {}
-  req.keys().forEach(key => {
-    const name = key
-      .replace(/^\.\//, '')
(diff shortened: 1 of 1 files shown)


diasdavid:
Thank you @alexmingoia ! This is great, definitly :+1: for making browserify work again :)

Dignifiedquire:
@diasdavid :( not a big fan of this, but at least this code should be consistent between node and the browser then when we are using individual requires as there is no need for using `require-dir` in the first place now

diasdavid:
It is a bit more verbose, but still looks good, you feel differently? 

> but at least this code should be consistent between node and the browser then when we are using individual requires as there is no need for using require-dir in the first place now

We are also close of getting the extra benefit of enabling devs to just require the api call they need like require('src/api/add'), if we make send an module that is required instead of passed on load :)

Issue #161, opened 2015-12-08
When there is no port, we get `host:/path`


harlantwood:
eg instead of 

`https://ipfs.io/api...`

we get

`https://ipfs.io:/api...`

see also https://github.com/jacksenechal/js-ipfs-api/commit/c5085b16de2cd5294113062a0afe27ef2df09f53

PR #162, opened 2015-12-08
Exclude port from path if `config.port` not set.

alexmingoia:
Fixes ipfs/js-ipfs-api#161.

### DIFF
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -90,9 +90,11 @@ function requestAPI (config, path, args, qs, files, buffer, cb) {
   // this option is only used internally, not passed to daemon
   delete qs.followSymlinks
 
+  const port = config.port ? `:${config.port}` : ''
+
   const opts = {
     method: files ? 'POST' : 'GET',
-    uri: `${config.protocol}://${config.host}:${config.port}${config['api-path']}${path}?${Qs.stringify(qs, {arrayFormat: 'repeat'})}`,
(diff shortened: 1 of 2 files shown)


Dignifiedquire:
thanks. could you add a test for this please?

alexmingoia:
Of course. Tests added.

src/request-api.js
```diff
@@ -90,9 +90,11 @@ function requestAPI (config, path, args, qs, files, buffer, cb) {
   // this option is only used internally, not passed to daemon
   delete qs.followSymlinks
 
+  const port = config.port ? `:${config.port}` : ''
+
   const opts = {
     method: files ? 'POST' : 'GET',
-    uri: `${config.protocol}://${config.host}:${config.port}${config['api-path']}${path}?${Qs.stringify(qs, {arrayFormat: 'repeat'})}`,
+    uri: `${config.protocol}://${config.host}${port}${config['api-path']}${path}?${Qs.stringify(qs, {arrayFormat: 'repeat'})}`,
```
travisperson: Handling this with the `url` module is probably best. It cleans up that template a lot. Not a huge issue though.

```javascript
let uri = url.format({
  protocol: config.protocol,
  hostname: config.host,
  port: config.port,
  pathname: `${config['api-path']}/${path}`.replace(/\/+/g, '/'), // Double slashes don't matter, but it looks nicer.
  query: qs
})
```

It will handle the undefined port and not include it.

Dignifiedquire: :+1:

harlantwood:
Nice, thanks. cc @jacksenechal.

diasdavid:
Thank you for fixing this and adding the test :):)

PR #169, opened 2015-12-18
0.3.10

diasdavid:
I've started this, found a bug, but seems to be on the gulp karma task and not js-ipfs-api itself, tests print a ton more stuff then they used too.

Running the tests on master yield the same error, @Dignifiedquire any ideas? My first thought is that there was some patch upgrade in some of the babel+gulp+karma dependencies that broke things, which makes sense why all greenkeeper PRs now warn of breaking build

moved to [gist](https://gist.github.com/Dignifiedquire/4c144b1f2814944074c4)

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,10 @@ When using the api from script tag for things that require buffers (`ipfs.add`, 
 
 ## CORS
 
-If are using this module in a browser with something like browserify, then you will get an error saying that the origin is not allowed. This would be a CORS ("Cross Origin Resource Sharing") failure. The ipfs server rejects requests from unknown domains by default. You can whitelist the domain that you are calling from by exporting `API_ORIGIN` and restarting the daemon, like:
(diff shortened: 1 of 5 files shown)


Dignifiedquire:
@diasdavid looks like webpack logging needs be silenced

Dignifiedquire:
- [x] Tests passing locally
- [ ] Use ndjson to parse streamed output

Dignifiedquire:
@diasdavid as far as I'm concerned this is ready to merge

diasdavid:
awesome :+1: will look into this tomorrow properly to merge and publish :)

diasdavid:
Thank you @Dignifiedquire 

@fazo96 I've also submitted a small PR to https://github.com/jbenet/vinyl-fs/pull/3 so that in the future other people that use it don't hit the same problem

PR #172, opened 2015-12-19
feat: Update allowed node versions to >= 4.2.2

Dignifiedquire:


Mithgol:
What's the reason behind this new requirement?

diasdavid:
@Mithgol before it was ^4.0.0, which would mean that js-ipfs-api would only work in Node.js 4.x.x and not in Node.js versions above 4, so it is more of making it loose than an actual requirement.

Mithgol:
Okay, but what does make it not to work in Node 4.0 and require Node 4.2.2? That's what I was curious about.

PR #177, opened 2015-12-27
fix: Always return the underlying request object

Dignifiedquire:


src/api/add.js
```diff
@@ -8,15 +8,13 @@ module.exports = send => {
     }
 
     if (typeof files === 'string' && files.startsWith('http')) {
-      Wreck.request('GET', files, null, (err, res) => {
+      return Wreck.request('GET', files, null, (err, res) => {
         if (err) return cb(err)
 
         send('add', null, opts, res, cb)
```
diasdavid: Let's add a note here that ipfsAPI().add(<url>) doesn't return the right request.

Dignifiedquire: okay

PR #179, opened 2015-12-27
Switch to flat files stream

Dignifiedquire:
Depends on https://github.com/jbenet/node-vinyl-multipart-stream/pull/3


diasdavid:
woot! :D

PR #191, opened 2016-01-21
Update README Docs + remove unmaintained examples

diasdavid:


RichardLitt:
LGTM.

README.md
```diff
@@ -61,12 +61,49 @@ If are using this module in a browser with something like browserify, then you w
...
+## Usage
 
-### Level 1 Commands
-Level 1 commands are simple commands
+We classify the API calls by 'core', 'extensions', 'tooling', and 'network', following the same API spec organization available at [ipfs/specs](https://github.com/ipfs/specs/blob/ipfs/api/api/core/README.md)
 
-#### add
+The tests folder also contains great examples that can be used to understand how this client library interacts with the HTTP-API. You can find the tests here: https://github.com/ipfs/js-ipfs-api/tree/master/test/api
+
+### Core
```
Dignifiedquire: I don't understand what comes below, why are you adding `curl`in here instead of examples of the js-ipfs-api usage? Also what does the empty `Block` mean?

diasdavid: I just reorganized what was already here to follow https://github.com/ipfs/specs/blob/ipfs/api/api/core/README.md, the `curl` example was here before.

Dignifiedquire: yeah just realized that, I would suggest dropping it though

PR #195, opened 2016-01-22
test(ls): Add tests

Dignifiedquire:
Based on the work in #150.

Closes #150.


diasdavid:
So this is the same PR but with `expect` instead of `assert`, right? nice, thank you and thank you @fazo96 :):)

PR #199, opened 2016-02-01
Fixes #124 (Cannot declare a parameter named 'mask' as it shadows the name of a strict mode function)

### LINKED ISSUE #124: Cannot declare a parameter named 'mask' as it shadows the name of a strict mode function
I am using this package in combination with Meteor. The latest version will crash in Safari with the following error: Cannot declare a parameter named 'mask' as it shadows the name of a strict mode function. So far, I traced it back to this function on line 30299 in the ipfsapi.js: 
ip.mask = function mask(addr, mask) {

otonomos_dashboard_and_project_find_results_-__applications_mamp_htdocs_otometeor_-_atom_and_otonomos_dashboard

priecint:
I renamed the parameter so it doesn't collide
Related to #124

### DIFF
--- a/src/api/id.js
+++ b/src/api/id.js
@@ -1,11 +1,11 @@
 'use strict'
 
 module.exports = send => {
-  return function id (id, cb) {
-    if (typeof id === 'function') {
-      cb = id
-      id = null
+  return function id (idParam, cb) {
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thanks :octocat:

PR #200, opened 2016-02-05
feat: Add promise based api

### LINKED ISSUE #80: Promisses API
Most environments now support the more pleasant API of `Promise` ( https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm/Promise ), that is easier to handle and chain together. We should think about migrating to start using promises instead of callbacks where applicable.

List of browsers that supports it current: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm/Promise
(IE doesn't support it + some olders versions of other browsers, probably have to provide an existing shim as well...)

In node, promises have been natively supported since 0.12 so shouldn't be an issue there.

dignifiedquire:
All methods return a promise when not passed a callback

Closes #80

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ node_js:
   - '5'
   - stable
 
+addons:
+  firefox: 'latest'
+
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
--- a/README.md
+++ b/README.md
@@ -49,6 +49,20 @@ If you omit the host and port, the api will parse `window.host`, and use this in
 var ipfs = window.ipfsAPI()
 ```
 
+### Using Promises
+
+If you do not pass in a callback all api functions will return a `Promise`, for example
+
+```js
+ipfs.id()
+  .then(function (id) {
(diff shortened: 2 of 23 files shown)


diasdavid:
After the past discussions about having a promises implementations and the implications that would bring, this seems by far the most elegant solution (as it is almost minimal and only touches the code partially and gives the desired feature).

Before merging, I would like everyone that chimed in on the promises API discussion to give their review (if possible). The discussion happened here https://github.com/ipfs/js-ipfs-api/issues/80 and the people involved are: @VictorBjelkholm, @RichardLitt, @harlantwood, @travisperson and @bcomnes

harlantwood:
+100

fazo96:
I really hope this gets merged!

RichardLitt:
LGTM!

diasdavid:
It is time.. :)

dignifiedquire:
whoo

RichardLitt:
@jbenet You mentioned that you would like to discuss this. Here is the relevant PR for that discussion.

PR #221, opened 2016-02-27
Auto set json and bool flags on config.set

### LINKED ISSUE #219: Auto set json flag on config.set
Currently you can't call [`config.set`](https://github.com/ipfs/js-ipfs-api/blob/master/src/api/config.js#L8-L14) with an object as the `value`. You need to `JSON.stringify` it and pass `{ json: true }` on the `opts`.

I suggest that we check the type of the `value`, `JSON.stringify` it if it's an `object` set the `opts` automatically if none were provided.

Let me know what you think, I would be happy to PR this :)

xicombd:
Closes #219

### DIFF
--- a/src/api/config.js
+++ b/src/api/config.js
@@ -10,6 +10,15 @@ module.exports = (send) => {
         cb = opts
         opts = {}
       }
+
+      if (typeof (value) === 'object') {
+        value = JSON.stringify(value)
+        opts = { json: true }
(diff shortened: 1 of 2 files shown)


RichardLitt:
LGTM

diasdavid:
@xicombd restarted travis to see if the timeouts disappear (since they happened in parts not related at all with the code you touched)

PR #222, opened 2016-02-29
Avoid crash for responses without 'content-type' header

xicombd:
For example, when the response is a 200 OK without body

### DIFF
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -25,7 +25,7 @@ function onRes (buffer, cb) {
 
     const stream = !!res.headers['x-stream-output']
     const chunkedObjects = !!res.headers['x-chunked-output']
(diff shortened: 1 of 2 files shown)


dignifiedquire:
can we have a test for this?

xicombd:
AFAIK `go-ipfs` which is used on the tests always replies with `content-type`, I got this problem while working on `js-ipfs`, as such I added a test that uses it's own very basic `http` server. 
Let me know what you think :octocat:

diasdavid:
@xicombd LGTM, but since it is such a odd looking test, could you add 2 lines of code comments explaining why that test exists, otherwise it might get quickly deleted by someone in the future

xicombd:
@diasdavid added the the comments

PR #233, opened 2016-03-20
Use dignified.js

dignifiedquire:
> DO NOT MERGE

This  is only for discussions for now.

Implemented

- stricter lint to enforce strict mode everywhere
- prepare lint to be extracted into its own module, so no depdencies into nested eslintconfigs anymore
- gitignore lib and dist
- node build using babel in lib
- browser build using webpack
- reference npmcdn for use of script tags

cc @diasdavid @nginnever

Ref https://github.com/ipfs/community/pull/111


dignifiedquire:
Started extraction into: https://github.com/dignifiedquire/dignified.js

src/get-files-stream.js
```diff
@@ -82,7 +82,8 @@ function loadPaths (opts, file) {
         // files inside symlinks and others
         return
       })
-      .filter((file) => !!file) // filter out null files
+      // filter out null files
+      .filter((f) => Boolean(f))
```
diasdavid: Oh nice, didn't know about this wrapper. We can make it just `.filter(Boolean)`, right?

src/request-api.js
```diff
@@ -23,16 +23,16 @@ function onRes (buffer, cb) {
...
     const isJson = res.headers['content-type'] && res.headers['content-type'].indexOf('application/json') === 0
 
     if (res.statusCode >= 400 || !res.statusCode) {
       const error = new Error(`Server responded with ${res.statusCode}`)
 
-      return Wreck.read(res, {json: true}, (err, payload) => {
-        if (err) {
-          return cb(err)
+      return Wreck.read(res, {json: true}, (err1, payload) => {
+        if (err1) {
```
diasdavid: Do we needed to change it to `err1`?

test/api/files.spec.js
```diff
@@ -35,17 +39,17 @@ describe('.files', () => {
...
 
   it('files.write', (done) => {
-    apiClients['a'].files
+    apiClients.a.files
       .write('/test-folder/test-file-2.txt', new Buffer('hello world'), {create: true}, (err) => {
         expect(err).to.not.exist
 
-        apiClients['a'].files.read('/test-folder/test-file-2.txt', (err, stream) => {
-          expect(err).to.not.exist
+        apiClients.a.files.read('/test-folder/test-file-2.txt', (err1, stream) => {
```
diasdavid: why all the `err1` all the sudden?

diasdavid:
I feel that we will need to improve the docs on 'why the heck, so many folders auto generated', but this is looking good :):)

src/request-api.js
dignifiedquire: Because we are enforcing `no-shadowing` now from the linter, as it is super risky to allow it. Open for discussion though.

diasdavid: I like shadowing. I think it is a cool feature of a language that you can redefine vars under narrower  scopes

src/get-files-stream.js
dignifiedquire: That's how "type casting" in js is done, exists for all the types `Number`, `Boolean`, `String`.

Yes we can do.

dignifiedquire:
Ref https://github.com/dignifiedquire/dignified.js/issues/1

diasdavid:
@dignifiedquire could you resubmit this PR to master? Thank you

PR #265, opened 2016-05-07
Fix browser instance instructions in README

haadcode:
The name of the ipfs api global object has changed from `ipfsAPI` to `IpfsApi`. This PR will update the README to reflect the change.


dignifiedquire:
thanks

PR #266, opened 2016-05-09
test: extend block coverage

gavinmcdermott:
`block.put` now returns a rejected promise when no callback is provided and an error occurs; more tests.

### DIFF
--- a/src/api/block.js
+++ b/src/api/block.js
@@ -9,7 +9,7 @@ module.exports = (send) => {
     put (file, cb) {
       if (Array.isArray(file)) {
         let err = new Error('block.put() only accepts 1 file')
-        if (typeof cb !== 'function') {
+        if (typeof cb !== 'function' && typeof Promise !== 'undefined') {
           return new Promise((resolve, reject) => reject(err))
         }
         return cb(err)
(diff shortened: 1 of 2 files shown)


diasdavid:
Thank you @gavinmcdermott :) 

It LGTM, but I defer the final verdict to our in house promise expert, @dignifiedquire which added the promise support to tell if this matches the pattern he designed for js-ipfs-api :)

src/api/block.js
```diff
@@ -8,7 +8,11 @@ module.exports = (send) => {
     stat: argCommand(send, 'block/stat'),
     put (file, cb) {
       if (Array.isArray(file)) {
-        return cb(null, new Error('block.put() only accepts 1 file'))
+        let err = new Error('block.put() only accepts 1 file')
+        if (typeof cb !== 'function') {
```
dignifiedquire: please also check for `Promise` being defined: https://github.com/ipfs/js-ipfs-api/blob/master/src/api/log.js#L8

gavinmcdermott: Thanks @dignifiedquire, just getting myself up to speed with the libs to contribute more meaningfully in the near term!

dignifiedquire:
Thanks :octocat:

src/api/block.js
diasdavid: awesome, thank you @gavinmcdermott :D

PR #267, opened 2016-05-11
use object tests from interface-ipfs-core

diasdavid:
**WIP**

- [x] needs https://github.com/ipfs/interface-ipfs-core/pull/1


src/api/object.js
```diff
@@ -1,15 +1,71 @@
...
+          console.log('->', node)
+        } else {
+          node = new DAGNode(obj.Data, obj.Links)
+        }
+
+        if (node.toJSON().Hash !== result.Hash) {
+          return callback(new Error('Stored object was different from constructed object'))
+        }
+
+        callback(null, node)
```
diasdavid: In order to support the variety of ways `ipfs.object.put` can be called, I had to break the elegant promise injection that was there before. @dignifiedquire wanna share some wisdom on how we can add promises again, in the most elegant way possible?

diasdavid:
It took me more time to do object.put than I thought, spent too much time looking at the promises thing and I believe that we this new Interface document and tests, I caught a bunch of scenarios where the hash returned and the possible hash created client side would not match, not this last part is all good :)

src/api/object.js
```diff
@@ -1,15 +1,71 @@
...
+      let tmpObj = {
+        Data: null,
+        Links: []
+      }
+
+      if (Buffer.isBuffer(obj)) {
+        if (!options.enc) {
+          tmpObj = { Data: obj.toString(), Links: [] }
+        } else {
+          tmpObj = JSON.parse(obj.toString())
```
dignifiedquire: this should only happen if `enc === 'json'`

dignifiedquire: wait no, we actually don't want to parse at all in here, that's the job of ipfs itself isn't it?

src/api/object.js
dignifiedquire: it's probably best to just wrap the function using sth like this: https://github.com/manuel-di-iorio/promisify-es6

src/api/object.js
dignifiedquire: *confused*

diasdavid: This is for the use case (that people have been doing forever) that goes like:

```
const obj = {
        Data: new Buffer('Some data').toString(),
        Links: []
      }

      const buf = new Buffer(JSON.stringify(obj))
```

dignifiedquire: But why do you need to parse that, you can just pass that to go-ipfs with the encoding

diasdavid:
Tests complete. @dignifiedquire @haadcode mind CR'ing this one? This will be the API breaking change that will enable you to jump from js-ipfs-api to js-ipfs with a simple line, @haadcode :)

src/api/object.js
```diff
@@ -1,33 +1,259 @@
...
+        options = {}
+      }
+      if (!options) {
+        options = {}
+      }
+      multihash = cleanMultihash(multihash, options)
+
+      send('object/stat', multihash, null, null, callback)
+    }),
+    new: promisify((callback) => {
```
dignifiedquire: what about layouts?

src/api/object.js
```diff
@@ -1,33 +1,259 @@
...
+      switch (options.enc) {
+        case 'base58': multihash = multihash.toString()
+      }
+    } else {
+      multihash = bs58.encode(multihash).toString()
+    }
+  } else if (typeof multihash === 'string') {
+    if (options.enc) {
+      switch (options.enc) {
+        case 'base58': ; // It is good
```
dignifiedquire: semicolon?

src/api/object.js
```diff
@@ -1,33 +1,259 @@
...
+    if (options.enc) {
+      switch (options.enc) {
+        case 'base58': multihash = multihash.toString()
+      }
+    } else {
+      multihash = bs58.encode(multihash).toString()
+    }
+  } else if (typeof multihash === 'string') {
+    if (options.enc) {
+      switch (options.enc) {
```
dignifiedquire: why are you using a `switch` statement for a single case?

src/api/object.js
```diff
@@ -1,33 +1,259 @@
...
+      })
+    }
+  }
+  return api
+}
+
+function cleanMultihash (multihash, options) {
+  if (Buffer.isBuffer(multihash)) {
+    if (options.enc) {
+      switch (options.enc) {
```
dignifiedquire: why are you using a `switch` statement for a single case?

src/api/object.js
diasdavid: I tried all the other layouts through the CLI and only `unixfs-dir` works. It seems one of those deprecated features that no one ever used to even notice. I updated the spec accordingly.

dignifiedquire: Hmm, I made it fully work in js-ipfs :/

src/api/object.js
diasdavid: so that when someone add another encoding, they just add one clause instead of c&p a if clause

src/api/object.js
diasdavid: same as above

src/api/object.js
dignifiedquire: then please add a `default` case that throws with unknown encoding

src/api/object.js
diasdavid: Would you like to propose that we shim that feature in, although go-ipfs not supporting? All is possible with code 🎤 :D

diasdavid:
greeen :D

diasdavid:
ok, merging this one :D

Issue #152, opened 2015-11-25
Improve description of this module in the readme


dignifiedquire:
Ref https://github.com/ipfs/js-ipfs-api/issues/151#issuecomment-159673813 for context

diasdavid:
What else do we want to add to the readme? Could you itemsize please?

RichardLitt:
- [ ] Contribute section
- [ ] License section
- [ ] Contribute to IPFS! Banner

diasdavid:
Sounds good :) @RichardLitt, Master of All Docs, wanna PR that? :D

diasdavid:
done in #268 , thank you @RichardLitt :)

PR #268, opened 2016-05-11
Added Contribute, License, and Banner

RichardLitt:
Related to #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ $ npm install --save ipfs-api
 
 #### Running the daemon with the right port
 
-To interact with the API, you need to have a local daemon running (at the moment; we have plans to make it work independently). It needs to be open on the right port. `5001` is the default, and is used in the examples below, but it can be set to whatever you need.
+To interact with the API, you need to have a local daemon running. It needs to be open on the right port. `5001` is the default, and is used in the examples below, but it can be set to whatever you need.
 
 ```sh
 # Show the ipfs config API port to check it is correct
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -18,6 +18,24 @@ IPFS API wrapper library in JavaScript
 $ npm install --save ipfs-api
 ```
 
+#### Running the daemon with the right port
+
+To interact with the API, you need to have a local daemon running (at the moment; we have plans to make it work independently). It needs to be open on the right port: `5001`. This is the default.
```
diasdavid: > local daemon running (at the moment; we have plans to make it work independently

What do you have in mind?

> It needs to be open on the right port: `5001`. This is the default.

The default port is 5001, but it can open any really, if the user changes the default port, then it has to change the port in the example too :)

README.md
```diff
@@ -66,3 +84,34 @@ $ ipfs config --json API.HTTPHeaders.Access-Control-Allow-Origin "[\"http://exam
...
+
+### Callbacks and promises
+
+If you do not pass in a callback all API functions will return a `Promise`. For example:
+
+```js
+ipfs.id()
+  .then(function (id) {
+    console.log('my id is: ', id)
+  })
```
diasdavid: missing the .catch()

diasdavid:
Thank you @RichardLitt :) CR'ed

dignifiedquire:
@RichardLitt why did you move the promise vs callback into the readme? It's part of the API description

RichardLitt:
Yeah, I understand that. I did it that section should apply to the entire repository, and I think it needs a higher visibility. The API.md file should be in the Readme, I think; it is more cohesive as a file on it's own without extra information like callbacks or promises. Does that make sense?

PR #271, opened 2016-05-13
files.get

nginnever:
@xicombd 

I added the compression args to what you had worked on earlier. Not sure if this is the most elegant way to handle these things so let me know what you think! :D

### DIFF
--- a/src/api/get.js
+++ b/src/api/get.js
@@ -1,0 +1,24 @@
+'use strict'
+
+module.exports = (send) => {
+  return function get (path, archive, compress, compressionLevel, cb) {
+    if (archive === true && typeof compress === 'function') {
+      cb = compress
+      compressionLevel = null
+      compress = null
+    }
+    if (archive === true && typeof compress === 'number') {
+      archive = null
+      cb = compressionLevel
+      compressionLevel = compress
+      compress = true
+    }
+    if (typeof archive === 'function') {
+      cb = archive
+      archive = null
+      compressionLevel = null
+      compress = null
+    }
+    return send('get', path, [archive, compress, compressionLevel], null, cb)
+  }
+}
--- a/src/api/get.js
+++ b/src/api/get.js
@@ -1,24 +1,11 @@
 'use strict'
 
 module.exports = (send) => {
-  return function get (path, archive, compress, compressionLevel, cb) {
-    if (archive === true && typeof compress === 'function') {
-      cb = compress
-      compressionLevel = null
-      compress = null
+  return function get (path, opts, cb) {
+    if (typeof opts === 'function' && !cb) {
+      cb = opts
+      opts = {}
     }
-    if (archive === true && typeof compress === 'number') {
-      archive = null
-      cb = compressionLevel
-      compressionLevel = compress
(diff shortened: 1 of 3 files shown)


src/api/get.js
```diff
@@ -0,0 +1,24 @@
+'use strict'
+
+module.exports = (send) => {
+  return function get (path, archive, compress, compressionLevel, cb) {
+    if (archive === true && typeof compress === 'function') {
```
RichardLitt: Maybe use a switch statement instead?

xicombd: Or since `archive`, `compress`, `compressionLevel` are all optional, why not pass them on the `opts` object?

I think that you can just return the `argCommand` and it will [take care of that](https://github.com/ipfs/js-ipfs-api/blob/master/src/cmd-helpers.js#L14) :)

src/api/get.js
```diff
@@ -0,0 +1,24 @@
...
+      compressionLevel = compress
+      compress = true
+    }
+    if (typeof archive === 'function') {
+      cb = archive
+      archive = null
+      compressionLevel = null
+      compress = null
+    }
+    return send('get', path, [archive, compress, compressionLevel], null, cb)
```
xicombd: Also, aren't the opts a object and not an array?

diasdavid:
awesome @nginnever, great start! :D `get` should offer the same API as js-ipfs, as you know, it returns an event for each file that is transmitted.

Also, you should include tests for 'get directories'

src/api/get.js
dignifiedquire: Agreed, please make `archive`, `compress` and `compressionLevel` into an options object.

src/api/get.js
dignifiedquire: yep

test/api/get.spec.js
```diff
@@ -0,0 +1,119 @@
...
+
+        let buf = ''
+        res
+          .on('error', (err) => {
+            expect(err).to.not.exist
+          })
+          .on('data', (data) => {
+            buf += data
+          })
+          .on('end', () => {
```
dignifiedquire: Let's use [`bl`](https://www.npmjs.com/package/bl) for these tests

diasdavid:
@nginnever what is the status of this PR?

nginnever:
@diasdavid still needs a lot of work.  Now that the add and cat command are starting to be finalized in all of the modules this can be restarted. 

Still needs unixfs-engine compression, js-ipfs update, and interface-core spec. There are some tests began here that could start to be moved to interface-core tests.

diasdavid:
> Still needs unixfs-engine compression 

The compression doesn't happen at the unixs-engine level right? It is just at the wire level that everything gets tar'ed

noffle:
Hey @nginnever! I gave this the last little bit of spit 'n shine it needed: https://github.com/ipfs/js-ipfs-api/pull/296

If that looks good, I think we can safely close this PR.

diasdavid:
Followed in https://github.com/ipfs/js-ipfs-api/pull/296

PR #272, opened 2016-05-13
fix: add support for protobuf in object.put

dignifiedquire:
Ref https://github.com/ipfs/interface-ipfs-core/pull/7

### DIFF
--- a/src/api/object.js
+++ b/src/api/object.js
@@ -68,7 +68,7 @@ module.exports = (send) => {
       }
       const enc = options.enc || 'json'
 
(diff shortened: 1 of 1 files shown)


src/api/object.js
```diff
@@ -68,21 +68,24 @@ module.exports = (send) => {
       }
       const enc = options.enc || 'json'
 
-      send('object/put', enc, null, buf, (err, result) => {
+      send('object/put', null, {inputenc: enc}, buf, (err, result) => {
```
diasdavid: sweet! Thank you (we've been passing enc as just enc for ages)

PR #273, opened 2016-05-14
Fixing small issues with the new object api

dignifiedquire:


diasdavid:
oh, the result is not always a stream? Good catch :)

PR #280, opened 2016-05-22
WIP cat interface core

nginnever:
This begins the js-ipfs-api cat command using the interace-core tests shared with the js-ipfs core tests.

This needs the completition of the cat interface-core <a href="https://github.com/ipfs/interface-ipfs-core/pull/21">PR</a>

### DIFF
--- a/src/api/cat.js
+++ b/src/api/cat.js
@@ -1,7 +1,32 @@
 'use strict'
 
-const argCommand = require('../cmd-helpers').argCommand
+const bs58 = require('bs58')
+const isIPFS = require('is-ipfs')
+const promisify = require('promisify-es6')
 
 module.exports = (send) => {
-  return argCommand(send, 'cat')
+  const cat = promisify((multihash, callback) => {
+    try {
+      multihash = cleanMultihash(multihash)
+    } catch (err) {
+      return callback(err)
+    }
+    send('cat', multihash, null, null, function (err, result) {
+      if (err) {
+        return callback(err)
+      }
+      return callback(null, result)
+    })
+  })
+  return cat
+}
+
+function cleanMultihash (multihash) {
+  if (!isIPFS.multihash(multihash)) {
+    throw new Error('not valid multihash')
+  }
+  if (Buffer.isBuffer(multihash)) {
+    return bs58.encode(multihash)
+  }
+  return multihash
 }
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "flatmap": "0.0.3",
     "glob": "^7.0.3",
     "ipfs-merkle-dag": "^0.6.0",
+    "is-ipfs": "^0.2.0",
     "multiaddr": "^2.0.0",
     "multipart-stream": "^2.0.1",
     "ndjson": "^1.4.3",
--- a/test/api/file.spec.js
+++ b/test/api/file.spec.js
@@ -1,0 +1,17 @@
+/* eslint-env mocha */
+/* globals apiClients */
+
+'use strict'
+
+const test = require('interface-ipfs-core')
+
+const common = {
+  setup: function (cb) {
+    cb(null, apiClients.a)
+  },
+  teardown: function (cb) {
+    cb()
+  }
+}
+
+test.files(common)


test/api/cat.spec.js
```diff
@@ -1,79 +1,17 @@
...
-            })
-            .on('end', () => {
-              expect(buf).to.be.equal(testfile.toString())
-              done()
-            })
-        })
-    })
-  })
-})
+test.files(common)
```
noffle: *Note to self: once I get in the `add` test PR, rename this file to `files-spec.js`, so we don't run the interface files tests twice.*

noffle:
:horse:LGTM

test/api/cat.spec.js
nginnever: thought of this as well, but calling it ```file.spec.js``` since ```files.spec.js``` is taken by the mfs tests

noffle: Wasn't there an issue somewhere about explicitly renaming that to `mfs` across js-ipfs? Having trouble finding that..

src/api/files.js
```diff
@@ -4,6 +4,7 @@ const argCommand = require('../cmd-helpers').argCommand
 
 module.exports = (send) => {
   return {
+    cat: argCommand(send, 'cat'),
```
nginnever: this is a mistake, should not be here

src/api/cat.js
```diff
@@ -1,7 +1,32 @@
...
 
-const argCommand = require('../cmd-helpers').argCommand
+const bs58 = require('bs58')
+const isIPFS = require('is-ipfs')
+const promisify = require('promisify-es6')
 
 module.exports = (send) => {
-  return argCommand(send, 'cat')
+	const cat = promisify((multihash, callback) => {
+		try {
```
dignifiedquire: I think we use spaces in JavaScript for indentation ;)

src/api/cat.js
```diff
@@ -1,7 +1,32 @@
...
+      if (err) {
+        return callback(err)
+      }
+      return callback(null, result)
+    })
+  })
+  return cat
+}
+
+function cleanMultihash (multihash) {
```
noffle: I think this is defined in `src/api/object.js` too. Could you toss it into a module in `src/`?

src/api/cat.js
```diff
@@ -1,7 +1,32 @@
...
+  if (!isIPFS.multihash(multihash)) {
+    throw new Error('not valid multihash')
+  }
+  if (Buffer.isBuffer(multihash)) {
```
noffle: Isn't input allowed to be a Buffer containing a base58-encoded multihash? If so, this would double b58-encode it.

nginnever: input is allowed to be a buffer for js-ipfs core, but the go-ipfs and js-ipfs-api could only take bs58 encoded strings of multihashes so i use this check to make sure that if a buffer multihash is supplied, then it is converted to a format everyone can use, bs58 encoded string.

It shouldn't double multihash

noffle: The linked to doc says "the raw Buffer of the multihash (or of and encoded version)". Did I misunderstand what "of the encoded version" means? It means "base58 encoded", right?

Underyling assumption: this `cat` method is supposed to satisfy interface-ipfs-core's `cat` method. Is this correct?

nginnever: >The linked to doc says "the raw Buffer of the multihash (or of and encoded version)". Did I misunderstand what "of the encoded version" means? It means "base58 encoded", right?

I don't think the raw Buffer is base58 encoded, this is borrowed from the object.get definition of a multihash input. The string mentioned after it is the base58 encoded multihash

>Underyling assumption: this cat method is supposed to satisfy interface-ipfs-core's cat method. Is this correct?

Right, this upgrade to cat will now take a buffer input so that it is interop with interface-core

noffle: Maybe I've misunderstood? In "the raw Buffer of the multihash (or of encoded version)" -- what encoding does "the raw Buffer" have vs "encoded version"?

nginnever: I've updated the <a href="https://github.com/ipfs/interface-ipfs-core/pull/21/files#diff-04c6e90faac2675aa89e2176d2eec7d8R70">readme</a> in interface-core, hopefully that simplifies the statement. A buffer is just a node.js default utf-8 buffer of a multihash, and a ```Qm...``` string is base58 encoded.

noffle: YES SO AWESOME

nginnever:
closing in favor of https://github.com/ipfs/js-ipfs-api/pull/292 where others can edit.

PR #281, opened 2016-05-24
Make ipfs.files.add return DAGNodes.

noffle:
- [x] do an object lookup to transform objects to `DAGNode`s
- [x] break out local file adding and url adding to separate methods
- [x] use + pass interface-ipfs-core tests for files.add
- [ ] `add` tests to get merged into `interface-ipfs-core` and published

NOTE: this breaks backwards compatibility for `add` and will require a major semver bump.

PR #288, opened 2016-06-05
Files add interface

diasdavid:
Builds on https://github.com/ipfs/js-ipfs-api/pull/281


src/add-to-dagnode-transform.js
```diff
@@ -0,0 +1,25 @@
...
+
+const async = require('async')
+const getDagNode = require('./get-dagnode')
+
+// transform { Hash: '...' } objects into { path: 'string', node: DAGNode }
+module.exports = function (err, res, send, done) {
+  if (err) {
+    return done(err)
+  }
+  async.map(res, function map (entry, next) {
```
dignifiedquire: please try to avoid adding async

diasdavid: roger. Due to time constraints, can we make that a issue to hit later? It would be good to explain in one paragraph (or even one line -> "async is a huge dependency, makes our browser lib fat") so that @noffle and other contribs try to avoid it in the future?

dignifiedquire: sure there is already an issue on js-ipfs for exactly that

src/add-to-dagnode-transform.js
```diff
@@ -0,0 +1,25 @@
...
+        return next(err)
+      }
+      var obj = {
+        path: entry.Name,
+        node: node
+      }
+      next(null, obj)
+    })
+  }, function (err, res) {
+    done(err, res)
```
dignifiedquire: this could have been just passed done directly

PR #291, opened 2016-06-05
feature - createAddStream

diasdavid:
- [x] fix adding empty dirs
- [x] createAddStream tests on interface-ipfs-core
- [x] createAddStream implemented

### DIFF
--- a/src/api/add-stream.js
+++ b/src/api/add-stream.js
@@ -1,0 +1,36 @@
+'use strict'
+
+const addCmd = require('./add.js')
+const Duplex = require('stream').Duplex
+const promisify = require('promisify-es6')
+
+module.exports = (send) => {
+  const add = addCmd(send)
+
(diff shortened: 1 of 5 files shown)


diasdavid:
- [ ] need to add the promisify

src/api/add-stream.js
```diff
@@ -0,0 +1,34 @@
...
+
+        res.forEach((tuple) => {
+          ds.push(tuple)
+        })
+
+        ds.push(null)
+      })
+    }
+    callback(null, ds)
+  })
```
haadcode: Are errors handled somewhere?

haadcode:
One question re. error handling, otherwise LGTM

src/api/add-stream.js
diasdavid: good catch, tahnk you. Emitted the error upstream :)

PR #292, opened 2016-06-05
Follow interface-ipfs-core for files.cat

nginnever:


noffle:
What needs to happen for this PR?

nginnever:
@noffle this needs interface-core merge https://github.com/ipfs/interface-ipfs-core/pull/21

And it should be good :D

PR #297, opened 2016-06-16
Include directory in pattern

VictorBjelkholm:
Prevents issue where the actual directory is not included when adding a directory with addFiles.

`addFiles('/usr/src/app')` would for example match everything inside `/usr/src/app` but not `/usr/src/app` itself.

### DIFF
--- a/src/get-files-stream.js
+++ b/src/get-files-stream.js
@@ -44,7 +44,7 @@ function loadPaths (opts, file) {
   }
 
   if (stats.isDirectory() && opts.recursive) {
-    const mg = new glob.sync.GlobSync(`${file}/**/*`, {
+    const mg = new glob.sync.GlobSync(`${file}/**`, {
       follow: followSymlinks
     })
 


VictorBjelkholm:
Hm, now when I try it out, it seems like it would get the same hash as another subdir. Not sure what's going on. Should probably wait with merging this, introduce some tests for `addFiles` and see what's going on.

VictorBjelkholm:
Created a issue regarding the hash differences here: https://github.com/ipfs/js-ipfs-api/issues/300

VictorBjelkholm:
I was using the wrong IPFS version, which lead to the bug where directory is not included. So this can be considered fixed.

diasdavid:
@VictorBjelkholm can you add some tests, so that the bug doesn't get introduced again? Thank you!

diasdavid:
@VictorBjelkholm just checkin in, did you see my last comment?

Issue #217, opened 2016-02-23
implement `bootstrap` calls to the API


diasdavid:
HTTP-API-SPEC for bootstrap - https://github.com/ipfs/http-api-spec/blob/master/apiary.apib#L1162

PR #298, opened 2016-06-16
Enhancement: Implement Bootstrap API

gavinmcdermott:
Implement the bootstrap API in reference to issue #217 raised by @diasdavid

### DIFF
--- a/test/api/bootstrap.spec.js
+++ b/test/api/bootstrap.spec.js
@@ -1,0 +1,180 @@
+/* eslint-env mocha */
+/* globals apiClients */
+'use strict'
+
+const expect = require('chai').expect
+
+const invalidArg = 'this/Is/So/Invalid/'
+const validIp4 = '/ip4/104.236.176.52/tcp/4001/ipfs/QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z'
+
+describe('.bootstrap', () => {
+  let peers
+
+  describe('.add', () => {
+    it('returns an error when called without args or options', (done) => {
+      return apiClients.a.bootstrap.add(null, (err) => {
+        expect(err).to.be.an.instanceof(Error)
+        done()
+      })
+    })
+
+    it('returns an error when called with an invalid arg', (done) => {
+      return apiClients.a.bootstrap.add(invalidArg, (err) => {
+        expect(err).to.be.an.instanceof(Error)
+        done()
+      })
+    })
+
+    it('returns a list of containing the bootstrap peer when called with a valid arg (ip4)', (done) => {
+      return apiClients.a.bootstrap.add(validIp4, (err, res) => {
+        expect(err).to.not.exist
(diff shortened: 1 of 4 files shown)


test/api/bootstrap.spec.js
```diff
@@ -0,0 +1,189 @@
...
+        expect(peers).to.exist
+        expect(peers.length).to.above(1)
+        done()
+      })
+    })
+  })
+
+  describe('.list', () => {
+    it('returns a list of peers', (done) => {
+      return apiClients.a.bootstrap.list(null, (err, res) => {
```
diasdavid: Why do we need the `null` arg?

test/api/bootstrap.spec.js
```diff
@@ -0,0 +1,189 @@
...
+  describe('.rm', () => {
+    it('returns an error when called with an invalid arg', (done) => {
+      return apiClients.a.bootstrap.rm(invalidArg, (err) => {
+        expect(err).to.be.an.instanceof(Error)
+        done()
+      })
+    })
+
+    it('returns empty list when called without an arg or options', (done) => {
+      return apiClients.a.bootstrap.rm(null, (err, res) => {
```
diasdavid: Does this mean 'delete all'?

diasdavid:
Awesome work @gavinmcdermott ! :D left a couple of questions

test/api/bootstrap.spec.js
gavinmcdermott: Great question.....we don't

test/api/bootstrap.spec.js
gavinmcdermott: Nope. According to the spec there is an explicit 'delete all' call—you just have to pass an `all: true` option in order to remove all bootstrap peers. The returned array contains the list of peers that were removed in the call (I'll also make the `it` string a better description).

test/api/bootstrap.spec.js
diasdavid: Can we take it off then?

gavinmcdermott: Of course! Was away from my machine yesterday - should have mentioned in the note that I'd update the pr within the day. It's up now :)

diasdavid:
LGTM :) Thank you @gavinmcdermott 

Could you squash the commits for the merge?

test/api/bootstrap.spec.js
```diff
@@ -0,0 +1,189 @@
...
+    })
+  })
+
+  describe('.promise', () => {
+    describe('.add', () => {
+      it('returns an error when called without args or options', (done) => {
+        return apiClients.a.bootstrap.add(null)
+          .catch((err) => {
+            expect(err).to.be.an.instanceof(Error)
+            done()
```
dignifiedquire: You can drop the `done` in all the promise tests, mocha detects that you return a promise (also need to remove it from the function signature.

PR #307, opened 2016-06-30
interface-ipfs-core config API spec compliant

diasdavid:
> **WIP**

### DIFF
--- a/src/api/config.js
+++ b/src/api/config.js
@@ -1,31 +1,64 @@
 'use strict'
 
-const argCommand = require('../cmd-helpers').argCommand
+const streamifier = require('streamifier')
 
 module.exports = (send) => {
   return {
-    get: argCommand(send, 'config'),
-    set (key, value, opts, cb) {
-      if (typeof (opts) === 'function') {
-        cb = opts
-        opts = {}
+    get (key, callback) {
+      if (typeof key === 'function') {
+        callback = key
+        key = undefined
       }
 
-      if (typeof (value) === 'object') {
+      if (!key) {
+        return send('config/show', null, null, null, true, callback)
+      }
+
+      return send('config', key, null, null, (err, result) => {
+        if (err) {
+          return callback(err)
+        }
+        callback(null, result.Value)
+      })
+    },
+    set (key, value, opts, callback) {
+      if (typeof opts === 'function') {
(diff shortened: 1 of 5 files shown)


src/api/config.js
```diff
@@ -1,31 +1,53 @@
...
-      if (typeof (opts) === 'function') {
-        cb = opts
+    get (key, callback) {
+      if (typeof key === 'function') {
+        callback = key
+        key = undefined
+      }
+
+      if (!key) {
+        return send('config/show', null, null, null, true, callback)
```
VictorBjelkholm: Probably be valuable for the `send` method to accept a config argument rather than having a parameter list filled with `null`s

diasdavid: 💯 agree. In fact, what I'm very much thinking is after merging the config-api PR into ipfs-api/ng branch, just make a PR to update how request-api.js works, break the tests (so that we don't have to run 300 tests while developing a single feature) and clean up whatever else I find.

src/api/config.js
```diff
@@ -1,31 +1,64 @@
...
+      // Its a path
+      if (typeof config === 'string') {
+        return send('config/replace', null, null, config, callback)
+      }
+
+      // Its a config obj
+      if (typeof config === 'object') {
+        config = streamifier.createReadStream(new Buffer(JSON.stringify(config)))
+        return send('config/replace', null, null, config, callback)
+      }
```
diasdavid: Just missing to make the config.replace tests to work. It seems that this was never working (see https://github.com/ipfs/js-ipfs-api/blob/master/test/api/config.spec.js#L123-L133), unfortunately.

The current problem is that the config never gets replaced and after we send the replace, the content type of the following config.get requests becomes text/plain instead of application/json

diasdavid: Learned that this is an error in actual go-ipfs https://github.com/ipfs/go-ipfs/issues/2927

src/api/config.js
VictorBjelkholm: 👍

diasdavid: It is happening :) https://github.com/ipfs/js-ipfs-api/pull/313/commits/43a1dfab4322dd06af7fd158dae29e7da71d4d86

Issue #301, opened 2016-06-21
ipfs.add is not a function


chpio:
im on *v6.0.1*

harlantwood:
I'm getting the same issue.  worked fine on 4.1.0, but missing in 6.0.1.

Also seems to be missing from the API docs: https://github.com/ipfs/js-ipfs-api/blob/master/API.md .  In the API docs there is an `add` within `Extensions`, though its `tests` link points to a page that does not exist: https://github.com/ipfs/js-ipfs-api/blob/master/test/api/add.spec.js

harlantwood:
Narrow version range: add works in 4.1.1, missing in 5.0.1

harlantwood:
Maybe related to #279

nginnever:
Hey @harlantwood!

We did change the way ```add``` is interfaced as you see in #279 

noffle added an alias <a href="https://github.com/ipfs/js-ipfs-api/blob/master/src/load-commands.js#L27">here</a> but it's not working. Going to look into it

nginnever:
Opened a <a href="https://github.com/ipfs/js-ipfs-api/pull/309">PR</a> that will allow people to call add with or without files until the namespaces are ironed out.

harlantwood:
Thanks @nginnever!

diasdavid:
Fixed on https://github.com/ipfs/js-ipfs-api/pull/309. Reopen as necessary.

Thank you @nginnever :)

PR #309, opened 2016-07-01
alias for add and createAddStream.

nginnever:
Currently calling ```ipfs.add``` and ```ipfs.createAddStream``` will fail since the alias isn't picked up until after the ```ipfs.files``` command is called. 

Not sure if this is the best approach but this will allow people to use add with or without files until the namespace collision is solved.

This was referenced in issue #301 

@noffle could you check this and see if there could be any issues aliasing this way instead of how you had it.

### DIFF
--- a/src/load-commands.js
+++ b/src/load-commands.js
@@ -2,6 +2,8 @@
 
 function requireCommands () {
   const cmds = {
+    add: require('./api/add'), // add alias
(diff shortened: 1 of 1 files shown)


diasdavid:
awesome @nginnever, several people reported this, thank you for fixing it! :)

PR #322, opened 2016-07-19
Add tests for add

taravancil:


dignifiedquire:
Thank you but we have more comprehensive tests for `add` in https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js. They are loaded here: https://github.com/ipfs/js-ipfs-api/blob/master/test/api/files.spec.js#L24

RichardLitt:
@dignifiedquire Can we make it clearer that that is the case? Duplicated effort should be avoidable.

dignifiedquire:
@RichardLitt we are in this weird transition period, eventually 95% of the tests will be moved to interface-ipfs-core and then it will be much clearer. Not sure how to do it in the meantime, PRs welcome though :)

RichardLitt:
Well, right now we have this in the README:

> Add tests. There can never be enough tests.

Maybe remove that?

diasdavid:
Thank you @taravancil, it is really our fault for not being more explicit on how we are defining the interface and testing it. You see, since we are shooting to have the same interface in both js-ipfs and js-ipfs-api, we created a module just for its definition and batch of tests.

I've added more information to the readme on https://github.com/ipfs/js-ipfs-api/pull/332, please do continue to contribute to js-ipfs and js-ipfs-api :) We are still far from 100% code coverage, would love to get there :) 

Going to close this PR for now. Check all the other tests at: https://github.com/ipfs/interface-ipfs-core/tree/master/src

Issue #323, opened 2016-07-27
Files API (add, cat, get) tests break go-ipfs 0.4.3


diasdavid:
js-ipfs-api files API (add, cat, get) tests break with the new go-ipfs 0.4.3 release. The error is always the same.

``` sh
1) .files .add stream:
  Uncaught TypeError: stream.pipe is not a function
    at done (src/get-dagnode.js:30:14)
    at node_modules/async/dist/async.js:3679:13
```

haadcode:
https://github.com/haadcode/orbit/issues/40 is blocked due to this bug

dignifiedquire:
Actually only `.add` breaks for me, and that is due to `object/get` breaking.

dignifiedquire:
Okay found the issue why the `object` requests are breaking. 

Running `ipfs add data.txt` returns

```js
// go-ipfs 0.4.1
[ { Name: 'data.txt',
    Hash: 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS' } ]

// go-ipfs 0.4.3-rc1
[ { Name: 'data.txt', Bytes: 9 },
  { Name: 'data.txt',
    Hash: 'QmVv4Wz46JaZJeH5PMV4LGbRiiMKEmszPYY3g6fjGnVXBS' } ]
```

jbenet:
how is this not caught in a go-ipfs test!? o/

dignifiedquire:
> how is this not caught in a go-ipfs test!? o/

Not sure, are there tests for all commands through the http api? (this only happens on the http api not on the cli as far as I can tell)

Kubuxu:
> how is this not caught in a go-ipfs test!? o/

It is very simple, the way that current ipfs-http-api was developed was for internal daemon - CLI communication. It is directly bound to CLI commands so every change there might introduce changes in ipfs-http-api.

In go-ipfs there are tests for CLI, and few tests for edge cases of the HTTP API but coverage of HTTP API is minimal. In my opinion go-ipfs HTTP daemon - CLI connector should have never be named the ipfs-http-api.

The way it was designed was directly for our CLI implementation, not as HTTP API, resulting in very hard to use API. It is directly bound to go-ipfs CLI which means that we have to be really careful even when fixing smallest bugs in CLI, not only from CLI userspace perspective but also HTTP perspective.

[...]

PR #305, opened 2016-06-30
next generation

diasdavid:
Since this module original inception, a lot has changed in the IPFS JavaScript ecosystem, now we have things like a full [implementation of the Protocol in JavaScript](https://github.com/ipfs/js-ipfs), new and more powerful features like the mfs (Mutable File System) API and a lot of the building blocks to make the IPFS experience feel native to JavaScript (having [multispecs](https://github.com/ipfs/notes/issues/135) implementations).

That being said and from all of the new and good things, a lot of technical debt was added, more and more the interfaces feel like something clunky/patched together, making it hard for the community to follow and for developers to decide what to use.

We started the endeavour of creating [interface-ipfs-core](), an endeavour to standardise what is the core interface an IPFS implementation should offer, this will help developers switch between embebed nodes (js-ipfs) and remote nodes (e.g go-ipfs) without having to change the code in their applications. In addition to this and since this module offered a lot of utilities that came from a direct mapping to the go-ipfs cli, we are keeping them under the `.util` scope, so that you can continue to use the feature you know and love.

Our goal is with this, get a really solid UX to interact with IPFS, we appreciate everyones feedback and contributions during the process. I'll keep this PR open while we add everything, creating other PR to submit to this one (Block API, mfs API, etc). Below you can see a list of tasks:

#### UPDATE ➡️ https://github.com/ipfs/js-ipfs-api/pull/305#issuecomment-240108393 ⬅️

- [x] generic API https://github.com/ipfs/interface-ipfs-core/pull/33
  - [x] .id
  - [x] .version
- [x] config API
  - [x] get
  - [x] set
  - [ ] replace - requires https://github.com/ipfs/go-ipfs/issues/2927
- [x] block API
  - [x] block.get
  - [x] block.put
  - [ ] block.rm
    - [ ] Implement in go-ipfs https://github.com/ipfs/go-ipfs/issues/2914
  - [x] block.stat
- [x] object API
  - [x] object.new
  - [x] object.put
  - [x] object.get
  - [x] object.data
  - [x] object.links
  - [x] object.stat
  - [x] object.patch.addLink
  - [x] object.patch.rmLink
  - [x] object.patch.appendData
  - [x] object.patch.setData
- [x] files API
  - [x] files.add
  - [x] files.add upgrade https://github.com/ipfs/interface-ipfs-core/issues/29
  - [x] files.createAddStream
  - [x] files.cat
  - [x] files.get https://github.com/ipfs/interface-ipfs-core/pull/28
[...]

PR #335, opened 2016-08-04
docs(examples): move to webpack

dignifiedquire:


nycoliver:
Compiles fine, but ipfs.add throws a TypeError: stream.pipe is not a function
Also in Safari a SyntaxError: Unexpected keyword 'const'. Const declarations are not supported in strict mode.

nycoliver:
For some reason the response from add has two objects, the first of which has no name and no hash: 
https://github.com/ipfs/js-ipfs-api/blob/master/src/add-to-dagnode-transform.js#L23

I think that's causing the `TypeError: stream.pipe is not a function`

Not sure where the `ipfs add error Error: Payload stream closed prematurely` is coming from though.

nycoliver:
Ok, seems the stream.pipe error was due to https://github.com/ipfs/js-ipfs-api/issues/323, not seeing it anymore running the latest go-ipfs master. Not sure where the payload stream error is coming from.

diasdavid:
Seems like it is a bug in go-ipfs. The file gets added through the example, but `ipfs object data` fails, even through go-ipfs cli:

```

» ipfs object data QmSpDiGm6cwbEo1WZyo1YmKH51SboG5brSLqDop4EfAh3o
# returns nothing
» ipfs cat QmSpDiGm6cwbEo1WZyo1YmKH51SboG5brSLqDop4EfAh3o
   sdfadsfas%                                                
» ipfs object get QmSpDiGm6cwbEo1WZyo1YmKH51SboG5brSLqDop4EfAh3o
{"Links":[],"Data":"\u0008\u0002\u0012\r    sdfadsfas\u0018\r"}
```

Kubuxu:
Seems to work here on the master:
```
 ~  ipfs add
ipfs: Reading from /dev/stdin; send Ctrl-d to stop.

Hello world! It is great day, isn't it?
ipfs: Reading from /dev/stdin; send Ctrl-d to stop.

added QmRYcKkokPSnBzwPPVcsRxhj6gGwdhGCUV2mfRRbef4X2V QmRYcKkokPSnBzwPPVcsRxhj6gGwdhGCUV2mfRRbef4X2V
 ~  ipfs cat QmRYcKkokPSnBzwPPVcsRxhj6gGwdhGCUV2mfRRbef4X2V
Hello world! It is great day, isn't it?
 ~  ipfs object data QmRYcKkokPSnBzwPPVcsRxhj6gGwdhGCUV2mfRRbef4X2V
(Hello world! It is great day, isn't it?
(%                                                                                                 ~  ipfs object get QmRYcKkokPSnBzwPPVcsRxhj6gGwdhGCUV2mfRRbef4X2V
{"Links":[],"Data":"\u0008\u0002\u0012(Hello world! It is great day, isn't it?\n\u0018("}
```

diasdavid:
@Kubuxu it is just the files that are added through js-ipfs-api, on that specific example, that make `ipfs object data` fail. 

Note that our tests have coverage for this and they pass, it is just this example. Which makes me think, could the example be just broken? As in, could the webpack compile step be messing with it?

nycoliver:
Yes, it appears to be a problem with the webpack compile step... the example works by importing dist/index.js and https://wzrd.in/standalone/buffer in a script tag. However I would love to be able to just `import IpfsApi from 'ipfs-api'` in my webpack code, especially without running a brfs transform.

diasdavid:
@nycoliver thank you for checking meanwhile! We got slammed yesterday with the IPFS hangouts (you are invited to join btw! https://github.com/ipfs/pm).

I do like the idea of having a pristine js-ipfs-api, just like js-ipfs, which is not poluted by assumptions of 'it can access fs', however, we are not able to pursue that change right now as many people depend on it. We might phase it out to an external module, once https://github.com/ipfs/js-ipfs-api/pull/305 is complete and stable :)

nycoliver:
305 looks great, and thanks for the invitation!

I'll use ipfs-api in a script tag for now but such a hacky solution is really uncomfortable. Hoping this issue gets solved and I can bundle with webpack soon.

nginnever:
Also ran into this webpacking a new project using js-ipfs-api. I was able to reproduce from calling the ``id`` command however...

``ipfs.id()`` both in promise and callback.

 returns

``Error: Payload stream closed prematurely``

dignifiedquire:
Will wait until #390 is merged, as that makes building and things a lot simpler

dignifiedquire:
Fixed in #414

Issue #339, opened 2016-08-10
go-ipfs sometimes closes the stream preemptively


daviddias:
From: https://github.com/ipfs/js-ipfs-api/pull/337#issuecomment-238541189

---

Ok, I thought I had fixed the adding dirs on the browser, but then realised that there is a bug that only happens sometimes (like 1 out of 20), where we don't receive all the hashes for the files added.

I thought this could be an js-ipfs-api problem, but further testing (and some fixes) narrowed down the problem to how go-ipfs is responding, even when requests are 100% the same, the response may differ.

The failing failing test is: https://github.com/ipfs/interface-ipfs-core/blob/feat/ipfs.files.get/src/files.js#L113-L143 (again: just fails in the browser and some of the times)

(left -> success; center -> fails)
![image](https://cloud.githubusercontent.com/assets/1211152/17516545/d202fbe4-5e36-11e6-9053-c715442b6a17.png)

[Full request + response of a Success](https://github.com/ipfs/js-ipfs-api/files/408905/success.txt)
[Full request + response of a Failure](https://github.com/ipfs/js-ipfs-api/files/408906/failure.txt)
[wireshark pcap (filter for HTTP and then look for the POST /add](https://ipfs.io/ipfs/Qmctcru7zYWrkHsfBQHFN7M6e6EsNbfGPMVcarooV1Fmod)
## How to reproduce

``` sh
> git clone git@github.com:ipfs/interface-ipfs-core.git -b feat/ipfs.files.get 
> cd interface-ipfs-core
> npm i && npm run build && npm link
> cd ..
> git clone git@github.com:ipfs/js-ipfs-api.git -b feat/ipfs.files.get
> cd js-ipfs-api 
> npm i && npm link interface-ipfs-core
> npm run test:browser # repeat this one till it fails
```

whyrusleeping:
@diasdavid @dignifiedquire could you find a way to reproduce this with a simple curl command? I guess i'm just not very good at javascript but i have trouble even getting these tests to run, consistently hitting new and exciting problems.

diasdavid:
After we talked and you getting the test to run, are you still getting that issue?

whyrusleeping:
apparently i just needed to rerun it a few times for the tests to run. I was having issues with karma and chrome, but a few more tries and it started running the tests.

whyrusleeping:
A relevant note: this has only been reproduced in the browser testing.

whyrusleeping:
[...]

PR #340, opened 2016-08-10
feat(factory): use the factory mode

diasdavid:


test/api/files.spec.js
```diff
@@ -15,7 +15,21 @@ testfile = require('fs').readFileSync(path.join(__dirname, '/../testfile.txt'))
...
+        if (typeof path === 'function') {
+          callback = path
+          path = undefined
+        }
+        switch (c) {
+          case 0: callback(null, apiClients.a); c++; break
+          case 1: callback(null, apiClients.b); c++; break
+          case 2: callback(null, apiClients.c); c++; break
+          default: callback(new Error('no more nodes available'))
+        }
```
diasdavid: This is 'hacky' right now.

What I would love to be able to do is to have a an actual spawnNode function, that new if run on the browser or in Node.js and would spawn the node through a Karma service. Essentially, RMI.

diasdavid: @dignifiedquire does karma offer a way to do this?

dignifiedquire: Not without writing a decent amount of code.

dignifiedquire: What you could do is 

- create a simple socket.io server that can spawn nodes on a specific message
- in the spawn function, connect to the socket.io server and tell it to spawn a node, and get as a response the needed info, this spawn function would then work in the browser and in node

diasdavid: 👍 

It would be a sweet feature for Karma btw :)

dignifiedquire:
@diasdavid PRs welcome :)

diasdavid:
RFM, needs CR

diasdavid:
The CI errors are https://github.com/ipfs/js-ipfs-api/issues/339

test/factory/factory-server.js
```diff
@@ -0,0 +1,28 @@
...
+const options = {
+  connections: {
+    routes: {
+      cors: true
+    }
+  }
+}
+
+module.exports = (callback) => {
+  const http = new Hapi.Server(options)
```
dignifiedquire: Do we need to bring in hapi? just using core-http should be enough for serving socket.io

diasdavid: I like `hapi` because then I can just set CORS to true without having to write headers by hand. Also `hapi` is pretty modular, a lot of features (like template engine) are buy in modules, which makes it a 'smaller' framework

diasdavid:
Thank you for CR @dignifiedquire :) merging

PR #356, opened 2016-08-24
feat(block): make the block api follow the interface definition

diasdavid:
Needs: 
- [x] merged and released https://github.com/ipfs/interface-ipfs-core/pull/64

Note, this will be a major release in terms of API impl, but not in spec. Apologies for the lapse, this one passed under the radar


dignifiedquire:
LGTM

PR #359, opened 2016-08-24
fix(block): cope with the fact that sometimes go-ipfs sends x-stream,…

diasdavid:
… other times it doesn't


diasdavid:
super small change, going ahead and merge this, all tests passing :)

PR #379, opened 2016-09-14
fix: Return req from tail

VictorBjelkholm:
This fixes the broken master-build. Something worrying is that this doesn't always fail (the test for log.tail) when running all tests with `npm test`, but when running only that test, it always fails.

This PR should pass all CI though.


diasdavid:
@VictorBjelkholm nice! good catch!

PR #384, opened 2016-09-16
fix(get): properly handled nested content

dignifiedquire:
Ref https://github.com/ipfs/js-ipfs/issues/498

Needs https://github.com/ipfs/interface-ipfs-core/pull/88

### DIFF
--- a/src/tar-stream-to-objects.js
+++ b/src/tar-stream-to-objects.js
@@ -5,28 +5,36 @@ const Readable = require('readable-stream')
 
 // transform tar stream into readable stream of
 // { path: 'string', content: Readable }
-module.exports = function (err, res, send, done) {
+module.exports = (err, res, send, done) => {
   if (err) {
     return done(err)
   }
 
-  var ex = tar.extract()
+  const ex = tar.extract()
   res.pipe(ex)
 
-  var objStream = new Readable({ objectMode: true })
+  const objStream = new Readable({ objectMode: true })
   objStream._read = function noop () {}
 
-  ex.on('entry', function (header, stream, next) {
-    objStream.push({
-      path: header.name,
-      content: header.type !== 'directory' ? stream : null
-    })
-    next()
+  ex.on('entry', (header, stream, next) => {
+    stream.on('end', next)
+
+    if (header.type !== 'directory') {
+      objStream.push({
+        path: header.name,
+        content: stream
+      })
+    } else {
+      objStream.push({
+        path: header.name
+      })
+      stream.resume()
+    }
   })
+
   ex.on('finish', () => {
     objStream.push(null)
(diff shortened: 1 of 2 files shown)


src/tar-stream-to-objects.js
```diff
@@ -5,28 +5,36 @@ const Readable = require('readable-stream')
...
   objStream._read = function noop () {}
 
-  ex.on('entry', function (header, stream, next) {
-    objStream.push({
-      path: header.name,
-      content: header.type !== 'directory' ? stream : null
-    })
-    next()
+  ex.on('entry', (header, stream, next) => {
+    stream.on('end', next)
```
diasdavid: This means that now, one file has to finish in order for us to 'receive' the next. Not sure how this is an improvement,  the previous version was pushing a readable stream that gets filled as soon as we know the file exists.

dignifiedquire: This has to happen already if you read the documentation of tar-stream. This only makes sure that we actually process things correctly. The current implementation throws up and dies on larger sets of files.

diasdavid: Checking https://www.npmjs.com/package/tar-stream#extracting, it only says that "next" will emit the next entry, that being said, I should be able to get a readable stream for all the files within the tar first and then consume them.

diasdavid:
`.get` tests are failing with timeouts

```
  1) .files .get with a base58 encoded multihash:
     Error: timeout of 80000ms exceeded. Ensure the done() callback is being called in this test.
```

src/tar-stream-to-objects.js
dignifiedquire: @diasdavid you missed this part:

> The tar archive is streamed sequentially, meaning you must drain each entry's stream as you get them or else the main extract stream will receive backpressure and stop reading.

dignifiedquire:
> .get tests are failing with timeouts

Yes they can't finish, because they do not drain the content of the file for each received item, instead they try to do `concat` which waits for the end which never happens.

I suggest changing the interface-ipfs-core tests accordingly for now, as the current version is broken for everything with multiple files and the alternative is writing our own tar extractor.

diasdavid:
> I suggest changing the interface-ipfs-core tests accordingly for now, as the current version is broken for everything with multiple files and the alternative is writing our own tar extractor

@dignifiedquire I'm understanding what you mean, however, something is off because these tests were passing before, which means that calling `next` will indeed start the next stream, even if the current one hasn't flushed completely. It seems that it is a tar-stream documentation incoherence, which we kind of found by chance and that helped us achieve what we wanted.

dignifiedquire:
@diasdavid we still need to figure this one out

PR #385, opened 2016-09-17
Move dht tests from js-ipfs-api to interface-ipfs-core

VictorBjelkholm:
Needs https://github.com/ipfs/interface-ipfs-core/pull/36

### DIFF
--- a/test/interface-ipfs-core/dht.spec.js
+++ b/test/interface-ipfs-core/dht.spec.js
@@ -1,4 +1,5 @@
 /* eslint-env mocha */
+'use strict'
 
 const test = require('interface-ipfs-core')
 const FactoryClient = require('../factory/factory-client')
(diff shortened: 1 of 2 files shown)


diasdavid:
Needs:
 - [ ] https://github.com/ipfs/interface-ipfs-core/pull/36#issuecomment-247822522

diasdavid:
interface-ipfs-core 0.18.3 released

VictorBjelkholm:
current status in CI:
```
Message:
    Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
Stack:
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

Seems linting in interface-ipfs-core runs in different env than the tests for js-ipfs-api.

test/interface-ipfs-core/dht.spec.js
```diff
@@ -1,65 +1,18 @@
 /* eslint-env mocha */
-/* globals apiClients */
-'use strict'
```
dignifiedquire: you shouldn't delete use strict

diasdavid:
all right :) thank you @VictorBjelkholm

PR #386, opened 2016-09-18
feat(object): add lru cache to object get/put

diasdavid:
Quick add of an lru-cache to js-ipfs-api for orbit. To make this effort complete we need:
- [x] Make the data and links calls use the same cache and fallback to object get (this way we can take the RT to the node to fetch the entire node, instead of just fetching parts of it)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "is-ipfs": "^0.2.1",
     "isstream": "^0.1.2",
     "multiaddr": "^2.0.3",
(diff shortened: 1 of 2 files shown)


diasdavid:
This is ready

dignifiedquire:
LGTM

diasdavid:
@haadcode now js-ipfs-api does the caching automatically :)

PR #391, opened 2016-09-29
test: remove dependency on init doc hashes

### LINKED ISSUE #365: Assumptions about init-docs hashes
I can't quite pin down what the issue is, but I think it's something about the assumptions js-ipfs-api makes about the init-docs hashes. My observations:

- I updated the init-docs in go-ipfs, and the js-ipfs build is failing. [1] It looks like only the `.refs` tests fail, while other tests seem to add the init-docs themselves and use the hash resulting from that.
- The whole `npm test` run works fine locally. Does js-ipfs-api use the go-ipfs from `$PATH`? Does it generate ephemeral fs-repos? I do have the new init-docs in my `~/.ipfs` fs-repo, and the tests might use that fs-repo in one spot and succeed because of that?

I don't have a good idea but it's blocking ipfs/go-ipfs#2997 which is a trivial PR in itself but turned into a yak shaving.

[1] http://ci.ipfs.team:8111/viewLog.html?buildId=5376&buildTypeId=GoIpfs_JsIpfsApiTests&tab=buildLog#_focus=1948

dignifiedquire:
Fixes #365

### DIFF
--- a/test/interface-ipfs-core/ls.spec.js
+++ b/test/interface-ipfs-core/ls.spec.js
@@ -3,19 +3,35 @@
 
 const expect = require('chai').expect
 const isNode = require('detect-node')
+const waterfall = require('async/waterfall')
+const path = require('path')
+
 const FactoryClient = require('../factory/factory-client')
+
 describe('ls', function () {
+  if (!isNode) {
+    return
+  }
+
   let ipfs
   let fc
+  let folderHash
 
   before(function (done) {
     this.timeout(20 * 1000) // slow CI
     fc = new FactoryClient()
-    fc.spawnNode((err, node) => {
-      expect(err).to.not.exist
-      ipfs = node
-      done()
-    })
+    waterfall([
+      (cb) => fc.spawnNode(cb),
+      (node, cb) => {
+        ipfs = node
+        const filesPath = path.join(__dirname, '../data/test-folder')
+        ipfs.util.addFromFs(filesPath, { recursive: true }, cb)
+      },
+      (hashes, cb) => {
+        folderHash = hashes[hashes.length - 1].hash
+        cb()
+      }
+    ], done)
   })
 
   after((done) => {
@@ -23,16 +39,12 @@ describe('ls', function () {
   })
 
   it('should correctly retrieve links', function (done) {
-    if (!isNode) {
-      return done()
-    }
-
-    ipfs.ls('QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG', (err, res) => {
+    ipfs.ls(folderHash, (err, res) => {
       expect(err).to.not.exist
 
       expect(res).to.have.a.property('Objects')
       expect(res.Objects[0]).to.have.a.property('Links')
-      expect(res.Objects[0]).to.have.property('Hash', 'QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG')
+      expect(res.Objects[0]).to.have.property('Hash')
       done()
     })
   })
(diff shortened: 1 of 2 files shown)


dignifiedquire:
@lgierth @whyrusleeping this should fix your ci issues

test/interface-ipfs-core/ls.spec.js
```diff
@@ -3,36 +3,48 @@
...
-    }
-
-    ipfs.ls('QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG', (err, res) => {
+    ipfs.ls(folderHash, (err, res) => {
       expect(err).to.not.exist
 
       expect(res).to.have.a.property('Objects')
       expect(res.Objects[0]).to.have.a.property('Links')
-      expect(res.Objects[0]).to.have.property('Hash', 'QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG')
+      expect(res.Objects[0]).to.have.property('Hash')
```
diasdavid: the hash value is not being checked anymore.

test/interface-ipfs-core/ls.spec.js
```diff
@@ -3,36 +3,48 @@
...
-    })
+    waterfall([
+      (cb) => fc.spawnNode(cb),
+      (node, cb) => {
+        ipfs = node
+        const filesPath = path.join(__dirname, '../data/test-folder')
+        ipfs.util.addFromFs(filesPath, { recursive: true }, cb)
+      },
+      (hashes, cb) => {
+        folderHash = hashes[hashes.length - 1].hash
```
diasdavid: Although we should use the value returned to make sure we are fetching the correct one, we should also have the expected hash hard coded, we've captured bugs in the past because the browser was adding different stuff then what it was added in Node.js

test/interface-ipfs-core/ls.spec.js
```diff
@@ -57,13 +67,11 @@ describe('ls', function () {
...
     it('should correctly retrieve links', () => {
-      if (!isNode) return
-
-      return ipfs.ls('QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG')
+      return ipfs.ls(folderHash)
         .then((res) => {
           expect(res).to.have.a.property('Objects')
           expect(res.Objects[0]).to.have.a.property('Links')
-          expect(res.Objects[0]).to.have.property('Hash', 'QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG')
+          expect(res.Objects[0]).to.have.property('Hash')
```
diasdavid: same as above

test/interface-ipfs-core/ls.spec.js
```diff
@@ -57,13 +67,11 @@ describe('ls', function () {
 
   describe('promise', () => {
     it('should correctly retrieve links', () => {
-      if (!isNode) return
```
diasdavid: This works all in the browser now? Sweet!

dignifiedquire: no, I just moved the `isNode` check to the top of the test

test/interface-ipfs-core/ls.spec.js
dignifiedquire: okay

dignifiedquire: fixed

test/interface-ipfs-core/ls.spec.js
dignifiedquire: fixed

test/interface-ipfs-core/ls.spec.js
dignifiedquire: fixed

lgierth:
thanks <3

PR #399, opened 2016-10-25
npmcdn.com → unpkg.com

Mithgol:
The site https://npmcdn.com/ has moved to https://unpkg.com/ (to refrain from using npm's name because of its [trademark policy](https://www.npmjs.com/policies/trademark), I presume).

Their name's change should be reflected in README.


dignifiedquire:
Thank you

PR #400, opened 2016-10-25
some security measures (for people not ready to trust their CDNs)

Mithgol:
This addition to README provides yet another example of CDN usage that ensures maximum security (but references a specific version of IPFS API).

The following minor changes are also introduced:
- “the Browser” → “a web browser”
- “through `<script>` tag” → “from CDN” (a necessary clarification because browserified versions, mentioned above it, are also likely to be used through `<script>`)
- the CORS-related problem (below) is specifically mentioned as affecting CDN-based IPFS API as well as browserified


VictorBjelkholm:
Looks good, we should merge this @diasdavid

diasdavid:
Thank you :)

PR #414, opened 2016-11-08
refactor: replace wreck with raw request

dignifiedquire:


src/request-api.js
```diff
@@ -1,32 +1,54 @@
...
 
-function onRes (buffer, cb, uri) {
-  return (err, res) => {
-    if (err) {
-      return cb(err)
-    }
+function parseRaw (res, cb) {
+  res
+    .once('error', cb)
+    .pipe(concat((data) => cb(null, data)))
```
diasdavid: why not use `bl`?

dignifiedquire: Because concat does the right thing for streams, arrays and buffers

diasdavid: Thank you :) Will remind me of that the next time I need to do this

examples/browser-add/index.js
```diff
@@ -11,24 +11,25 @@ function store () {
...
+        console.log('successfully stored', file.hash)
+        display(file.hash)
+      }
     })
   })
 }
 
 function display (hash) {
-  ipfs.cat(hash, function (err, res) {
+  ipfs.cat(hash, {buffer: true}, function (err, res) {
```
diasdavid: It needs to be buffered?

examples/browser-add/package.json
```diff
@@ -4,15 +4,13 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "start": "browserify -t brfs index.js > bundle.js && http-server -a 127.0.0.1 -p 8888"
+    "start": "browserify index.js > bundle.js && http-server -a 127.0.0.1 -p 8888"
```
diasdavid: <3

examples/browser-add/index.js
diasdavid: I still think this `buffer: true` is one of the most misleading features. Can you add a comment that says:
"Buffer the whole response instead of streaming it"

diasdavid:
btw, you missed the chance of calling this brach 'reeeeekt!' ahah :D

PR #418, opened 2016-11-10
Green CI

diasdavid:


diasdavid:
hm.. apparently there is more than the nested-dir test to make this fail. browser tests also exist with 1 in my local machine, although they all pass

diasdavid:
done with https://github.com/ipfs/js-ipfs-api/pull/427

PR #445, opened 2016-11-23
fix: handle missing daemon errors

diasdavid:


diasdavid:
@pfrazee this seems to solve your issue, want to try it out?

pfrazee:
Yeah that solved it! Thanks

src/request-api.js
```diff
@@ -147,6 +147,10 @@ function requestAPI (config, options, callback) {
     headers: headers
   }, onRes(options.buffer, callback))
 
+  req.on('error', (err) => {
+    callback(err)
```
dignifiedquire: you should use to avoid having double callback calls

```js
const once = require('once')
callback = once(callback)
```

diasdavid: it is already done, see above :)

dignifiedquire: :)

pfrazee:
thanks!

Issue #450, opened 2016-11-27
Trouble adding folders with specific kinds of characters in the folder


elsehow:
Adding folders with `{recursive: true}` works on all of the folders I try, even big ones.

However, when adding a folder with characters like "[" or "(" in the name, it fails in a mysterious way:

```
> var ipfs = ipfsAPI('localhost', '5001', {protocol: 'http'}) 
undefined
> function cb (err, res) { console.log(err, res) }
undefined
> var path = '/Users/ffff/Music/John Coltrane - A Love Supreme (Vinyl) (1964) [V0]/'
undefined
> ipfs.util.addFromFs(path, {recursive: true}, cb)
null []
```
When I add this folder with `ipfs add -r`, it works with no problem.

nycoliver:
Relevant: https://github.com/ipfs/js-ipfs-api/issues/114

elsehow:
Thanks for the pointer @nycoliver.

I wrote a test that fails `tests/ipfs-api/util.spec.js`:

```javascript
  it('.fsAdd a directory with an odd name', (done) => {
    const filesPath = path.join(__dirname, '../fixtures/weird name folder [v0]')
    ipfs.util.addFromFs(filesPath, { recursive: true }, (err, result) => {
      expect(err).to.not.exist
      expect(result.length).to.be.above(8)
      done()
    })
  })
```

Will do a bit more digging when I get a chance

elsehow:
This issue happens becuase filenames are passed into a glob [see L47 in `get-files-stream.js`](https://github.com/ipfs/js-ipfs-api/blob/master/src/get-files-stream.js#L47). This issue is resolved when we escape the characters that have special, glob-y significance. PR coming soon!

diasdavid:
12.0.1 released with fix by @elsehow. Thank you @elsehow :D

PR #456, opened 2016-11-28
Handle directories with odd names

elsehow:
Previously, adding a directory with square brackets
resulted in a [strange, silent error](https://github.com/ipfs/js-ipfs-api/issues/450#issuecomment-263106356). This happened
because directory names were being passed to a file-
matching glob designed to recursively stream files from
the directory. When filenames contained meaningful glob
symbols (like [ or *), the glob matching would fail.

This commit resolves this problem by escaping glob
characters on filenames, before the filename is inserted
into the main glob in src/get-file-streams.js L48.
We add a dependency to glob-escape in the process,
and a new test in test/ipfs-api/util.spec.js.


diasdavid:
awesome @elsehow! thank you for fixing this :)

PR #462, opened 2016-12-02
Warn if incompatible go-ipfs version

SidHarder:
To solve the issue where the running daemon version is less then 0.4.0 I have added some code in index.js.  I test to see if version is less than 0.4.0 and if it is I throw and error.

If this seems reasonable I will wright some tests to finish things up,

Let me know what you think.


.gitignore
```diff
@@ -3,3 +3,4 @@
 test/setup/tmp-disposable-nodes-addrs.json
 dist
 coverage
+.DS_Store
```
diasdavid: You can ignore .DS_Store globally. See how here: http://stackoverflow.com/questions/18393498/gitignore-all-the-ds-store-files-in-every-folder-and-subfolder :)

src/index.js
```diff
@@ -40,7 +40,26 @@ function IpfsAPI (hostOrMultiaddr, port, opts) {
...
 
+  cmds.version(function (err, res) {
+    if (err) console.log(err)
+    var dotSplit = res.version.split('.')
+    if (dotSplit[0] === 0) {
+      if (dotSplit[1] < 4) {
+        throw new DaemonVersionError(res)
+      }
+    }
+  })
```
diasdavid: since this check is async and the creation of the instance is sync, that throw will be an uncaught exception, you would have to make the creation of ipfs api async so that error could be passed on the callback.

diasdavid:
@SidHarder in order for this to be achievable, it requires: https://github.com/ipfs/http-api-spec/issues/114 to happen

diasdavid:
@SidHarder I'm going to close this PR as the solution, once https://github.com/ipfs/http-api-spec/issues/114 is in, will look really different, hope you don't mind. Sorry for not having communicated well through the other issue is blocked on https://github.com/ipfs/http-api-spec/issues/114

PR #465, opened 2016-12-05
refactor(streams): Refactor response stream handling.

haadcode:
This PR will:

- Refactor the response stream handling.
- Add dagnode-stream to transform file results to DAGNode objects.
- Add stream-to-value to convert a response stream to a single value.
- Refactor request-api so that chunked JSON objects are not buffered anymore. This touches a bunch of the files pushing the transform function down to the individual commands.

The general idea in this PR is to  cleanup the architecture and implementation regards to how the http responses are handled. Previously some of it was done in request-api.js and some in the individual commands making request-api coupled with the commands' response type. I've moved the response transform logic down to the individual commands so that request-api is agnostic to what is returned to the consumer of the commands. I've also refactored individual types of transformations to their own files to make it clear, clean and much more reusable (we can consider moving them to their own modules, but not in this PR).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
   },
   "dependencies": {
     "async": "^2.1.4",
-    "bl": "^1.1.2",
     "bs58": "^4.0.0",
     "concat-stream": "^1.6.0",
     "detect-node": "^2.0.3",
@@ -43,6 +42,7 @@
     "peer-id": "^0.8.1",
     "peer-info": "^0.8.1",
     "promisify-es6": "^1.0.2",
+    "pump": "^1.0.2",
     "qs": "^6.3.0",
     "readable-stream": "1.1.14",
     "stream-http": "^2.5.0",
--- a/src/api/add.js
+++ b/src/api/add.js
@@ -1,24 +1,26 @@
 'use strict'
 
 const isStream = require('isstream')
-const addToDagNodesTransform = require('../add-to-dagnode-transform')
 const promisify = require('promisify-es6')
+const DAGNodeStream = require('../dagnode-stream')
 
 module.exports = (send) => {
   return promisify((files, callback) => {
-    const good = Buffer.isBuffer(files) ||
+    const ok = Buffer.isBuffer(files) ||
                isStream.isReadable(files) ||
                Array.isArray(files)
 
-    if (!good) {
-      callback(new Error('"files" must be a buffer, readable stream, or array of objects'))
+    if (!ok) {
+      return callback(new Error('"files" must be a buffer, readable stream, or array of objects'))
     }
 
-    const sendWithTransform = send.withTransform(addToDagNodesTransform)
-
-    return sendWithTransform({
+    const request = {
       path: 'add',
       files: files
-    }, callback)
+    }
+
+    // Transform the response stream to DAGNode values
+    const transform = (res, callback) => DAGNodeStream.streamToValue(send, res, callback)
+    send.andTransform(request, transform, callback)
   })
 }
--- a/src/dagnode-stream.js
+++ b/src/dagnode-stream.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const pump = require('pump')
 const TransformStream = require('readable-stream').Transform
 const streamToValue = require('./stream-to-value')
 const getDagNode = require('./get-dagnode')
@@ -31,7 +32,9 @@ class DAGNodeStream extends TransformStream {
   }
 
   static streamToValue (send, inputStream, callback) {
-    const outputStream = inputStream.pipe(new DAGNodeStream({ send: send }))
+    const outputStream = pump(inputStream, new DAGNodeStream({ send: send }), (err) => {
+      if (err) callback(err)
+    })
     streamToValue(outputStream, callback)
   }
 
--- a/src/dagnode-stream.js
+++ b/src/dagnode-stream.js
@@ -1,0 +1,55 @@
+'use strict'
+
+const TransformStream = require('readable-stream').Transform
+const streamToValue = require('./stream-to-value')
+const getDagNode = require('./get-dagnode')
+
+/*
+  Transforms a stream of objects to DAGNodes and outputs them as objects.
+
+  Usage: inputStream.pipe(DAGNodeStream({ send: send }))
+
+  Input object format:
+  {
+    Name: '/path/to/file/foo.txt',
+    Hash: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+  }
+
+  Output object format:
+  {
+    path: '/path/to/file/foo.txt',
+    hash: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP',
+    size: 20
+  }
+*/
+class DAGNodeStream extends TransformStream {
+  constructor (options) {
+    const opts = Object.assign(options || {}, { objectMode: true })
+    super(opts)
+    this._send = opts.send
+  }
+
+  static streamToValue (send, inputStream, callback) {
+    const outputStream = inputStream.pipe(new DAGNodeStream({ send: send }))
+    streamToValue(outputStream, callback)
+  }
+
+  _transform (obj, enc, callback) {
+    getDagNode(this._send, obj.Hash, (err, node) => {
+      if (err) {
+        return callback(err)
+      }
+
+      const dag = {
+        path: obj.Name,
+        hash: obj.Hash,
+        size: node.size
+      }
+
+      this.push(dag)
+      callback(null)
+    })
+  }
+}
+
+module.exports = DAGNodeStream
--- a/src/dagnode-stream.js
+++ b/src/dagnode-stream.js
@@ -33,7 +33,9 @@ class DAGNodeStream extends TransformStream {
 
   static streamToValue (send, inputStream, callback) {
     const outputStream = pump(inputStream, new DAGNodeStream({ send: send }), (err) => {
-      if (err) callback(err)
+      if (err) {
+        callback(err)
+      }
     })
     streamToValue(outputStream, callback)
   }
--- a/src/dagnode-stream.js
+++ b/src/dagnode-stream.js
@@ -41,13 +41,13 @@ class DAGNodeStream extends TransformStream {
         return callback(err)
       }
 
-      const dag = {
+      const result = {
         path: obj.Name,
         hash: obj.Hash,
         size: node.size
       }
 
-      this.push(dag)
+      this.push(result)
       callback(null)
     })
   }
--- a/src/dagnode-stream.js
+++ b/src/dagnode-stream.js
@@ -5,7 +5,8 @@ const streamToValue = require('./stream-to-value')
 const getDagNode = require('./get-dagnode')
 
 /*
-  Transforms a stream of objects to DAGNodes and outputs them as objects.
+  Transforms a stream of {Name, Hash} objects to include size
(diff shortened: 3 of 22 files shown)


src/dagnode-stream.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+*/
+class DAGNodeStream extends TransformStream {
+  constructor (options) {
+    const opts = Object.assign(options || {}, { objectMode: true })
+    super(opts)
+    this._send = opts.send
+  }
+
+  static streamToValue (send, inputStream, callback) {
```
dignifiedquire: is `static` available in Node4?

haadcode: Is it? Is it not? And if so, why is that important. Provide context and solutions or guidance.

src/api/add.js
```diff
@@ -1,24 +1,26 @@
...
 module.exports = (send) => {
   return promisify((files, callback) => {
-    const good = Buffer.isBuffer(files) ||
+    const ok = Buffer.isBuffer(files) ||
                isStream.isReadable(files) ||
                Array.isArray(files)
 
-    if (!good) {
-      callback(new Error('"files" must be a buffer, readable stream, or array of objects'))
+    if (!ok) {
```
diasdavid: unnecessary change

src/api/add.js
```diff
@@ -1,24 +1,26 @@
...
+      return callback(new Error('"files" must be a buffer, readable stream, or array of objects'))
```
diasdavid: good catch

src/tar-stream-to-objects.js
```diff
@@ -1,38 +1,45 @@
...
+          })
+          stream.resume()
+        }
+      })
+      .on('finish', () => outputStream.push(null))
+
+    callback(null, outputStream)
+  }
+
+  _read () {}
```
diasdavid: A Class with a constructor and a static method? What is the goal here?

src/dagnode-stream.js
```diff
@@ -0,0 +1,55 @@
...
+  Input object format:
+  {
+    Name: '/path/to/file/foo.txt',
+    Hash: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP'
+  }
+
+  Output object format:
+  {
+    path: '/path/to/file/foo.txt',
+    hash: 'Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP',
```
diasdavid: hash or `multihash`?

diasdavid: This "Output object format" is not a DAGNode `Transforms a stream of objects to DAGNodes and outputs them as objects.`, what am I missing?

src/request-api.js
```diff
@@ -173,33 +144,21 @@ exports = module.exports = function getRequestAPI (config) {
...
     if (typeof options !== 'object') {
       return callback(new Error('no options were passed'))
     }
 
     return requestAPI(config, options, callback)
   }
 
-  // Wraps the 'send' function such that an asynchronous
-  // transform may be applied to its result before
-  // passing it on to either its callback or promise.
```
diasdavid: can you preserve the comment please?

src/stream-to-value.js
```diff
@@ -0,0 +1,12 @@
...
+const concat = require('concat-stream')
+
+/*
+  Concatenate a stream to a single value.
+*/
+function streamToValue (res, callback) {
+  res.pipe(concat((data) => callback(null, data)))
+}
+
+module.exports = streamToValue
```
diasdavid: In case of error on the stream (connection dropped) this fails silently

src/stringlist-to-array.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+// Converts a go-ipfs "stringList" to an array
+// { Strings: ['A', 'B'] } --> ['A', 'B']
+function stringlistToArray (res, cb) {
+  cb(null, res.Strings || [])
+}
+
+module.exports = stringlistToArray
```
diasdavid: What is a string list? Is there a better name in this context?  Does it has to be a separate file to do this?

haadcode:
Thanks for the review @dignifiedquire and @diasdavid. Will work on this soon, mainly change it to use pump as well as address other CR pieces.

src/request-api.js
haadcode: This comment doesn't make sense, and it doesn't hold true for the new implementation.

src/dagnode-stream.js
diasdavid: 🛎

src/dagnode-stream.js
```diff
@@ -0,0 +1,55 @@
...
+    getDagNode(this._send, obj.Hash, (err, node) => {
+      if (err) {
+        return callback(err)
+      }
+
+      const dag = {
+        path: obj.Name,
+        hash: obj.Hash,
+        size: node.size
+      }
```
diasdavid: Not really a `dag`, it is just a json version.

src/stringlist-to-array.js
diasdavid: @whyrusleeping wanna comment?

src/tar-stream-to-objects.js
```diff
@@ -1,38 +1,45 @@
...
-    .on('finish', () => {
-      objStream.push(null)
-    })
-
-  done(null, objStream)
+  static from (inputStream, callback) {
+    let outputStream = new TarStreamToObjects()
+
+    inputStream
+      .pipe(tar.extract())
```
diasdavid: better to use `pump`

src/request-api.js
diasdavid: It was trying to convey that before the response is returned, either through the callback or promise, we transform the response from go-ipfs to match interface-ipfs-core definition.

haadcode:
Will update this today and address the CR feedback.

src/dagnode-stream.js
haadcode: Changed the description. Does that make sense now?

src/request-api.js
haadcode: Added the description back and trying to describe as simple as I can what it does. What do you think?

haadcode: Here: https://github.com/ipfs/js-ipfs-api/pull/465/commits/1566751ee3c703dfc18a5be59bcf8f5d4d48a580

src/dagnode-stream.js
haadcode: Here: https://github.com/ipfs/js-ipfs-api/pull/465/commits/1566751ee3c703dfc18a5be59bcf8f5d4d48a580

src/stringlist-to-array.js
haadcode: We shouldn't change this on go-ipfs side for this PR. If we do that, it's an API breaking change that we need to handle with proper care. That's why I put it in one file, so that it's wrapped and isolated and if/once we change the return values, we only need to change change it in this file and whichever command is using the current API doesn't need changes. Obviously once you do that, we don't need the stringlist-to-array.js conversion function and can get rid of the file completely, but I recommend we keep it as such now so that we isolate the data conversion to one place. Does that make sense?

src/dagnode-stream.js
haadcode: @diasdavid I'm not sure about the `multihash` vs. `hash`. It's a breaking change again and I'd be hesitant to change it in this PR. Have we made the change elsewhere in the code already? Does anything that returns DAGNode-like information return .multihash instead of .hash? If we have, I can change it so that we're consistent, but if not, I would leave it for later.

haadcode:
I've refactored the pipes to use pump. I wasn't sure if the pump is needed in dagnode-stream.js since streamToValue internally uses pump (so I take it would handle the errors), but I added it just in case. The tests are not running in the browser atm, but run fully on node (will look into why the browser tests are not running).

I also reworked the comments for the two functions @diasdavid had a confusion about, hope they're now more descriptive. I refactored the TarStreamToObjects according to @diasdavid's question, perhaps now the intent is clearer. I couldn't get the ReadableStream to work when created directly with `new Readable() and adding `_read() = () => {}` function to it like it's normally done, but by wrapping it in a class it works. If you have an idea how to make it work without the class, please do refactor it.

There's also a problem with browser version now as we're using pump because its dependency on `fs`. @dignifiedquire suggested we fix it there: https://github.com/mafintosh/pump/issues/6. If you know how, please do.

So while this is still WIP, can you take a look and see if there's anything else to address in this one?

dignifiedquire:
@haadcode pump fix is here: https://github.com/mafintosh/pump/pull/13 and the issue with cors was correct, you changed them, now they are back: https://github.com/ipfs/js-ipfs-api/pull/465/commits/b2c09b2a40cc2761dc767abb0a6400d1c24df859

haadcode:
Should we wait for the PR to get merged or can we use the PR commit as our dep in the meanwhile?

> the issue with cors was correct, you changed them, now they are back: b2c09b2

Interesting. I had to change it in the first place because it wasn't working. Good that it's fixed now 👍

dignifiedquire:
> Should we wait for the PR to get merged or can we use the PR commit as our dep in the meanwhile?

Pinged @mafintosh on irc, if it's not merged when we want to ship this lets switch to the fork.

package.json
```diff
@@ -43,6 +42,7 @@
     "peer-id": "^0.8.1",
     "peer-info": "^0.8.1",
     "promisify-es6": "^1.0.2",
+    "pump": "^1.0.1",
```
dignifiedquire: lets bump this to `1.0.2` so we are sure to have the fix in

src/dagnode-stream.js
```diff
@@ -0,0 +1,59 @@
...
+class DAGNodeStream extends TransformStream {
+  constructor (options) {
+    const opts = Object.assign(options || {}, { objectMode: true })
+    super(opts)
+    this._send = opts.send
+  }
+
+  static streamToValue (send, inputStream, callback) {
+    const outputStream = pump(inputStream, new DAGNodeStream({ send: send }), (err) => {
+      if (err) callback(err)
```
dignifiedquire: @diasdavid will want this to be 

```js
if (err) {
  callback(err)
}

src/api/add.js
diasdavid: fine :)

src/request-api.js
diasdavid: 👍

src/stringlist-to-array.js
diasdavid: Understood :)

haadcode:
I fixed the style issues, updated pump to 1.0.2 and rebased on master.

> Other than the hash vs multihash thing, it LGTM. I'm fine if Files API uses hash, although I would prefer consistency.

@diasdavid would you be ok if we leave the change from `.hash` to `.multihash` to a separate PR? This is a large PR in itself and I'm afraid of the consequences the aforementioned change would have on users (being an API change). Perhaps you have more confidence to do it as you know the codebase throughout, but I would prefer doing it in one, separate PR where we update *all* functions (that return `.hash` currently) at the same time. Let me know what you think.

diasdavid:
> @diasdavid would you be ok if we leave the change from .hash to .multihash to a separate PR? 

sounds good :)

dignifiedquire:
@diasdavid did you pull in the latest version of #465 before merging?

dignifiedquire:
Also this is using ipfsd-ctl fork from @haadcode I though we wanted to wait for that release and switch to happen

haadcode:
Didn't even notice this got merged. Thank you! 😄  🎉

PR #471, opened 2016-12-06
WIP: feat/pubsub

diasdavid:
> **From @haadcode https://github.com/ipfs/js-ipfs-api/pull/470**

This PR will add complete go-ipfs Pubsub API.

Missing:
- Documentation

Dependencies:
- https://github.com/ipfs/js-ipfs-api/pull/465
- Need to have go-ipfs 0.4.5-x released so that there's a binary that can be used with this api

PR #493, opened 2016-12-21
feat(pubsub): Add pubsub api

dignifiedquire:
Rebased onto master and squashed #471

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,30 @@ $ ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods "[\"PUT\", \"P
 
 > `js-ipfs-api` follows the spec defined by [`interface-ipfs-core`](https://github.com/ipfs/interface-ipfs-core), which concerns the interface to expect from IPFS implementations. This interface is a currently active endeavor. You can use it today to consult the methods available.
 
+#### Caveats
+
+##### Pubsub
+
+**Currently, the [PubSub API only works in Node.js envinroment](https://github.com/ipfs/js-ipfs-api/issues/518)**
+
+We currently don't support pubsub when run in the browser, and we test it with separate set of tests to make sure if it's being used in the browser, pubsub errors.
+
+More info: https://github.com/ipfs/js-ipfs-api/issues/518
+
+This means:
+- You can use pubsub from js-ipfs-api in Node.js
+- You can use pubsub from js-ipfs-api in Electron
+  (when js-ipfs-api is ran in the main process of Electron)
+- You can't use pubsub from js-ipfs-api in the browser
+- You can't use pubsub from js-ipfs-api in Electron's
+  renderer process
+- You can use pubsub from js-ipfs in the browsers
+- You can use pubsub from js-ipfs in Node.js
+- You can use pubsub from js-ipfs in Electron
+  (in both the main process and the renderer process)
+- See https://github.com/ipfs/js-ipfs for details on
+  pubsub in js-ipfs
+
 ##### [bitswap]()
 
 - [`ipfs.bitswap.wantlist()`]()
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "eslint-plugin-react": "^6.10.3",
     "gulp": "^3.9.1",
     "hapi": "^16.1.0",
-    "interface-ipfs-core": "~0.26.1",
+    "interface-ipfs-core": "~0.26.2",
     "ipfsd-ctl": "~0.20.0",
     "pre-commit": "^1.2.2",
     "socket.io": "^1.7.3",
--- a/src/api/pubsub.js
+++ b/src/api/pubsub.js
@@ -1,0 +1,162 @@
+'use strict'
+
+const promisify = require('promisify-es6')
+const EventEmitter = require('events')
+const eos = require('end-of-stream')
+const isNode = require('detect-node')
+const PubsubMessageStream = require('../pubsub-message-stream')
+const stringlistToArray = require('../stringlist-to-array')
+
+const NotSupportedError = () => new Error('pubsub is currently not supported when run in the browser')
+
+/* Public API */
+module.exports = (send) => {
+  /* Internal subscriptions state and functions */
+  const ps = new EventEmitter()
+  const subscriptions = {}
+  ps.id = Math.random()
+  return {
+    subscribe: (topic, options, handler, callback) => {
+      const defaultOptions = {
+        discover: false
+      }
+
+      if (typeof options === 'function') {
+        callback = handler
+        handler = options
+        options = defaultOptions
+      }
+
+      if (!options) {
+        options = defaultOptions
+      }
+
+      // Throw an error if ran in the browsers
+      if (!isNode) {
+        if (!callback) {
+          return Promise.reject(NotSupportedError())
+        }
+        return callback(NotSupportedError())
+      }
+
+      // promisify doesn't work as we always pass a
+      // function as last argument (`handler`)
+      if (!callback) {
+        return new Promise((resolve, reject) => {
+          subscribe(topic, options, handler, (err) => {
+            if (err) {
+              return reject(err)
+            }
+            resolve()
+          })
+        })
+      }
+
+      subscribe(topic, options, handler, callback)
+    },
+    unsubscribe: (topic, handler) => {
+      if (!isNode) {
+        throw NotSupportedError()
+      }
+
+      if (ps.listenerCount(topic) === 0 || !subscriptions[topic]) {
+        throw new Error(`Not subscribed to '${topic}'`)
+      }
+
+      ps.removeListener(topic, handler)
+
+      // Drop the request once we are actualy done
+      if (ps.listenerCount(topic) === 0) {
+        subscriptions[topic].abort()
+        subscriptions[topic] = null
+      }
+    },
+    publish: promisify((topic, data, callback) => {
+      if (!isNode) {
+        return callback(NotSupportedError())
+      }
+
+      if (!Buffer.isBuffer(data)) {
+        return callback(new Error('data must be a Buffer'))
+      }
+
+      const request = {
+        path: 'pubsub/pub',
+        args: [topic, data]
+      }
+
+      send(request, callback)
+    }),
+    ls: promisify((callback) => {
+      if (!isNode) {
+        return callback(NotSupportedError())
+      }
+
+      const request = {
+        path: 'pubsub/ls'
+      }
+
+      send.andTransform(request, stringlistToArray, callback)
+    }),
+    peers: promisify((topic, callback) => {
+      if (!isNode) {
+        return callback(NotSupportedError())
+      }
+
+      const request = {
+        path: 'pubsub/peers',
+        args: [topic]
+      }
+
+      send.andTransform(request, stringlistToArray, callback)
+    }),
+    setMaxListeners (n) {
+      return ps.setMaxListeners(n)
+    }
+  }
+
+  function subscribe (topic, options, handler, callback) {
+    ps.on(topic, handler)
+    if (subscriptions[topic]) {
+      return callback()
+    }
+
+    // Request params
+    const request = {
+      path: 'pubsub/sub',
+      args: [topic],
+      qs: {
+        discover: options.discover
+      }
+    }
+
+    // Start the request and transform the response
+    // stream to Pubsub messages stream
+    subscriptions[topic] = send.andTransform(request, PubsubMessageStream.from, (err, stream) => {
+      if (err) {
+        subscriptions[topic] = null
+        ps.removeListener(topic, handler)
+        return callback(err)
+      }
+
+      stream.on('data', (msg) => {
+        ps.emit(topic, msg)
+      })
+
+      stream.on('error', (err) => {
+        ps.emit('error', err)
+      })
+
+      eos(stream, (err) => {
+        if (err) {
+          ps.emit('error', err)
+        }
+
+        subscriptions[topic] = null
+        ps.removeListener(topic, handler)
+      })
+
+      callback()
+    })
+  }
+}
--- a/src/pubsub-message-stream.js
+++ b/src/pubsub-message-stream.js
@@ -1,0 +1,33 @@
+'use strict'
+
+const TransformStream = require('readable-stream').Transform
+const PubsubMessage = require('./pubsub-message-utils')
+
+class PubsubMessageStream extends TransformStream {
+  constructor (options) {
+    const opts = Object.assign(options || {}, { objectMode: true })
+    super(opts)
+  }
+
+  static from (inputStream, callback) {
+    let outputStream = inputStream.pipe(new PubsubMessageStream())
+    inputStream.on('end', () => outputStream.emit('end'))
+    callback(null, outputStream)
+  }
+
+  _transform (obj, enc, callback) {
+    let msg
+    try {
+      msg = PubsubMessage.deserialize(obj, 'base64')
+    } catch (err) {
+      // Not a valid pubsub message
+      // go-ipfs returns '{}' as the very first object atm, we skip that
+      return callback()
+    }
+
+    this.push(msg)
+    callback()
+  }
+}
+
+module.exports = PubsubMessageStream
--- a/test/pubsub-in-browser.spec.js
+++ b/test/pubsub-in-browser.spec.js
@@ -1,0 +1,182 @@
+/*
+  We currently don't support pubsub when run in the browser,
+  and we test it with separate set of tests to make sure
+  if it's being used in the browser, pubsub errors.
+
+  More info: https://github.com/ipfs/js-ipfs-api/issues/518
+
+  This means:
+  - You can use pubsub from js-ipfs-api in Node.js
+  - You can use pubsub from js-ipfs-api in Electron
+    (when js-ipfs-api is ran in the main process of Electron)
+
+  - You can't use pubsub from js-ipfs-api in the browser
+  - You can't use pubsub from js-ipfs-api in Electron's
+    renderer process
+
+  - You can use pubsub from js-ipfs in the browsers
+  - You can use pubsub from js-ipfs in Node.js
+  - You can use pubsub from js-ipfs in Electron
+    (in both the main process and the renderer process)
+  - See https://github.com/ipfs/js-ipfs for details on
+    pubsub in js-ipfs
+*/
+
+/* eslint-env mocha */
+/* eslint max-nested-callbacks: ['error', 8] */
+'use strict'
+
+const series = require('async/series')
+const waterfall = require('async/waterfall')
+const isNode = require('detect-node')
+const FactoryClient = require('./ipfs-factory/client')
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+
+const expectedError = 'pubsub is currently not supported when run in the browser'
+
+function spawnWithId (factory, callback) {
+  waterfall([
+    (cb) => factory.spawnNode(cb),
+    (node, cb) => node.id((err, res) => {
+      if (err) {
+        return cb(err)
+      }
+      node.peerId = res
+      cb(null, node)
+    })
+  ], callback)
+}
+
+if (!isNode) {
+  describe('.pubsub-browser (pubsub not supported in the browsers currently)', () => {
+    const topic = 'pubsub-tests'
+
+    let factory
+    let ipfs1
+
+    before((done) => {
+      factory = new FactoryClient()
+
+      series([
+        (cb) => spawnWithId(factory, cb)
+      ], (err, nodes) => {
+        if (err) {
+          return done(err)
+        }
+
+        ipfs1 = nodes[0]
+        done()
+      })
+    })
+
+    after((done) => {
+      factory.dismantle(done)
+    })
+
+    describe('everything errors', () => {
+      describe('Callback API', () => {
+        describe('.publish', () => {
+          it('throws an error if called in the browser', (done) => {
+            ipfs1.pubsub.publish(topic, 'hello friend', (err, topics) => {
+              expect(err).to.exist()
+              expect(err.message).to.equal(expectedError)
+              done()
+            })
+          })
+        })
+
+        describe('.subscribe', () => {
+          const handler = () => {}
+          it('throws an error if called in the browser', (done) => {
+            ipfs1.pubsub.subscribe(topic, {}, handler, (err, topics) => {
+              expect(err).to.exist()
+              expect(err.message).to.equal(expectedError)
+              done()
+            })
+          })
+        })
+
+        describe('.peers', () => {
+          it('throws an error if called in the browser', (done) => {
+            ipfs1.pubsub.peers(topic, (err, topics) => {
+              expect(err).to.exist()
+              expect(err.message).to.equal(expectedError)
+              done()
+            })
+          })
+        })
+
+        describe('.ls', () => {
+          it('throws an error if called in the browser', (done) => {
+            ipfs1.pubsub.ls((err, topics) => {
+              expect(err).to.exist()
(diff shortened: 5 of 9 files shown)


dignifiedquire:
cc @diasdavid @haadcode

diasdavid:
Just tested it with new updates deps, tests pass in Node.js but seems to hang in the browser.

dignifiedquire:
> Just tested it with new updates deps, tests pass in Node.js but seems to hang in the browser.

Got them running again locally, but seeing some other errors, probably something got lost in the rebase, investigating

dignifiedquire:
Fixed browser tests, in so far that they run through now. Looks like some of the pubsub tests are not passing yet though, probably due to timing issues with request abortion and `unsubscribe` calls.

dignifiedquire:
The browser tests are failing because `fetch` requests can not be aborted. I tried switching to forced `xhr`, but they seem to be entirely failing and I don't understand why. 

Before we ship this I would like to propose a change to this api, as I expect that this will cause more headaches down the line. 

So the question is to what to change the `subscribe` api over http. I know of three ideas so far

1. Use `FIN` packets and leave the connection hanging or abort on the server side
2. Use websockets (ideally with heartbeat support + fin packet) 
3. Use `/unsubscribe` to manually unsubscribe, and get the content from `/topic?arg="topicname"` 

cc @diasdavid @haadcode @jbenet @whyrusleeping

whyrusleeping:
It sounds like we just need to figure out how to get the xhr working, you can abort those.

dignifiedquire:
xhr can not do streaming as far as I understand, meaning we will only get the response body once it's done. So making it effectively useless for us in this case, as you subscribe to get content

dignifiedquire:
I've collected some interfaces from other places

- RabbitMq: Uses [AMQP](http://www.amqp.org/) which can be used over sockjs and websockets (https://www.rabbitmq.com/blog/category/web-messaging/)
- Google Pubsub: https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/pull uses a pulling mechanism
- Firebase, seems to provide some sort of streaming over http requests: https://firebase.google.com/docs/reference/rest/database/#section-streaming

diasdavid:
### 2017 is the year of the new HTTP(+WebSockets) API 🎆 https://github.com/ipfs/http-api-spec/issues/116

diasdavid:
What are people thoughts here? Wanna consider disabling browser tests until http-api is updated, enabling people to use pubsub from Node.js?

diasdavid:
**Update**: https://github.com/ipfs/go-ipfs/issues/3522#issuecomment-273250082

diasdavid:
Just tested with all the assertions possible and with regards to pubsub, it is 👌🏾

The only thing failing now, is name.resolve, but that seems like a new issue:

```
  1) .name .name.resolve:

      Uncaught AssertionError: expected { Object (Path) } to deeply equal { Object (Path) }
      + expected - actual

       {
      -  "Path": "/ipfs/Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP"
      +  "Path": "/ipfs//ipfs/Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP"
       }
```

@whyrusleeping

diasdavid:
@haadcode what's the plan to move this forward?

diasdavid:
From IRC
```
13:12 <@daviddias> dignifiedquire: victorbjelkholm haad should we merge this and stamp browser pubsub as not supported through an issue? https://github.com/ipfs/js-ipfs-api/pull/493
13:16 <@dignifiedquire> daviddias: I think so
13:18 <@daviddias> thank you
13:28 <@victorbjelkholm> daviddias: could we possibly do something for users trying to use pubsub in the browser? Like if you start subscribing and in browser, make a console.log with information about it
```

diasdavid:
Just realized going through the code again, that this is actually implementing the OLD API of pubsub in interface core. Needs more work then, I'll just squash all the commits for now to make rebasing nicer.

# TODO

- [ ] Update interface to match latest interface-ipfs-core

haadcode:
> what's the plan to move this forward?

As mentioned in the call on Monday, I'll work on this in this sprint. Most likely next week. I'll bring the API to match the one of `interface-ipfs-core`, make sure all tests pass and bring it to a point where it's mergeable.

> 13:12 <@daviddias> dignifiedquire: victorbjelkholm haad should we merge this and stamp browser pubsub as not supported through an issue? https://github.com/ipfs/js-ipfs-api/pull/493

I'd be totally fine doing this :+1:

> 13:28 <@victorbjelkholm> daviddias: could we possibly do something for users trying to use pubsub in the browser? Like if you start subscribing and in browser, make a console.log with information about it

We could do that, yeah. I'm thinking we can detect browser with `if (typeof window !== 'undefined')` and if true, we'll console.warn it. Is there a better way for the detection? @VictorBjelkholm @diasdavid @dignifiedquire does that sound like a good approach?

dignifiedquire:
we have a package `is-node` for doing exactly this

README.md
```diff
@@ -137,7 +137,12 @@ $ ipfs config --json API.HTTPHeaders.Access-Control-Allow-Methods "[\"PUT\", \"P
 
 ### API
 
-> `js-ipfs-api` follows the spec defined by [`interface-ipfs-core`](https://github.com/ipfs/interface-ipfs-core), which concerns the interface to expect from IPFS implementations. This interface is a currently active endeavor - expect it to be complete in the next few weeks (August 2016). You can use it today to consult the methods available.
+> `js-ipfs-api` follows the spec defined by [`interface-ipfs-core`](https://github.com/ipfs/interface-ipfs-core), which concerns the interface to expect from IPFS implementations. . You can use it today to consult the methods available.
```
diasdavid: extra `.`

package.json
```diff
@@ -11,7 +11,7 @@
   },
   "scripts": {
     "test": "gulp test",
-    "test:node": "gulp test:node",
+    "test:node": "TEST_PUBSUB=true gulp test:node",
```
diasdavid: can't PubSub tests be enabled by default/

package.json
```diff
@@ -63,10 +63,12 @@
     "aegir": "^10.0.0",
     "chai": "^3.5.0",
     "eslint-plugin-react": "^6.9.0",
+    "go-ipfs-dep": "^0.4.5",
     "gulp": "^3.9.1",
     "hapi": "^16.1.0",
-    "interface-ipfs-core": "~0.24.1",
+    "interface-ipfs-core": "github:ipfs/interface-ipfs-core#fix\/pubsub-tests",
```
diasdavid: can I get a PR on interface-ipfs-core and a note here that says to use that one? Remember how this 'pointing to branches and commits' has bitten us in the past, let's stop using them as we talked in January.

package.json
```diff
@@ -63,10 +63,12 @@
...
     "ipfsd-ctl": "~0.18.2",
+    "is-node": "^1.0.2",
```
diasdavid: we use `detect-node` in all the modules, it is already a dep. See line 29

test/name.spec.js
```diff
@@ -99,7 +99,7 @@ describe('.name', () => {
         .then((res) => {
           expect(res).to.exist
           expect(res).to.be.eql({
-            Path: '/ipfs/' + name.Value
+            Path: name.Value
```
diasdavid: This sounds like a breaking change in go-ipfs land. @whyrusleeping why has go-ipfs stopped returning the resolved hash with /ipfs/?

package.json
haadcode: Sure we ca do that too

package.json
haadcode: See comments below. Added the link to the PR.

haadcode:
This PR currently depends on https://github.com/ipfs/interface-ipfs-core/pull/117.

package.json
haadcode: Changed in the latest commit

package.json
haadcode: Enabled the pubsub tests by default in the latest commit

README.md
haadcode: This came from the rebase I think. Fixed it in the latest commit.

package.json
```diff
@@ -63,9 +63,10 @@
     "aegir": "^10.0.0",
     "chai": "^3.5.0",
     "eslint-plugin-react": "^6.9.0",
+    "go-ipfs-dep": "^0.4.5",
```
haadcode: @diasdavid also note that this PR pulls in go-ipfs 0.4.5 as ipfsd-ctl doesn't have it yet. We need to remove this line once we've updated the ecosystem modules to 0.4.5.

package.json
haadcode: To clarify, I'm pulling in interface-ipfs-core from a branch to make sure the tests run on CI. Before merging, we need that PR merged and use the regular version for the module. Ie. this is not intended as final.

test/name.spec.js
haadcode: @diasdavid I think it is now actually returning the value with `/ipfs/` in the path. The two lines here starting with `>` are the result value from go-ipfs http request:

```
    Promise API
> { Name: 'QmQ8qViFq8ke3vQVV3Qfb84GwPxmGAUm1a8TzDPD3u6GM1',
  Value: '/ipfs/Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP' }
      ✓ .name.publish
> { Name: 'QmQ8qViFq8ke3vQVV3Qfb84GwPxmGAUm1a8TzDPD3u6GM1',
  Value: '/ipfs/Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP' }
      1) .name.resolve
```

Is it the case that previously it returned just the hash without the `/ipfs/` in front and now it's returning it correctly (as expected in the tests, ie. we don't need to prepend it to the return value anymore).

package.json
diasdavid: yeah, can we do that first? https://github.com/ipfs/js-ipfsd-ctl/pull/145

haadcode: Absolutely!

haadcode:
Ok to recap the dependencies, in order of which they need to be merged:

1. https://github.com/ipfs/js-ipfs-api/pull/525
2. https://github.com/ipfs/js-ipfs-api/pull/526
3. https://github.com/ipfs/interface-ipfs-core/pull/117

And then one more review for the changes here.

After that, we need to remove the PR deps from this PRs `package.json`, make sure CI is green and we should be good to go! 👍

package.json
diasdavid: We can now use go-ipfs 0.4.6 \o/

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+  - You can't use pubsub from js-ipfs-api in Electron's
+    renderer process
+
+  - You can use pubsub from js-ipfs in the browsers
+  - You can use pubsub from js-ipfs in Node.js
+  - You can use pubsub from js-ipfs in Electron
+    (in both the main process and the renderer process)
+  - See https://github.com/ipfs/js-ipfs for details on
+    pubsub in js-ipfs
+*/
```
diasdavid: This should be in the README

test/interface/pubsubs.spec.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-env mocha */
```
diasdavid: This file has a typo on the name `s/pubsubs/pubsub`

test/name.spec.js
diasdavid: 👍

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+            done('unsubscribe() didn\'t throw an error')
+          } catch (e) {
+            expect(e).to.exist
+            expect(e.message).to.equal(expectedError)
+            done()
+          }
+        })
+      })
+    })
+  })
```
diasdavid: Are these tests testing something that is not tested in the interface-ipfs-core? They all seem to be interface expectation tests, if they are missing in interface-ipfs-core, then they should be present there.

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+          })
+        })
+      })
+
+      describe('.unsubscribe', () => {
+        it('throws an error', (done) => {
+          try {
+            ipfs1.pubsub.unsubscribe(topic)
+            done('unsubscribe() didn\'t throw an error')
+          } catch (e) {
```
diasdavid: Please always use `err`, so that `standard` linter catches if we don't check fot it.

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+              })
+          })
+        })
+      })
+
+      describe('.unsubscribe', () => {
+        it('throws an error', (done) => {
+          try {
+            ipfs1.pubsub.unsubscribe(topic)
+            done('unsubscribe() didn\'t throw an error')
```
diasdavid: I don't think you have to pass a string here

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+              .catch((e) => {
+                expect(e).to.exist
+                expect(e.message).to.equal(expectedError)
+                done()
+              })
+          })
+        })
+
+        describe('.ls', () => {
+          it('throws an error', () => {
```
diasdavid: Why does it throw an error on `ls`?

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+                expect(e.message).to.equal(expectedError)
+                done()
+              })
+          })
+        })
+
+        describe('.peers', () => {
+          it('throws an error', (done) => {
+            ipfs1.pubsub.peers(topic)
+              .catch((e) => {
```
diasdavid: `s/e/err`, reason below

test/pubsub-in-browser.spec.js
```diff
@@ -0,0 +1,179 @@
...
+                expect(e).to.exist
+                expect(e.message).to.equal(expectedError)
+                done()
+              })
+          })
+        })
+
+        describe('.peers', () => {
+          it('throws an error', (done) => {
```
diasdavid: could you add why it throws?

src/api/pubsub.js
```diff
@@ -0,0 +1,162 @@
...
+
+      subscribe(topic, options, handler, callback)
+    },
+    unsubscribe: (topic, handler) => {
+      if (!isNode) {
+        throw NotSupportedError()
+      }
+
+      if (ps.listenerCount(topic) === 0 || !subscriptions[topic]) {
+        throw new Error(`Not subscribed to '${topic}'`)
```
mitar: I really dislike that this pull request assumes that the only source of truth is the local state. I could subscribe to a topic through go IPFS implementation and I could then decide to unsubscribe from it from JS.

I think the JS API should just call and pass everything to IPFS daemon as it is. If IPFS API wants to throw an error if you are not subscribed to a topic and wants to unsubscribe, then IPFS daemon will pass it to JS API.

src/api/pubsub.js
```diff
@@ -0,0 +1,162 @@
...
+      ps.removeListener(topic, handler)
+
+      // Drop the request once we are actualy done
+      if (ps.listenerCount(topic) === 0) {
+        subscriptions[topic].abort()
+        subscriptions[topic] = null
+      }
+    },
+    publish: promisify((topic, data, callback) => {
+      if (!isNode) {
```
mitar: Why would this not be supported in browser? It just sends data to a HTTP endpoint?

haadcode:
Thanks @diasdavid for the CR and feedback! Will address it once I pick this up again, which I hope will happen later this week.

haadcode:
(that said, if anyone wants to get the work done asap, feel free to continue this PR before I get to it)

mitar:
BTW, for browser, we should probably use EventStreams, a w3c standard. Wikipedia is also using it now: https://blog.wikimedia.org/2017/03/20/eventstreams/

test/pubsub-in-browser.spec.js
haadcode: Added to the README

test/pubsub-in-browser.spec.js
haadcode: I'll make this even more explicit. Although, it's all mentioned in the spec file as to why "everything errors" :)

test/pubsub-in-browser.spec.js
haadcode: Will change this

test/pubsub-in-browser.spec.js
haadcode: In order to make sure we throw the error when used in the browser, I added these tests as separate ones from the interface ones. See the spec file descriptions and comments. These tests are only to test that we error correctly when js-ipfs-api is used in the browser (as per our decision to not support it for now). These shouldn't go to the interface as they would error js-ipfs, too (which is not what we want). Makes sense?

haadcode:
@diasdavid Addressed your CR feedback. Are we good to merge? :)

src/api/pubsub.js
diasdavid: Publishing should be fine, it is subscribing and being 'able to cancel a subscription' that is not.

package.json
diasdavid: This needs to point to the latest interface-ipfs-core

haadcode:
CI not happy yet, fixing.

test/pubsub-in-browser.spec.js
diasdavid: Got it

haadcode:
Rebased and squashed.

haadcode:
Discovered the need to use dirty-chai for expect. Updated and replaced the head.

src/pubsub-message-stream.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  _transform (obj, enc, callback) {
+    let msg
+    try {
+      msg = PubsubMessage.deserialize(obj, 'base64')
+    } catch (e) {
+      // Not a valid pubsub message
+      // go-ipfs returns '{}' as the very first object atm, we skip that
+      return callback()
```
diasdavid: Has this been reported? //cc @whyrusleeping @Kubuxu

src/pubsub-message-stream.js
```diff
@@ -0,0 +1,33 @@
...
+    let outputStream = inputStream.pipe(new PubsubMessageStream())
+    inputStream.on('end', () => outputStream.emit('end'))
+    callback(null, outputStream)
+  }
+
+  _transform (obj, enc, callback) {
+    let msg
+    try {
+      msg = PubsubMessage.deserialize(obj, 'base64')
+    } catch (e) {
```
diasdavid: Always use `err` for errors to make sure they are captured by the linter. Here, this error is being silenced.

src/pubsub-message-stream.js
haadcode: There was a lengthy discussion about this in https://github.com/ipfs/go-ipfs/issues/3304. (short: it's always been like that, go team was involved)

src/pubsub-message-stream.js
haadcode: Fixed and amended the latest commit.

haadcode:
The pubsub load tests are failing on Travis: https://travis-ci.org/ipfs/js-ipfs-api/jobs/214309202 (all travis platforms). Have no idea why. Will need to look into it tomorrow.

haadcode:
Update on this from IRC:
```
04:55 <+haad> daviddias: pubsub is not supposed to guarantee delivery-in-order for the pubsub messages, right?
04:56 <+haad> ie. message 2 can be received by another peer before message 1
04:56 <@dignifiedquire> haad: correct, there is no ordering to the messages
05:02 <+haad> ok great, then that's the problem with failing travis tests. the messages are all received (correctly) but sometimes they come in different order
              but we expect them to arrive in order in our interface test. daviddias dignifiedquire I'll change the interface test to only test if all messages
              are received and remove the expectation that they are in order, sounds good?
05:05 <@dignifiedquire> from my side yes I think that makes sense
05:07 <+haad> I'll wait for wild d-dash to appear to make sure we're all in agreement that's what we should do :)
```

Will catch up with @diasdavid on IRC once he's online and decide how we proceed.

diasdavid:
"And then d-hash showed up"

```
10:03 <@daviddias> haad: it is not guaranteed
10:04 <@daviddias> haad: although it is weird that in a 1-1 directly connection same host
10:04 <@daviddias> that they arrive out of order
```

I wonder if go-ipfs has a weird way of queue the http requests to the API which makes them actually switch order.

haadcode:
After some discussion on IRC and more investigation, it seems the messages in the load test get out of order on go-ipfs side. See #ipfs-dev for the full log, but here are the important parts:

```
06:39 <+haad> keks[m]: ok, makes sense. yeah so the issue we're seeing is that sometimes the messages are out of order for the pubsub load test (which
              sends/receives 10k messages as fast as it can) and our current test expects them to be in order. so we're wondering if we can drop the expectation
              that they're always in order.
06:40 < keks[m]> imho that is the right way to go
06:40 <+haad> cool
06:40 <+haad> cc daviddias ^
06:40 <+haad> Kubuxu: any thoughts on the above discussion?
06:40 <+haad> or lgierth
06:42 <@Kubuxu> haad: are there multiple sending daemons?
06:42 <+haad> Kubuxu: one is sending, one is receiving
06:46 <@Kubuxu> haad: so the seqnumber is nanotime. Can you check if seqnumber is in order.
06:46 <@Kubuxu> this would allow me to narrow it down
06:48 <+haad> Kubuxu: haven't checked if the seq number is in order but I believe they're not. we send "message #n" as the payload/data for each message which use to determine if the messages are in order.
06:49 <@Kubuxu> I am asking about seqnumber as this will tell me whether  it is floodsub or go-ipfs code.
06:50 <+haad> btw. if it helps, on Travis the load test runs at ~1000ops/s whereas locally it's around 300ops/s (ops == messages send + received / duration), so
              it's passing the message about 3x faster than locally (where it never fails)
06:52 <+haad> Kubuxu: what I'm saying is I don't know what the seq numbers look like but the payloads say they're out of order. is there a difference to payload
              vs. seq number that it can behave in different ways?
06:53 <@Kubuxu> yes, the payload is established in go-ipfs, the seqnumber after go-ipfs calls `SendMessage` after that the floodsub code looks quite stable
                regarding oredring.
06:53 <@Kubuxu> but I am not 100% sure
06:54 <+haad> ok
06:55 <+haad> I'll try to check that but it'll take a while as I need to wait for Travis to run it
```

And once Travis had ran the tests:
```
08:47 <+haad> Kubuxu: I can confirm that the seqno and payload are matching (ie. when our check for order from payload fails, the sequence numbers are also out
              of order and they are in order otherwise)
08:47 <+haad> Kubuxu: does that help you narrow it down?
08:47 <@Kubuxu> yeah, and it is quite interesting
08:48 <+haad> you can take a look at the output here if you want: https://travis-ci.org/ipfs/js-ipfs-api/jobs/214612281. the funny thing is, it's (I believe)
              always one message that arrives out of order, other 9999 arrive in order :)
08:50 <+haad> does that mean it's in floodsub (as per your comment in PM)?
09:11 <@Kubuxu> yeah, looks like it, either in floodsub code or receiving side of go-ipfs
09:11 <@Kubuxu> sends are sequential 100% if seqnumbers are also swapped
09:30 <+haad> Kubuxu: then the question is, is that working as-intended or something we ought to fix?
09:30 <@Kubuxu> not sure, you would have to ask why
09:30 <+haad> ok
```

So @whyrusleeping, what's your take on this?

whyrusleeping:
Floodsub does not guarantee the order in which you receive messages. They can come out of order, or not come at all

diasdavid:
I guess we need to take that assumption from the test then. (Still weird that in such a short network with almost 0 latency, there is disordering..)

PR #495, opened 2016-12-24
fix so that multihashes may be followed by sub-path strings

substack:
In the command-line tool you can have a hash followed by a path:

``` sh
ipfs cat QmXJ8KkgKyjRxTrEDvmZWZMNGq1dk3t97AVhF1Xeov3kB4/meta.json
```

Using ipfs-api before this patch I would get these errors when I try to do the same thing:

```
not valid multihash
```

With this patch, I get the expected output, the same as the command-line tool.


diasdavid:
Sweet, thank you @substack :)

PR #499, opened 2017-01-02
fix error: browserify -> webpack

harlantwood:


diasdavid:
Thank you @harlantwood :)

PR #512, opened 2017-01-23
Add add-from-url executable

maxlath:
```sh
add-from-url https://example.org/whatever.jpg
# => IPFS hash
```

This command isn't available from the `ipfs` CLI but can be quite convenient and easily done from this module

### DIFF
--- a/bin/add-from-url
+++ b/bin/add-from-url
@@ -1,0 +1,15 @@
+#!/usr/bin/env node
+'use strict'
+
+const ipfs = require('..')()
+const [ url ] = process.argv.slice(2)
+
+if (!url) {
+  console.error('expected a url as argument')
+  process.exit(1)
+}
+
+ipfs.util.addFromURL(url)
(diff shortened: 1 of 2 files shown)


VictorBjelkholm:
I'm not sure this is a good idea to add to js-ipfs-api which is supposedly a library for connecting to a running ipfs daemon. Might be better to write this as a independent tool instead.

diasdavid:
@maxlath this looks more like a separate tool, we want to keep js-ipfs-api as a client library. Thank you for submitting it as a PR though, let me know if you decide to make it a separate tool. Closing this one.

maxlath:
make sense, so I made a small module of it: [ipfs-add-from-url](https://github.com/maxlath/ipfs-add-from-url)

VictorBjelkholm:
@maxlath awesome! Looks great! Maybe you can submit a PR to ipfs/awesome-ipfs as well to add it there :)

PR #519, opened 2017-01-31
refactor: tests

diasdavid:
- [x] remove dedup spawning daemons code
- [x] put things where they should be
- [x] make everything use the ipfs-factory


test/ipfs-factory/daemon-spawner.js
```diff
@@ -66,30 +56,34 @@ function spawnEphemeralNode (callback) {
...
       },
       (cb) => node.startDaemon(cb)
-    ], (err) => {
-      if (err) {
-        return callback(err)
-      }
-
-      callback(null, node)
-    })
+    ], (err) => callback(err, node))
```
diasdavid: @dignifiedquire could you take a look at this function, I'm getting errors setting the config as the config is always empty, not sure what I messed up right now.

dignifiedquire: the setting looks right to me, why do you need the getConfig?

diasdavid: I added `getConfig` to test it, it yield nothing, meaning that the daemon is being created with no config at all

diasdavid:
@dignifiedquire @VictorBjelkholm could you pull this branch and try to run it and tell me what you see, I'm confused why the heck this is not working, just tested ipfsd-ctl directly and works fine, not sure why here it is going havoc

test/ipfs-factory/daemon-spawner.js
diasdavid: Solved, there were too issues, ipfsd out of that and now ipfs for some reason doesn't accept setting Boolean values with the --json flag (??)

diasdavid: The good news is, that we actually never needed `'API.HTTPHeaders.Access-Control-Allow-Credentials': true,`

test/bitswap.spec.js
```diff
@@ -0,0 +1,86 @@
...
+        done()
+      })
+    })
+  })
+
+  describe('Promise API', () => {
+    it('.wantlist', () => {
+      return ipfs.bitswap.wantlist()
+        .then((res) => {
+          expect(res).to.have.to.be.eql({
```
VictorBjelkholm: could probably be `to.be.eql` rather than that

diasdavid:
Exactly, it was super painful to expand the test suite and also isolate groups of tests, this will give us much more sanity! :) thank you for the review!

test/name.spec.js
```diff
@@ -0,0 +1,107 @@
...
+      ipfs.name.publish('Qma4hjFTnCasJ8PVp3mZbZK5g2vGDT4LByLJ7m8ciyRFZP', (err, res) => {
+        expect(err).to.not.exist
+        name = res
+        expect(name).to.exist
+        done()
+      })
+    })
+
+    it('.name.resolve', (done) => {
+      ipfs.name.resolve(name.Name, (err, res) => {
```
VictorBjelkholm: In general, it's no good if the test-cases themselves are not isolated and can't be run by themselves, in this case, `name` is being set in the `publish` test.

test/ping.spec.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-env mocha */
+'use strict'
+
+const expect = require('chai').expect
+const series = require('async/series')
+const FactoryClient = require('./ipfs-factory/client')
+
+describe.skip('.ping', () => {
```
VictorBjelkholm: We want to skip this still?

VictorBjelkholm:
Reading through this though, makes it clear to me that many of the tests are missing any assertions outside of "does not return an error" which is good for quick, sweeping tests that everything is at least not exploding. But it's not really good in terms of making sure that things actually work. Something we should keep in mind moving forward.

PR #520, opened 2017-02-01
feat(dag): basics (put, get)

daviddias:
Started implementing https://github.com/ipfs/interface-ipfs-core/pull/112 on js-ipfs-api, but there are somethings missing, from ipfs dag cli we have:

```
» ipfs dag put --help
USAGE
  ipfs dag put <object data> - Add a dag node to ipfs.

SYNOPSIS
  ipfs dag put [--format=<format> | -f] [--input-enc=<input-enc>] [--] <object data>

ARGUMENTS

  <object data> - The object to put

OPTIONS

  -f,        --format string - Format that the object will be added as. Default: cbor.
  --input-enc         string - Format that the input object will be. Default: json.

DESCRIPTION

  'ipfs dag put' accepts input from a file or stdin and parses it
  into an object of the specified format.
```

It would be great if we could use the multicodec names for the `format` flag, namely:

- dag-cbor
- dag-pb
- eth-block
- and so on

It will solve a lot of ambiguity. Can we get that @whyrusleeping ?

Another thing, it has no option to pick the hash algorithm to use.

Last note, the `--input-enc` defaults to json, but if we can use the multicodec of the type, we could send them serialized directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "eslint-plugin-react": "^6.9.0",
     "gulp": "^3.9.1",
     "hapi": "^16.1.0",
-    "interface-ipfs-core": "^0.23.5",
+    "interface-ipfs-core": "^0.23.6",
     "ipfsd-ctl": "^0.18.2",
     "pre-commit": "^1.2.2",
     "socket.io": "^1.7.2",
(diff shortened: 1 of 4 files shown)


diasdavid:
Just realized now that already had started implemented this here and found the same blockers that I have here: https://github.com/ipfs/js-ipfs-api/pull/534

Closing in favour of https://github.com/ipfs/js-ipfs-api/pull/534

PR #530, opened 2017-03-01
Implemented API for pubsub

mitar:
One more take on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,8 @@
     "readable-stream": "1.1.14",
     "stream-http": "^2.6.3",
     "streamifier": "^0.1.1",
-    "tar-stream": "^1.5.2"
+    "tar-stream": "^1.5.2",
+    "through2": "^2.0.3"
   },
   "engines": {
     "node": ">=4.0.0",
--- a/src/api/pubsub.js
+++ b/src/api/pubsub.js
@@ -4,6 +4,7 @@ const pump = require('pump')
 const through = require('through2')
 const promisify = require('promisify-es6')
 const PeerId = require('peer-id')
+const stringlistToArray = require('../stringlist-to-array')
 
 module.exports = (send) => {
   return {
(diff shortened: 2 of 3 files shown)


diasdavid:
@mitar did you find https://github.com/ipfs/js-ipfs-api/pull/493 ? @haadcode seems other people are looking for this so much that they even go the full effort of implementing themselves, how soon can we get that branch merged?

mitar:
Yes, I have seen it, but it looked much too complicated than it is necessary. Or at least other JS APIs are implemented simpler, so I do not know why pub/sub should be implemented in a more complicated way. It keeps additional client state, it throws extra exceptions.

I can comment there a bit. :-)

haadcode:
@diasdavid haven't had the time yet to pick this up again. Haven't addressed your latest CR yet, which I'll do once I get back to it again, but the steps needed to land that PR are described in https://github.com/ipfs/js-ipfs-api/pull/493#issuecomment-280321062.

@mitar not sure why it's "too complicated" for you, but it's true that the existing PR does a bit more. We needed a way to control the connection closing from the API (in order to not leave connections open) and that proved to be non-trivial. I can't remember the exact details as to why the implementation re. the subscriptions is the way it is now, I believe @dignifiedquire and @diasdavid can share thoughts on that if needed as they figured out that part. 

@mitar can you share your thoughts on what was complicated in a way that prevents you to use it? The public/user API? The implementation details? Something else? Pubsub is still very much experimental and all feedback is highly appreciated in order to make it work for everyone, as well as to find a good implementation that we can use in the long-term. The current design for the interface is specified in https://github.com/ipfs/interface-ipfs-core/blob/master/src/pubsub.js.

mitar:
Oh, no no. Do not get me wrong. I think #493 looks great. Just when I wanted something quick and working, when I looked at #493 initially, it looked like it is still in the progress and that there will be some API changes (to support browsers) and so on. It just felt overwhelming. Looking more deeper later on it looks like a good approach.

Otherwise, pub/sub with this simple API here works great for me. I am using it in this [blockchain implementation on top of IPFS](https://lucky.tnode.com/) to distribute all transactions and blocks between peers. So for me it works well. I had some issues using the encoding initially. I tried to send binary data directly, but it didn't work for some reason. I switched to send just strings and now it works.

So I do not have really anything constructive to add here. Keep up with good work and I hope all this gets at some point into a stable release.

haadcode:
Thanks for the feedback @mitar, that explains it clearly 👍 😄 

We'll get the pubsub API merged soon (finally).

diasdavid:
Going to close this one in favor of https://github.com/ipfs/js-ipfs-api/pull/493, fixing the test so that it passes green today :)

PR #531, opened 2017-03-02
Links have been ignored

mitar:
If one passes a plain object (like something returned from `toJSON`), `Links` are not copied over.


diasdavid:
oh, great catch, thanks @mitar!

PR #533, opened 2017-03-08
Fix/load fixtures

diasdavid:
ref: https://github.com/dignifiedquire/aegir/issues/109


diasdavid:
Sauce Labs is not happy, but that is not a novelty :( 

We need to make that to work nice, right now it is just annoying and not adding any value.

PR #534, opened 2017-03-09
feat: support DAG API

diasdavid:
Will need changes that come from https://github.com/ipfs/go-ipfs/issues/3771#issue-213068794


atfornes:
@diasdavid, could you please share what have to be done to complete the DAG API?

you mentioned that different formats have to be supported at https://github.com/ipfs/js-ipfs-api/pull/549, is it related with https://github.com/ipld/interface-ipld-format implementations?

if the tasks are relatively easy I can try to help with this in the JavaScript part.

Thanks

diasdavid:
Right now, we just need these tasks to be done: https://github.com/ipfs/go-ipfs/issues/3771#issuecomment-286165091, essentially, the http-api needs to support what we need in order to complete the JS. Would you like to take a look at that? :)

atfornes:
@diasdavid, I am afraid I am not familiar enough with Go language to help with the http-api. Is there something else I can do to help pushing forward the solution for this issue?

atfornes:
@diasdavid, what about implementing it in js-ipfs http API? does it makes sense? https://github.com/ipfs/js-ipfs/tree/master/src/http-api

magik6k:
So with fixes/additions in https://github.com/ipfs/go-ipfs/pull/4133 current status is that:

* `dag.put` works on both sides both for dag-cbor and dag-pb
* `dag.get` doesn't because:
  * We need to resolve top level block in path, get it, resolve remaining path on JS side
  * For that either `ipfs resolve` must be extended to return IPLD path remainder or something like `ipfs dag resolve` must be added
  * We then need to `block get` the resolved block and parse it on JS side
  * Then return subnode under whatever remainder path it was within a block.

@diasdavid Is my thinking here correct or do you think this can be solved in a simpler way?

magik6k:
So I rebased this on current master and implemented `get` as as `dag resolve`+`block get`. Everything seems to work on both (go-ipfs / js-api) sides, test failures here appear to be related more to the tests itself or deeper IPLD stuff.

src/dag/get.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+      (blk, path, cb) => {
+        if (blk.cid.codec === 'dag-cbor') {
+          dagCBOR.resolver.resolve(blk, path, cb)
+        }
+        if (blk.cid.codec === 'dag-pb') {
+          dagPB.resolver.resolve(blk, path, cb)
+        }
+      }
+    ], callback)
```
diasdavid: @magik6k this is an interesting approach. It misses any other format that is not `dag-cbor` and or `dag-pb`. I also believe that it won't pass every test currently present in https://github.com/ipfs/interface-ipfs-core/blob/master/src/dag.js, which is a requisite to get things merged in js-ipfs-api.

I had understood that you were working on implementing the dag api on go-ipfs, is there anything blocking you there from finishing the implementation?

src/dag/put.js
```diff
@@ -0,0 +1,68 @@
...
+    function prepare () {
+      inputEnc = 'raw'
+
+      if (format === 'dag-cbor') {
+        dagCBOR.util.serialize(dagNode, finalize)
+      }
+      if (format === 'dag-pb') {
+        dagPB.util.serialize(dagNode, finalize)
+      }
+    }
```
diasdavid: Same drawbacks as my comment above https://github.com/ipfs/js-ipfs-api/pull/534/files#r133178508

src/dag/get.js
magik6k: go-ipfs side is done. Now that i think of it, is there any reason the code for put which was here earlier didn't just pass the data as json to go-ipfs? Passing json directly instead of serializing it here would allow to support every encoding go side supports without additional work here.

The same is the case for `get` with the difference that using go-ipfs dag get doesn't return `remainderPath`(I don't think there is any case where it wouldn't be empty). Does the value passed to callback  in `get` implement any API or is it much like just like parsing json from `ipfs dag get`?

diasdavid: > Passing json directly instead of serializing it here would allow to support every encoding go side supports without additional work here.

Using json for all the serialization has a big drawback described here: https://github.com/ipfs/interface-ipfs-core/issues/81#issuecomment-277271941

tl;dr; JSON doesn't support Buffers, which makes it really hard to distinguish a Buffer from a String, you can see that we do a cast to Buffer every single time in `object get` just because we know that Data in the protobuf is always binary. This is the tl;dr; but please do read the whole thread :)

> The same is the case for get with the difference that using go-ipfs dag get doesn't return remainderPath(I don't think there is any case where it wouldn't be empty)

`remainderPath` should return any path that the node wasn't able to resolve either because it couldn't find more nodes or because it didn't have the right parser for it.

diasdavid:
@Stebalien what is the state of IPLD in go-ipfs? Any possible luck with having an API that works for _every format_ soon so that we can have this feature enabled in js-ipfs-api??

magik6k:
We want to implement corporate API to not lose type info. Code in this PR works but doesn't support binary buffers iirc

Stebalien:
@diasdavid Which APIs were you looking for. We already have a general purpose decoding framework that decodes any node into an IPLD "Node" object (that has methods for resolving through the node).

Missing:

1. UnixFS using arbitrary IPLD codecs.
2. Encoding arbitrary objects (currently, the interface is read-only).
3. Nice (de)serialization to/from types using reflection.

diasdavid:
@Stebalien I meant an HTTP API endpoint that would enable this DAG API to work fully (retrieve any kind of node and type) without getting it mangled when it gets into the HTTP wire

Stebalien:
@diasdavid Ah. I see. Well, if you *just* want to get a node by CID, yes. You can use the `block` interface:

* `ipfs block get XYZ` to get a node.
* `ipfs block put --format=my-codec --mhtype=sha256 --mhlen=256 <<<"some data"`

The `dag` interface is supposed to do this but it kind of sucks at the moment.

diasdavid:
@Stebalien That is a very clever idea -- https://github.com/ipfs/js-ipfs-api/pull/534#issuecomment-352067738 -- I can indeed import the ipld-resolver -- http://github.com/ipld/js-ipld-resolver -- here and pass the ipfs.block as the block service to it and it will work like a charm! YEAAS! TINY REUSABLE MODULES :D

magik6k:
It's actually what's done here now: https://github.com/ipfs/js-ipfs-api/pull/534/files#diff-3bb5dae69a8220612dea17dbebe6bc22R49

diasdavid:
@vmx as you go through the DAG API, check this PR as well. You can implement the same `dag API` we have in js-ipfs land by putting the `ipld-resolver` on top of the `ipfs block` API. Idea suggested at https://github.com/ipfs/js-ipfs-api/pull/534#issuecomment-352067738

koalalorenzo:
Do we have any update on this? ETA? :) 
Do you need any help for this PR?

vmx:
@koalalorenzo I don't have an ETA for this. I currently working on getting the CLI to support the `dag` command (https://github.com/ipfs/js-ipfs/issues/1212).

In regards to help: sure, any help is welcome. After a quick look at this PR, I think it shouldn't use the IPLD formats directly, but instead use the API from https://github.com/ipld/js-ipld. This way you needed care format specific things yourself, but have support for all.

@diasdavid Do you agree that using `js-ipld` would make sense?

pcardune:
I got here looking for a way to put an IPLD serialized object into an IPFS node while that IPFS node was running as a daemon.

I figured out how to use ipld-js to store the data while the IPFS node is not running: https://gist.github.com/pcardune/411c93eb4dd7d15a642b3750a4bd8cfa. But then you have to start the IPFS node to make that data available publicly.

I'm guessing that this PR would solve my problem, but since it's unclear if and when this PR is going to get merged, I'm wondering if there is another way to accomplish this?

diasdavid:
> @diasdavid Do you agree that using js-ipld would make sense?

I do :) That was my realization on https://github.com/ipfs/js-ipfs-api/pull/534#issuecomment-352068424, that we can use js-ipld and use the ipfsAPI.block as the block service that js-ipld uses. This way we don't have any serialization problems.

vmx:
@koalalorenzo @pcardune @jaycenhorton I'm working on this now. I thought it's a piece of cake, but it turned out that it's harder than I thought (too many bugs and outdated code on the way). Though I'll have a look again next week.

diasdavid:
@vmx AFAIU you won't get the nice dag API -- as in https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DAG.md -- from go-ipfs http-api soon due to how things are wired inside go-ipfs (@Stebalien can confirm). You will need to implement the dag API by bringing js-ipld and mounting it on top of the Block API, making the resolve happen clientside.

vmx:
@diasdavid https://github.com/ipfs/go-ipfs/issues/4728#issuecomment-367532434 sounds promising.

At least it should be possible to do it "manually" in JS land. But currently even that is not really possible. If you request `<cid>/Data` you will get an error and not the CID with `Data` as remainder path back. Just to be clear that totally makes sense, the JS implementations do the same for paths they don't know. So let's just keep the fingers crossed that there will be a js-ipld compatible representation of Protocol Buffer blocks in the future. Until then it will just not be supported by js-ipfs-api.

vmx:
Many people have worked on this. I've rebased the commits and added my own ones. This is ready for review, so @diasdavid and @magik6k please review it.

src/dag/put.js
```diff
@@ -0,0 +1,69 @@
...
+
+    function prepare () {
+      inputEnc = 'raw'
+
+      if (format === 'dag-cbor') {
+        dagCBOR.util.serialize(dagNode, finalize)
+      }
+      if (format === 'dag-pb') {
+        dagPB.util.serialize(dagNode, finalize)
+      }
```
diasdavid: What about all the other types?

magik6k: Fair enough.

I'd import `git` here too as it's the one other (well) supported codec on go-ipfs side (and would be really handy to have for https://github.com/ipfs-shipyard/IGiS which can't use `window.ipfs` from api now)

vmx: Is it OK if I I open an issue for it and do it in a separate PR? That' also something newcomers can work on.

vmx: @diasdavid Sounds like a test for that is missing in interface-ipfs-core then.

It will work once js-ipld is used.

vmx: Summary of this thread: everything mentioned here is solved if js-ipld is used. Hence I propose merging this change as is currently is and open a new issue (I'll do that) to use js-ipld instead.

diasdavid: Sounds good my IPLD captain! 🚢

PR #543, opened 2017-04-04
Add FOSSA badge/report

xizhao:


diasdavid:
Hi @xizhao! Looks like http://fossa.io/ is a super useful service. I see that you opened several PR's to IPFS modules but with a little context, could you explain what lead you to do that? Thank you!

PR #548, opened 2017-04-26
Add key api from go-ipfs

MichaelMure:
Hi,

I needed that for my project so maybe it's useful for you as well :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,11 @@ This means:
 - [`ipfs.ping()`]()
 - [`ipfs.log()`]()
 
+#### [key]()
+
+- [`ipfs.key.gen(name, [options, callback])`]()
+- [`ipfs.key.list([options, callback])`]()
+
 ##### [name]()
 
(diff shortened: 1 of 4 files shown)


MichaelMure:
Fixed the dangling comma. I don't think I'm to blame for the remaining test fail.

MichaelMure:
@diasdavid Is it planned to support the key api here ? I'd like to avoid using a fork.

If yes I can make a better PR.

dignifiedquire:
Thank you @MichaelMure. The goal is to support the whole api that go exposes in this library. It would be great if you could a test for the new method.

MichaelMure:
@dignifiedquire what about interface-ipfs-core ? It needs to be properly spec-ed to go there and the api in go-ipfs and here will be adjusted later ?

diasdavid:
I don't think we will need this to be part of the main interface library, it will be eventually exposed directly through .crypto. Meanwhile this PR looks good, merging it in :)

PR #549, opened 2017-05-04
Implement dag.get

magik6k:
Implemented `dag.get` for #534. Arguments should behave exactly as in js-ipfs.

### DIFF
--- a/src/api/dag.js
+++ b/src/api/dag.js
@@ -71,7 +71,40 @@ module.exports = (send) => {
       }
     }),
     get: promisify((cid, path, options, callback) => {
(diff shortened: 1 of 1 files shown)


diasdavid:
Hi @magik6k thank you for PR'ing this. It is a bit more complex than it seems as the DAG API needs to support multiple formats and not just protobufs.

I started the work here: https://github.com/ipfs/js-ipfs-api/pull/534 but it will require some updates on go-ipfs as well. Would you like to help land both changes in js and go land?

Let's keep chatting over #534 , closing this one.

PR #551, opened 2017-05-06
request-api allow the http method to be specified

kumavis:
via `options.method = 'GET'`

### DIFF
--- a/src/api/block.js
+++ b/src/api/block.js
@@ -47,6 +47,7 @@ module.exports = (send) => {
       }
 
       const request = {
+        method: 'GET',
         path: 'block/get',
         args: args,
         qs: opts
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -96,7 +96,7 @@ function requestAPI (config, options, callback) {
   // this option is only used internally, not passed to daemon
   delete options.qs.followSymlinks
 
-  const method = 'POST'
+  const method = options.method || 'POST'
   const headers = {}
 
   if (isNode) {


kumavis:
not sure if test failure is related...
```
  1 failing
  1) .files callback API .add .createAddStream stream of valid files and dirs:
     Timeout of 80000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

diasdavid:
What was the issue you were having? CORS related?

kumavis:
@diasdavid Parity only supports GET for `block/get` per the apiary API spec
https://github.com/ipfs/http-api-spec/blob/master/apiary.apib#L807

diasdavid:
ah, I see. Well, would it be fine for parity to support just method * for now on that one? I feel that it is easier to make it mimic what go-ipfs does than having to remember everyone to set the method for this specific call.

kumavis:
can we either:
1) change the spec or
2) adhere to the spec

kumavis:
> I feel that it is easier to make it mimic what go-ipfs does

go-ipfs and js-ipfs both handle this method when made as a GET request

kumavis:
https://github.com/ipfs/http-api-spec/issues/124

kumavis:
since parity only implements the single route (block/get) I'll just hit that manually and drop `js-ipfs-api` in the datastore adapter

diasdavid:
on https://github.com/ipfs/js-ipfs-api/pull/551#issuecomment-300046811

We've a lot to do in that land - https://github.com/ipfs/http-api-spec/issues/116

PR #555, opened 2017-05-13
Use CID for api.get/api.cat

magik6k:
go-ipfs supports CIDv1 for cat/get commands, so I guess it should also be supported here.

https://github.com/ipfs/interface-ipfs-core/tree/master/API/files#cat should be updated if this gets merged.

### DIFF
--- a/src/api/cat.js
+++ b/src/api/cat.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const promisify = require('promisify-es6')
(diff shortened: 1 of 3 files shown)


diasdavid:
Awesome @magik6k, thank you for adding this in :) 

Mind adding some tests to https://github.com/ipfs/interface-ipfs-core/blob/master/src/files.js that contemplate this change?

PR #556, opened 2017-05-16
Fix the `ping` results

ngotchac:
I've see cases where the `ping` response is in fact on 2 lines, the second one containing the Success (`false`), the Time and the Text. This happens for example if you try to ping your self.

### DIFF
--- a/src/api/ping.js
+++ b/src/api/ping.js
@@ -21,10 +21,14 @@ module.exports = (send) => {
 
         // go-ipfs http api currently returns 3 lines for a ping.
         // they're a little messed, so take the correct values from each lines.
+        const Success = res[1].Success
+        const Time = res[1].Time
+        const Text = res.length > 2 ? res[2].Text : res[1].Text
+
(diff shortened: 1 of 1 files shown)


diasdavid:
@Kubuxu any thoughts on why this happens?

diasdavid:
@ngotchac mind rebasing master onto this PR?

@Kubuxu thoughts?

Kubuxu:
It has happened because `ping` endpoint is ndjson streaming endpoing. The order of responses that can be received there are not part of the API.

What the code should do is parse the response as NDJSON (line by line) and look for `.Time != 0`.

diasdavid:
@ngotchac wanna follow up on this PR?

Kubuxu (changes requested):
This isn't the right way to do it as API does not specify order of messages. See https://github.com/ipfs/js-ipfs-api/pull/556#issuecomment-313846598

(cleaning up my CR queue)

PR #560, opened 2017-06-04
readme: added missing API links

maxlath:
even when linking to a 404 octocat, still better than the current empty links resolving to https://github.com/ipfs/js-ipfs-api/blob/master


maxlath:
you're welcome :)

Issue #544, opened 2017-04-11
Make js-ipfs-api modular


diasdavid:
Users would be able to achieve really small js-ipfs-api bundle sizes if it was structured like async or lodash, where a user can pick only a subset of the module to be bundle into their app.

Example with async:

```
const parallel = require('async/parallel')
```

We want to achieve this for js-ipfs-api, so that users don't have to create their own sharded forks (ie. https://github.com/SilentCicero/ipfs-mini)

Ideally, this should work so that both `const files = require('js-ipfs-api/files')` and `const add = require('js-ipfs-api/files/add')` should work.

diasdavid:
Done with #561 \o/

PR #561, opened 2017-06-05
modularize js-ipfs-api

nunofmn:
Related to #544.

The structure will be similar to the `async` module structure, but to do this with minimal changes to the current structure, is necessary a "module init" step where the config is passed to the `request-api` inner module.

The usage will be similar to the following:
```js
const bitswap = require('ipfs-api/src/bitswap')('/ip4/127.0.0.1/tcp/5001')

bitswap.unwant(key, (err) => {
  // ...
}
```

- [x] bitswap
- [x] block
- [x] bootstrap
- [x] config
- [ ] ~~dag~~
- [x] dht
- [x] files
- [x] id
- [x] version
- [x] ping
- [x] log
- [x] key
- [x] name
- [x] object
- [x] pin
- [x] pubsub
- [x] refs
- [x] repo
- [x] swarm
- [x] diag
- [x] commands
- [x] ls
- [x] mount
- [x] update
- [x] util
- [x] add
- [x] get
- [x] cat


src/api/bitswap.js
```diff
@@ -1,29 +1,9 @@
 'use strict'
 
-const promisify = require('promisify-es6')
-
-module.exports = (send) => {
+module.exports = (arg) => {
```
dignifiedquire: might be nicer to rename to `bitswap/index.js`

src/api/bitswap/stat.js
```diff
@@ -0,0 +1,22 @@
...
+  let send
+
+  if (typeof arg === 'function') {
+    send = arg
+  } else if (typeof arg === 'object') {
+    const requestAPI = require('../../request-api')
+    send = requestAPI(arg)
+  } else {
+    throw new Error('Argument must be a send function or a config object.')
+  }
```
dignifiedquire: seems like the above boilerplate should be extracted into its own file and be reused, rather than copied into each file

test/bitswap.spec.js
```diff
@@ -50,13 +68,39 @@ describe('.bitswap', () => {
       })
     })
 
+    it('.stat (modular)', (done) => {
```
dignifiedquire: It would be nice to find a way to not duplicate the tests, but rather reuse them and just pass in the two different versions

src/api/bitswap.js
diasdavid: 👍

src/api/bitswap/stat.js
diasdavid: 👍

test/bitswap.spec.js
diasdavid: 👍

diasdavid:
@nunofmn I like this approach, it builds nicely. Follow @dignifiedquire suggesting, making it modular should be about importing only what you need but also avoiding code duplication. 

Thanks! 🌟

diasdavid:
Hi @nunofmn, any hopes of getting this PR finished? This will make the life so much easier for everyone and I would love to get this merged before other PR's are made :)

src/api/bitswap/stat.js
nunofmn: Yes, I was thinking in extracting the config processing part in the `IpfsAPI` function, and use it in all modules.

test/bitswap.spec.js
nunofmn: Do you have any idea of how this could be done?

src/api/bitswap.js
nunofmn: Initially I had some problems when I structured it like you are saying. But I don´t remember what was the problem. I will try again 😃

README.md
```diff
@@ -224,7 +224,12 @@ This means:
 - [`ipfs.id([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic#id)
 - [`ipfs.version([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic#version)
 - [`ipfs.ping()`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic#ping)
-- [`ipfs.log()`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic#log)
+
+#### [log](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic)
```
diasdavid: Let's remove these headings.

See how it is done in line 212

README.md
```diff
@@ -224,7 +224,12 @@ This means:
...
+
+- [`ipfs.log.ls([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic)
+- [`ipfs.log.tail([callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic)
+- [`ipfs.log.level(subsystem, level, [options, callback])`](https://github.com/ipfs/interface-ipfs-core/tree/master/API/generic)
 
 #### [key](https://github.com/ipfs/interface-ipfs-core/tree/master/API/key)
```
diasdavid: Let's remove these headings.

See how it is done in line 212

src/api/bitswap/index.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+module.exports = (arg) => {
+  return {
+    wantlist: require('./wantlist')(arg),
+    stat: require('./stat')(arg),
+    unwant: require('./unwant')(arg)
+  }
+}
```
diasdavid: This needs to be bubbled up.

`src/bitswap/index.js` and not `src/api/bitswap/index.js`

src/api/bitswap/stat.js
```diff
@@ -0,0 +1,14 @@
+'use strict'
+
+const promisify = require('promisify-es6')
+const moduleConfig = require('../../module-config')
+
+module.exports = (arg) => {
+  const send = moduleConfig(arg)
```
diasdavid: Do this step at the `index.js` level. Otherwise, you are creating a 'send' instance for every API call.

And also, at the index.js, you should check if the `arg` is already a `send` instance, for when the user uses ipfs-api directly and not a subapi.

src/api/block/get.js
```diff
@@ -0,0 +1,58 @@
...
+        cid = new CID(args)
+        args = cid.toBaseEncodedString()
+      } else if (typeof args === 'string') {
+        cid = new CID(args)
+      } else {
+        return callback(new Error('invalid argument'))
+      }
+    } catch (err) {
+      return callback(err)
+    }
```
diasdavid: Could you refactor this into a function?

diasdavid:
- [ ] As you move all the `/src/api/*` to `/src/*`, move all the other files in (https://github.com/ipfs/js-ipfs-api/tree/master/src) to `/src/utils/*`
- [ ] Please create an example showing how to require just a subAPI

src/api/bitswap/stat.js
nunofmn: The main problem of doing this step at `index.js` level is that will lose the capability of require individual sub-modules (for example, `require('ipfs-api/bitswap/unwant')(config)`)

The `moduleConfig` already checks if the `arg` is a `send` instance, or if is necessary to init one.

src/module-config.js
```diff
@@ -0,0 +1,12 @@
...
+module.exports = (arg) => {
+  if (typeof arg === 'function') {
+    return arg
+  } else if (typeof arg === 'object') {
+    const requestAPI = require('./request-api')
+    return requestAPI(arg)
+  } else {
+    throw new Error('Argument must be a send function or a config object.')
+  }
+}
```
diasdavid: This doesn't seem to be correct. What if a user does `const node = ipfsApi('/ip4/127.0.0.1/tcp/5001')` ?

diasdavid:
@nunofmn almost there :)

- [ ] resolve merge conflict
- [ ] git ignore the generated files in the example image
- [ ] update the readme

examples/sub-module/package.json
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "author": "Nuno Nogueira",
+  "license": "MIT",
+  "devDependencies": {
+    "babel-core": "^6.25.0",
+    "babel-loader": "^7.1.0",
+    "babel-preset-babili": "^0.1.4",
+    "babel-preset-env": "^1.5.2",
+    "babili": "^0.1.4",
+    "babili-webpack-plugin": "^0.1.1",
```
diasdavid: Why so much babel? Shouldn't this module require any babel at all?

src/bitswap/stat.js
```diff
@@ -5,7 +5,7 @@ const promisify = require('promisify-es6')
 module.exports = (send) => {
   return promisify((callback) => {
     send({
-      path: 'commands'
+      path: 'bitswap/stat'
     }, callback)
   })
 }
```
diasdavid: Interesting that git thinks this was a rename from commands.js

diasdavid:
Values in KBs
image

Good results! Once we get https://github.com/ipfs/go-ipfs/issues/4008 it will even be better! :)

examples/sub-module/bundles-size-KBs.csv
```diff
@@ -0,0 +1,28 @@
...
+name - 684.63 - 232.06
+object - 923.66 - 340.57
+pin - 685.51 - 232.39
+ping - 684.59 - 231.90
+pubsub - 740.40 - 249.54
+refs - 684.39 - 231.94
+repo - 684.56 - 232.02
+swarm - 1324.18 - 527.03
+update - 684.45 - 231.96
+version - 684.21 - 231.88
```
diasdavid: Please use commas in the comma separated values files :) 

Also, add the units to the top.

examples/sub-module/test-modules-size.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
```
VictorBjelkholm: Should have at least `set -e` so it exists the entire script if there is a error

diasdavid:
This has been an impressive refactor! Thank you @nunofmn for pulling this off.

There is one failing test, not related with the changes itself. Needs to be investigated separately. Merging!

Issue #562, opened 2017-06-08
add --nocopy not supported


MichaelMure:
Hi,

I'm trying to use the --nocopy option of the filestore through js-ipfs-api but files added this way are stored in the regular blockstore. After a lot of digging I think it's due to how the file is streamed in the http request:

Capture of the multipart/form-data of a request made with the CLI:
```
 --35471fb491711d32f7a9885e488899d24c6483bd528959cb002c6a6e32b5 
 Abspath: /home/michael/go/src/gx/ipfs/QmaeRR9SpXumU5tYLRkq6x6pfMe8qKzxn4ujBpsTJ2zQG7/go-os-rename/README.md 
 Content-Disposition: file; filename="README.md" 
 Content-Type: application/octet-stream 
  
 # go-os-rename
 ...
```

Capture of the multipart/form-data of a request made with js-ipfs-api:
```
--6ezi3e1udjdvx3sly5t01i3sor 
Content-Disposition: file; filename="README.md" 
Content-Type: application/octet-stream 
 
# go-os-rename
```

The absolute path is missing, so my guess is that the filestore can't file the file on disk and the data is then stored in the regular blockstore. Bad things is that happen silently ...

CC @kevina, is that correct ?

Kubuxu:
cc @VictorBjelkholm when you got time.

diasdavid:
@MichaelMure we don't yet support the filestore API from go-ipfs. Wanna help us by taking a stap at it?

```
» ipfs filestore
USAGE
  ipfs filestore - Interact with filestore objects.

  ipfs filestore

SUBCOMMANDS
  ipfs filestore dups              - List blocks that are both in the filestore and standard block storage.
  ipfs filestore ls [<obj>]...     - List objects in filestore.
  ipfs filestore verify [<obj>]... - Verify objects in filestore.

Use 'ipfs filestore --help' for more information about this command.
```

obo20:
@MichaelMure can I ask how you were capturing these requests? I'm attempting to work on a fix for this, but have been running into some strange issues where the go-ipfs daemon seems to accept the nocopy flag / abspath header, and adds the content in a different manner than normal, but this different manner isn't adding it to the filestore.

Being able to capture what's getting sent would help with debugging.

MichaelMure:
@obo20 IIRC I used wireshark

obo20:
[...]

Issue #522, opened 2017-02-10
reduce size of `ipfs.add` bundle and number of RPC requests.


slothbag:
It appears the first request is to add the content, which we get a hash value returned.  But then it makes two more requests to get the data and links back.

Is there are a way to disable the last 2 requests?  I only need the hash.

Cheers

diasdavid:
@slothbag I believe it only issues 2 requests, correct?

We could skip the second one if this https://github.com/ipfs/go-ipfs/issues/4008 comes to happen.

Now that https://github.com/ipfs/js-ipfs-api/pull/561 is happening, we will see a ton of bundle size reduction for `ipfs-api/add` if we can avoid that extra RTT

diasdavid:
Now that https://github.com/ipfs/go-ipfs/pull/4082 has been merged and released, we can finish this up.

diasdavid:
Now that the Dragon is gone (once released) https://github.com/ipfs/js-ipfsd-ctl/pull/165#issuecomment-327443185 I'll be able to PR for this issue in this issue :)

diasdavid:
Seems that I was wrong, the feature required in go-ipfs is only going to be released in 0.4.11. Nevertheless, I went ahead and developed the PR for it once it gets released -> https://github.com/ipfs/js-ipfs-api/pull/595

Closing this one, let's track on the PR itself https://github.com/ipfs/js-ipfs-api/pull/595

diasdavid:
Released with `v14.3.7` 🚀

PR #564, opened 2017-06-13
make ipfs add --nocopy works

MichaelMure:
Related to https://github.com/ipfs/js-ipfs-api/issues/562

Provide the Abspath header in the stream for files so the filestore can find the actual file on disk.
It works but I still get errors like 'Error: expected protobuf dag node' for some files but not for some others.

@kevina, any ideas ?

### DIFF
--- a/src/get-files-stream.js
+++ b/src/get-files-stream.js
@@ -16,6 +16,7 @@ function headers (file) {
   } else if (file.symlink) {
     header['Content-Type'] = 'application/symlink'
   } else {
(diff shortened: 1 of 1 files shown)


diasdavid:
This will work with the update for https://github.com/ipfs/js-ipfs-api/issues/522

diasdavid:
@MichaelMure this PR is now unblocked because options are correctly passed. Wanna look at it again?

Issue #569, opened 2017-06-20
pubsub.publish mangles binary buffers


vith:
I was trying to send a CBOR serialized object through the pubsub API and found that the bytes were not being round tripped intact. This is because the Buffer passes through a querystring on its way to the go-ipfs instance, and the qs library being used does not handle a binary buffer safely in this function: https://github.com/ljharb/qs/blob/v6.4.0/lib/utils.js#L88

I have a patch for js-ipfs-api that uses qs.stringify's encoder option to override the encoding of Buffer objects, escaping every byte that isn't an rfc3986 unreserved character without further mutations. I'll submit it as a PR shortly in case it's the desired solution. I'm not familiar enough with this library/spec to say, but maybe the buffer should just be sent in the request body instead of in a query string parameter.

vith:
Fixed in https://github.com/ipfs/js-ipfs-api/releases/tag/v14.0.4

PR #570, opened 2017-06-21
fix: don't let qs mangle binary buffers (#569)

vith:
I will be making a PR against interface-ipfs-core to add a test for the scenario that failed without this

### DIFF
--- a/src/request-api.js
+++ b/src/request-api.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const Qs = require('qs')
+const qsDefaultEncoder = require('qs/lib/utils').encode
 const isNode = require('detect-node')
 const ndjson = require('ndjson')
 const pump = require('pump')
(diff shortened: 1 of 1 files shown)


vith:
CI appears to fail because of an npm@5 bug. `webcrypto-shim` isn't being installed during `npm install`. I had the same problem locally.

You can either use npm@4 or delete `package-lock.json` before running `npm install`.

I believe it's the same bug as reported here https://github.com/npm/npm/issues/16902

diasdavid:
Just removed package-lock.json from master. It is still not ready for prime time.

atfornes:
the  `webcrypto-shim` issue is due to the unmet `chai` dependency that [this comment](https://github.com/ipfs/js-ipfs-api/commit/01e48cbe46af4e07ad34a0c6cebcc4a135ba1dfd#commitcomment-22511615) explains.

The output of th command `npm list webcrypto-shim` shows that the reason is because version discrepancies for chai:
```
ipfs-api@14.0.3 /home/antonio/Desarrollo/Projects/js-ipfs-api
├── UNMET PEER DEPENDENCY chai@4.0.2
└─┬ peer-id@0.8.7
  └─┬ libp2p-crypto@0.8.8
    └── webcrypto-shim@0.1.1  (git://github.com/dignifiedquire/webcrypto-shim.git#effe03d3401ccbfb4837191bc393bb347e0c0f0a)

npm ERR! peer dep missing: chai@<1.10.0 || >1.10.0 <4, required by dirty-chai@1.2.2

```

atfornes:
the `webcrypto-shim` issue should be fixed by commit https://github.com/ipfs/js-ipfs-api/commit/195b1e5fdb1e4566cdcc824ed43c40b8d7d28b64, which updates `dirty-chai`.

vith:
@atfornes I'm not sure they're related. I can still get the same problem just by doing `npm install && rm node_modules -rf && npm install` even after that commit. The first time it works but creates a `package-lock.json`; the second time it fails to install `webcrypto-shim`.

Then karma tests fail as expected with `Error: Uncaught Error: Cannot find module "webcrypto-shim" (webpack:///~/libp2p-crypto/src/crypto/webcrypto-browser.js?431a**************:8:0 <- test/repo.spec.js:6639)`

atfornes:
@vith, `chai` and `dirty-chai` are development dependencies. Could you try to `npm install --dev` and let us know if that fixes the issue?

vith:
npm installs dev dependencies by default (unless you pass --prod or similar). also:

```
npm WARN install Usage of the `--dev` option is deprecated. Use `--only=dev` instead.
```
Here is an isolated way to reproduce what I'm describing:

`docker run node:8.1.2 bash -c "git clone https://github.com/ipfs/js-ipfs-api && cd js-ipfs-api && npm install --dev && echo 'first check' && stat ./node_modules/webcrypto-shim && rm node_modules -rf && npm install --dev && echo 'second check' && stat ./node_modules/webcrypto-shim"`

In the middle you can see
```
first check
  File: './node_modules/webcrypto-shim'
  Size: 131       	Blocks: 0          IO Block: 4096   directory
Device: fe01h/65025d	Inode: 8457230     Links: 3
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2017-06-21 19:47:22.517319430 +0000
Modify: 2017-06-21 19:47:22.575317097 +0000
Change: 2017-06-21 19:47:22.575317097 +0000
 Birth: -
```

After ~2m22s it will end with

```
second check
stat: cannot stat './node_modules/webcrypto-shim': No such file or directory
```

PR #578, opened 2017-07-14
object patch add-link: don't force to have a Buffer for 'ref'

MichaelMure:
This allows to send a multi-hash as a string as well.

### DIFF
--- a/src/object/addLink.js
+++ b/src/object/addLink.js
@@ -27,7 +27,7 @@ module.exports = (send) => {
       args: [
         multihash,
         dLink.name,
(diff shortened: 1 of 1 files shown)


MichaelMure:
Also while I was looking at the code, I saw that in a lot of place there is a call to `cleanMultihash(multihash, opts)` with a second argument `opts` that doesn't exist in the `cleanMultihash` function. Am I missing something or should it be cleaned ?

MichaelMure:
@diasdavid what about my question just above ? https://github.com/ipfs/js-ipfs-api/pull/578#issuecomment-315325744

Issue #587, opened 2017-08-31
Support "+" character in file paths


petethomas:
Adding a file that includes ```+``` characters in ```path``` can result in a stored path that uses spaces instead.  For example:

```javascript
const ipfsAPI = require('ipfs-api')
const ipfs = ipfsAPI()
ipfs.files.add(
  [{
    path: 'ti,c64x+megamod-pic.txt',
    content: Buffer.from('testing', 'utf-8')
  }],
  (err, res) => {
    if (err) done(err)
    console.log(res)
  }
)
```
Current result:

    [ { path: 'ti,c64x megamod-pic.txt',
        hash: 'QmZJsQUWUes6vz6DzDT3z6383k1Y4hFYUKun9UkPN1PYSX',
        size: 15 } ]

Expected result:

    [ { path: 'ti,c64x+megamod-pic.txt',
        hash: 'QmZJsQUWUes6vz6DzDT3z6383k1Y4hFYUKun9UkPN1PYSX',
        size: 15 } ]

Expected result preserves ```+```s similar to the command-line utility.  For example:

    λ ipfs add -w ~/projects/linux/Documentation/devicetree/bindings/interrupt-controller/ti,c64x+megamod-pic.txt
    added QmQ2eHGDMqC7dBD44EJFQRqkypPKQ9z9coBi1VX7vBZhPZ ti,c64x+megamod-pic.txt
    added QmXdMfwwg4xrVVHo8YGwjbRChnM22UCs21r4zWFgmjeZKU
    λ ipfs get QmXdMfwwg4xrVVHo8YGwjbRChnM22UCs21r4zWFgmjeZKU
    Saving file(s) to QmXdMfwwg4xrVVHo8YGwjbRChnM22UCs21r4zWFgmjeZKU
     4.03 KB / 4.03 KB [============================================================================================================================================================================] 100.00% 0s
    λ cd QmXdMfwwg4xrVVHo8YGwjbRChnM22UCs21r4zWFgmjeZKU/
    λ ls
    ti,c64x+megamod-pic.txt
    λ

I  think this package can achieve parity by URI-encoding the path before getting the file stream.

-pete

PR #588, opened 2017-08-31
Support "+" character in file paths

petethomas:
Addresses #587 Preserve "+" in file paths.


diasdavid:
Thank you @petethomas :)

PR #589, opened 2017-09-01
Allow --cid-version option to be passed to ipfs files add

alanshaw:
Simply passes options to add command into the query string.

I've had to change `getDagNode` to be able to deal with a CID **or** a multihash as it's hash argument. This is because when specifying `--cid-version=1` we get back a CID in the response from `files/add`, and then the API uses `object/get` and `object/data` to fetch the content that was just added.


diasdavid:
image

Going to release with the new go-ipfs 0.4.10 support.

PR #594, opened 2017-09-06
Feat/add opts

diasdavid:


diasdavid:
Travis is just not in the mood to start today..

PR #595, opened 2017-09-06
feat: avoid doing multiple RPC requests for files.add, fixes #522

### LINKED ISSUE #522: reduce size of `ipfs.add` bundle and number of RPC requests.

diasdavid:
This has been tested with **go-ipfs master => go-ipfs 0.4.11** which hasn't been released yet. Once that gets released, we can merge this PR solving #522


examples/sub-module/bundles-size-KBs.csv
```diff
@@ -1,28 +1,28 @@
...
-name, 684.63, 232.06
-object, 923.66, 340.57
-pin, 685.51, 232.39
-ping, 684.59, 231.90
-pubsub, 740.40, 249.54
-refs, 684.39, 231.94
-repo, 684.56, 232.02
-swarm, 1324.18, 527.03
-update, 684.45, 231.96
-version, 684.21, 231.88
```
diasdavid: Now even with more space savings!

diasdavid:
go-ipfs v0.4.11 is out there. Need to do the release dance and then it should be all good :)

PR #596, opened 2017-09-06
Adds wiring for a progress bar

bmordan:
Passes the 'progress bar' option for transfers over `ipfs-api` (http). Progress is shown for your payload being streamed up to the server.

There is a pause.

Then you get your response (list of files).

progress-ipfs-api-folder

This example shows a folder of movie clips being added via `ipfs-api` with the daemon running.

### DIFF
--- a/src/utils/request-api.js
+++ b/src/utils/request-api.js
@@ -10,6 +10,7 @@ const getFilesStream = require('./get-files-stream')
 const streamToValue = require('./stream-to-value')
 const streamToJsonValue = require('./stream-to-json-value')
 const request = require('./request')
+const Transform = require('readable-stream').Transform
 
 // -- Internal
 
@@ -160,7 +161,17 @@ function requestAPI (config, options, callback) {
   })
 
   if (options.files) {
-    stream.pipe(req)
+    if (options.progress && typeof options.progress === 'function') {
+      const progressStream = new Transform({
+        transform: (chunk, encoding, cb) => {
+          options.progress(chunk.byteLength)
+          cb(null, chunk)
+        }
+      })
+      stream.pipe(progressStream).pipe(req)
+    } else {
+      stream.pipe(req)
+    }
   } else {
     req.end()
   }
--- a/test/files.spec.js
+++ b/test/files.spec.js
@@ -73,6 +73,15 @@ describe('.files (the MFS API part)', function () {
       })
     })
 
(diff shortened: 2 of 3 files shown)


test/files.spec.js
```diff
@@ -73,6 +73,15 @@ describe('.files (the MFS API part)', function () {
       })
     })
 
+    it.only('files.add with progress options', (done) => {
```
diasdavid: only needs to be removed.

src/utils/request-api.js
```diff
@@ -160,7 +161,17 @@ function requestAPI (config, options, callback) {
...
+      const progressStream = new Transform({
+        transform: (chunk, encoding, cb) => {
+          options.progress(chunk.byteLength)
+          cb(null, chunk)
+        }
+      })
+      stream.pipe(progressStream).pipe(req)
+    } else {
+      stream.pipe(req)
+    }
```
diasdavid: @Kubuxu how does go-ipfs tell the CLI how much 'data' was ingested files? Isn't it returning a ndjson update?

Kubuxu: I am not sure but I am not sure what exactly  you are asking for.

> how much 'data' was ingested files?

diasdavid: @Kubuxu when running a `ipfs add BIGFILE`, how does the CLI know how much data has been processed by go-ipfs?

Kubuxu: Following object is being set by the daemon to the CLI as file is being digested:
```
type AddedObject struct {
	Name  string
	Hash  string `json:",omitempty"`
	Bytes int64  `json:",omitempty"`
	Size  string `json:",omitempty"`
}
```
It causes update of the progress bar.

diasdavid: Just confirmed with @whyrusleeping, go-ipfs http-api does indeed send a ndjson stream of updates when adding files if --progress is passed.

We can support streaming both in node and in the browser with stream-http (like we already do). 

@bmordan your PR needs to change to capture that and instead of measuring on bytes sent, measure on bytes reported back. Could you change to that, please?

bmordan: @diasdavid OK I'm having a look at this.

diasdavid:
continues here: https://github.com/ipfs/js-ipfs-api/pull/604

PR #597, opened 2017-09-08
Support specify hash algorithm in files.add

alanshaw:
### Specify multihash hashing alogrithm

This allows the optional param `hash` or `hashAlg` to be passed to `files.add` in order to specify the hash algorithm that should be used when generating the multihash(es) for the added content.

This PR also is more strict about what query string options `js-ipfs-api` supports by pulling out the supported params into a new query string object.

It also allows camel case aliases for params that are dash separated. This is because the HTTP API uses dash separated param names, but it is far easier to use camel case param names for internal APIs.

The other reason for allowing these aliases is because this code may be called when talking to a remote IPFS instance via the CLI where the qs params will have already been converted into camelCase.

### Change `object/get` and `object/data` to accept CIDs

This is necessary because `get-dagnode.js` uses `object/get` and `object/data` immediately after adding files. When specifying `--cid-version` or using `--hash` (which has the effect of setting `--cidversion=true`) `get-dagnode.js` is passed a CIDv1 not a multihash/CIDv0.

**Previously `get-dagnode.js` was extracting a multihash from a CID and sending that, effectively downgrading a CIDv1 to a CIDv0**.

That causes the object to not be found (see https://github.com/ipfs/js-ipfs-unixfs-engine/pull/185) since you have to retrieve using the same CID version as the data was stored.

This actually brings js-ipfs more in line with go-ipfs. Where both `object/get` and `object/data` accept a CID:

```sh
# With go ipfs daemon running...
$ ipfs add test.txt --cid-version=1 --raw-leaves=false
added zdj7WdDqrSLbsDNdtsHU5eWAeBukPoKfX9doKuTAphCowKjZR test.txt
$ ipfs object get zdj7WdDqrSLbsDNdtsHU5eWAeBukPoKfX9doKuTAphCowKjZR
{"Links":[],"Data":"\u0008\u0002\u0012\u001f(╯°□°）╯︵ ┻━┻\n\u0018\u001f"}
$ ipfs object data zdj7WdDqrSLbsDNdtsHU5eWAeBukPoKfX9doKuTAphCowKjZR
(╯°□°）╯︵ ┻━┻
```

### DIFF
--- a/src/object/get.js
+++ b/src/object/get.js
@@ -5,7 +5,7 @@ const dagPB = require('ipld-dag-pb')
 const DAGNode = dagPB.DAGNode
 const DAGLink = dagPB.DAGLink
 const bs58 = require('bs58')
-const cleanMultihash = require('../utils/clean-multihash')
+const CID = require('cids')
 const LRU = require('lru-cache')
 const lruOptions = {
   max: 128
@@ -14,7 +14,7 @@ const lruOptions = {
 const cache = LRU(lruOptions)
 
 module.exports = (send) => {
-  return promisify((multihash, options, callback) => {
+  return promisify((hash, options, callback) => {
     if (typeof options === 'function') {
       callback = options
       options = {}
@@ -24,13 +24,16 @@ module.exports = (send) => {
       options = {}
     }
 
+    let cid, b58Hash
+
     try {
-      multihash = cleanMultihash(multihash, options)
+      cid = new CID(hash)
+      b58Hash = cid.toBaseEncodedString()
     } catch (err) {
       return callback(err)
     }
 
-    const node = cache.get(multihash)
+    const node = cache.get(b58Hash)
 
     if (node) {
       return callback(null, node)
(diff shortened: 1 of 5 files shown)


src/object/get.js
```diff
@@ -14,7 +14,7 @@ const lruOptions = {
 const cache = LRU(lruOptions)
 
 module.exports = (send) => {
-  return promisify((multihash, options, callback) => {
+  return promisify((hash, options, callback) => {
```
diasdavid: If it is going to support both, better call it `cid`

src/object/get.js
```diff
@@ -24,21 +24,24 @@ module.exports = (send) => {
       options = {}
     }
 
+    let cid, b58Hash
```
diasdavid: We default to declare each variable in separate lines. Mind replacing for:

```
let cid
let b58Hash
```

Also, with the added support to cid, makes sense to call it cidB58Str

PR #598, opened 2017-09-08
Only hash support

bmordan:
Changes for the HTTP client to accept the only-hash option and the logic to skip writing to disk.


diasdavid:
Why closing this one @bmordan ?

santiagorp:
When is this going to be merged into master @bmordan?

PR #609, opened 2017-10-16
pass the config protocol to http requests

jfmherokiller:
While calling `request(config.protocol)` sets the right http protocol api to use it does not correctly pass the protocol information to node_modules/stream-http.

If the page is loaded over https and the daemon is located on http it fails to send the api request to the deamon.

This is because the code believes the daemon is located at https://127.0.0.1:5001/<insert api string here> when it is actually located at http://127.0.0.1:5001/<insert api string here>. 

Without this change the code throws an unrecoverable error and doesnt provide the user with information to fix the issue, with the change it still throws an error but it also lets the user resolve the issue by clicking the mixed content shield shield image and allowing insecure content.

### DIFF
--- a/src/utils/request-api.js
+++ b/src/utils/request-api.js
@@ -152,7 +152,8 @@ function requestAPI (config, options, callback) {
     path: `${config['api-path']}${options.path}?${qs}`,
     port: config.port,
     method: method,
-    headers: headers
+    headers: headers,
+    protocol: `${config.protocol}:`
   }, onRes(options.buffer, callback))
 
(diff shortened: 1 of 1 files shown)


diasdavid:
Thank you @jfmherokiller :)

jfmherokiller:
has this change been pushed out to npm/unpkg yet?

diasdavid:
@jfmherokiller yep, should be there since 14.3.6 -> https://github.com/ipfs/js-ipfs-api/commits/master

PR #615, opened 2017-10-24
add missing has-binary2 package

atfornes:


atfornes:
It is a socket.io dependency.

PR #616, opened 2017-10-24
Add instructions to install eslint

atfornes:


atfornes:
It is not needed with an updated npm. Currently works with `npm 5.5.1`

PR #629, opened 2017-11-17
send files HTTP request should stream

### LINKED ISSUE #628: createAddStream implementation buffers instead of streaming
`ipfs.createAddStrean` does not stream. Instead, if buffers all the files with open readable streams, which can eventually lead to file descriptor exhaustion.

Symptom well described in: https://github.com/ipfs/js-ipfs/issues/823

pgte:
The HTTP multipart implementation provides a streaming interface but doesn't allow adding files mid-flight.
This PR does that and bases all the APIs that need to send files on that.
Should fix #628 (and, by association, https://github.com/ipfs/js-ipfs/issues/823).


src/utils/multipart.js
```diff
@@ -0,0 +1,113 @@
...
+
+    content.once('end', () => {
+      this.push(NEW_LINE_BUFFER)
+      callback()
+
+      // TODO: backpressure!!! wait once self is drained so we can proceed
+      // This does not work
+      // this.once('drain', () => {
+      //   callback()
+      // })
```
diasdavid: No needed anymore?

src/files/create-add-stream.js
```diff
@@ -1,35 +1,5 @@
...
-      add(tuples, options, (err, res) => {
-        if (err) { return ds.emit('error', err) }
-        res.forEach((tuple) => ds.push(tuple))
-        ds.push(null)
-      })
-    }
-    callback(null, ds)
-  })
-}
+module.exports = (send) => SendFilesStream(send, 'add')
```
diasdavid: This file doesn't exist anymore, you need to rebase master onto this branch

src/files/add.js
```diff
@@ -1,51 +1,47 @@
...
-    } else if (opts.rawLeaves != null) {
-      qs['raw-leaves'] = opts.rawLeaves
-    }
+    const files = [].concat(_files)
 
-    if (opts.hash != null) {
-      qs.hash = opts.hash
-    } else if (opts.hashAlg != null) {
-      qs.hash = opts.hashAlg
-    }
```
diasdavid: Where did all of this options went?

src/files/add.js
```diff
@@ -1,51 +1,47 @@
...
+        return callback(err)
+      }
 
-    const request = { path: 'add', files: files, qs: qs, progress: opts.progress }
+      const concat = ConcatStream((result) => callback(null, result))
+      stream.once('error', callback)
+      stream.pipe(concat)
 
-    send.andTransform(request, (response, cb) => {
-      converter(ProgressStream.fromStream(opts.progress, response), cb)
```
diasdavid: Is this still doing Progress reports?

src/utils/request-api.js
```diff
@@ -6,7 +6,6 @@ const isNode = require('detect-node')
 const ndjson = require('ndjson')
```
diasdavid: While you are at it, wanna rename this file to `send-request`?

src/files/add.js
pgte: Here: https://github.com/ipfs/js-ipfs-api/blob/fix/add-stream/src/utils/send-files-stream.js#L87-L103

src/files/add.js
pgte: Yes, here: https://github.com/ipfs/js-ipfs-api/blob/fix/add-stream/src/utils/send-files-stream.js#L139-L141

src/utils/multipart.js
pgte: We do back-pressure per file item, but once inside a file we don't. So it's optimised for a lot of files, but not for large files. If that comes to be a problem, I think we should open a different issue to tackle it..

diasdavid: Mind writing this note as the comment as well?

pgte: Decided to fix individual file back-pressure in this PR: https://github.com/ipfs/js-ipfs-api/pull/629/commits/782beb9a628b5d9bde047400d1d44f6d57b17ef8

src/utils/request-api.js
pgte: Done. In 6f25c3f56888f534882747370efef729155859e5

src/utils/multipart.js
pgte: Had to limit it to Node only though, because browser HTTP request does not stream..
7403afc368c2992a4e19f4ac14b0815fbcb3c045

diasdavid:
This is fantastic @pgte !

Issue #611, opened 2017-10-20
UglifyJS


povi:
Running `npm run build` results in error:

```
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file: 

./node_modules/ipfs-api/src/utils/module-config.js:7
```

when trying to build a react.js project that has ipfs-api as a dependency.

As suggested by https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify , the project should be published pre-compiled.

atfornes:
I did some work on this, providing a prepublish routine for npm at https://github.com/atfornes/js-ipfs-api/commit/bd3a68ed814646bcb0672f29ec60356b48c8ceea.

Please, let us know if it helped

adrianmcli:
@atfornes Can you make a PR and ask the maintainers to republish the package with the pre-compiled code?

GoodBoy962:
@atfornes Could I simply add prepublish in node_modules/ipfs-api/package.json file?

GoodBoy962:
@povi add `help wanted` label, please.

atfornes:
@GoodBoy962, @adrianmcli, I am going to have a look into this today, hopefully sharing a useful pull request.

atfornes:
@GoodBoy962 @adrianmcli, please let me know if the pull request #630 solves your issues.

GoodBoy962:
@atfornes Thanks! Waiting for pr confirmation

diasdavid:
Hi all!

For a time, we published all of our dependencies transpiled for a while and then we decided to revert back. The reasons where the following:

- It would increase bundle size
- There are no source maps that can bring you back to the original line of code (for when a error happens) forcing you to npm link everything when debugging
- One more step to maintain

The reality at the time was that there was no uglifyJS with ES6 support, but today (roughly started 6 months ago) there is indeed uglifyJS with ES6 support and we use it for our dist version.

You can use that version by requiring the module as `require('ipfs-api/dist/index.min.js')` and avoid hitting this problem.

What I would like to see is create-react-app to upgrade to the latest uglifyJS. Asking the entire world to publish ES5 code -- https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/template/README.md#npm-run-build-fails-to-minify -- specially when pretty much the only thing we do is Arrow Functions. Let me enphasize that there is an update uglifyJS that does indeed support ES6

[...]

Issue #571, opened 2017-06-21
Add usage of dist/index.js as an install option


atfornes:
ipfs-api npm package includes a dist folder with the "webpacked" sources.

To include `node_modules/ipfs-api/dist/index.js` file makes avaliable the IpfsApi object at window. This is simpler than the browserify/webpack options provided to use Ipfs API at browsers.

diasdavid:
Thanks for the suggestion @atfornes :) However, do note that would stop users from treeshaking it themselves. Plus, note that the bundled versions is only usable in browsers, it would break Node.js usage.

Closing this one.

atfornes:
thanks @diasdavid for considering the proposal.

Regarding the "treeshacking", is what I successfully did in my experimental branch https://github.com/atfornes/js-ipfs-api/tree/withDag where by adding a `prepare` npm script I am able to automatically build ipfs api on npm install https://github.com/atfornes/js-ipfs-api/commit/79e2fb0ad656f5945cfa850db200c95a7bc7adbb.

Actually, this makes it really easy to use modified ipfs API branches, and was the way I managed to used the modified IPFS API branch in another repository: https://github.com/atfornes/embark-framework, where a npm compatible way of managing dependencies was proposed to substitute their current manner (including a built version of ipfs api in their source coude repository) (See https://github.com/iurimatias/embark-framework/pull/279) 

I suggest that you check this way of automatically building the package for the project.

Also, the use of this built source could be proposed only to be used in browsers, as currently the README propose using browserify.

Please let me know if any of this information is useful for the project and if I can contribute my effort to make it easier to develop broswer applications with ipfs api with this approach.

The automatic build of the project on `npm install` by using `prepare` script could be useful for the project with independence to the proposal of this issue.

diasdavid:
Thank you for following up @atfornes. I'll have to think a bit more about this and also check with others, but just for context, we already had tried this approach a year ago and it ended up to be super painful for the users and us to debug apps as it wasn't as trivial to pinpoint lines where errors were coming from and it required to educate users that we were doing that.

pcowgill:
I think this still is worth mentioning in the `README` as an option. Being able to try this in an environment like React Native where async iterators aren't supported by default was very helpful before I found the Babel config needed to get the `src` version of this lib working.

[...]

PR #630, opened 2017-11-17
Babel

atfornes:
Providing transpiled version of the code at lib/** and changing main file at package.json for compatibility reasons. See #611 

It "transpiles" and "builds" on `prepare` script to provide dist/ and lib/ folders not only in official builds but also in custom installations. See #571, https://stackoverflow.com/q/44680574/4928558


diasdavid:
Note: Just want to point out that @fazo96 provided some good intel on how to work around the fact that uglify2 doesn't support ES6, check it here https://github.com/ipfs/js-ipfs/issues/938#issuecomment-367280091

diasdavid:
Closing in favor of tracking this discussion at https://github.com/ipfs/js-ipfs/issues/938 //cc @vmx

PR #634, opened 2017-11-22
Fixed config.replace

### LINKED ISSUE #633: config.replace stopped working
@pgte with the new multipart streams PR, config.replace stopped working which I believe is just because it is using the wrong way to send the file. Mind checking?

pgte:
, which should now use the new streaming API.
(This wasn't caught on tests because `config.replace` tests are disabled <= this needs fixing)

Should fix #633.

### DIFF
--- a/src/config/replace.js
+++ b/src/config/replace.js
@@ -2,17 +2,15 @@
 
 const streamifier = require('streamifier')
 const promisify = require('promisify-es6')
+const SendOneFile = require('../utils/send-one-file')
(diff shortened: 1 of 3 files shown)


pgte:
Broke the build, I think:

job__2494_2_-_ipfs_js-ipfs-api_-_travis_ci

Looking into it..

diasdavid:
Thank you @pgte :) There you go @hacdias

PR #644, opened 2017-12-14
Updating CI files

VictorBjelkholm:
This commit updates all CI scripts to the latest version

PR #646, opened 2017-12-16
fix(files.add): handle weird directory names

richardschneider:
`glob` is interpreting the '[' in `weird [v0]` as a special magic character.  According to its documentation, a caller should be able to escape it with `\`; but it does not work. See the [glob issue](https://github.com/isaacs/node-glob/issues/277).

Should fix [a directory with an odd name](https://github.com/ipfs/js-ipfs-api/pull/644).

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "detect-node": "^2.0.3",
     "flatmap": "0.0.3",
     "glob": "^7.1.2",
-    "glob-escape": "0.0.2",
     "ipfs-block": "~0.6.1",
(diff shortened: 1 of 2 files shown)


richardschneider:
This is getting an existing test to work.  So I don't need to add another one.

richardschneider:
@diasdavid The test already exists.

diasdavid:
Thanks for submitting this patch @richardschneider :) 

PS: It's frustrating that we can see Appveyor getting green because there is no windows-386 of go-ipfs.

PR #647, opened 2017-12-16
test(refs): fix symlinks from git

### LINKED ISSUE #645: .refs test on Windows
The [.refs test](https://github.com/ipfs/js-ipfs-api/blob/master/test/refs.spec.js) uses a symlink file that is in the **git repository**.  On *nix `ipfs add` will use the contents of the target file.  On Windows, the file is not a symlink, so `ipfs add` uses the contents of the file not the target.

This breaks the tests, because the hash of the top level directory is now different on Windows. See https://github.com/ipfs/js-ipfs-api/pull/644

richardschneider:
Symlinks in a repo don't always clone well, especially on Windows. So if the 'hello-link' is not a symlink, then make it one.

Should fix #645 and help with #644

### DIFF
--- a/test/refs.spec.js
+++ b/test/refs.spec.js
@@ -9,6 +9,7 @@ const isNode = require('detect-node')
 const waterfall = require('async/waterfall')
 const path = require('path')
 const FactoryClient = require('./ipfs-factory/client')
+const fs = require('fs')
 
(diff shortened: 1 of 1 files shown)


diasdavid:
thanks @richardschneider !! :)

Issue #649, opened 2017-12-20
ipfs.files.add(Buffer) > TypeError: content.once is not a function


joaosantos15:
Hello, I'm using the ipfs-js-api and am having a problem with `ipfs.files.add()` when adding a Buffer.

I'm following the [this usage example](https://github.com/ipfs/js-ipfs-api/blob/master/examples/name-api/index.js) from  `js-ipfs-api/examples/name-api/index.js`:
```js
let buffer = Buffer.from(input.value)

  ipfs.add(buffer)
    .then(res => {
      publish(res[0].path)

      input.value = ''
    })
    .catch(err => {
      showStatus('failed to add the data', COLORS.error)
      console.error(err)
    })
}
```
This is my  code:
```js
var Promise = require('promise')
var Buffer = require('buffer/').Buffer
var ipfsAPI = require('ipfs-api')
var ipfs = ipfsAPI('/ip4/127.0.0.1/tcp/5001')

var myBuffer = Buffer.from('hello')

ipfs.files.add(myBuffer, function (err, result) {
  if (err) {
    console.log(err)
  } else {
    console.log(result[0].hash)
  }
})
```
And I'm getting this output (in Node.js):
```sh
Error: "files" must be a buffer, readable stream, or array of objects
```
In the browser, I can solve the problem by putting the Buffer inside an array (I don't know why that solves it though...):
```js
ipfs.files.add([myBuffer], function (err, result) {
```
But in Node.js I get this error:
```sh
TypeError: content.once is not a function
    at Multipart._pushFile ($proj/src/node_modules/ipfs-api/src/utils/multipart.js:80:13)
    at Multipart._maybeDrain ($proj/src/node_modules/ipfs-api/src/utils/multipart.js:48:14)
    at Multipart._transform ($proj/src/node_modules/ipfs-api/src/utils/multipart.js:40:10)
    at Multipart.Transform._read (_stream_transform.js:167:10)
    at Multipart.Transform._write (_stream_transform.js:155:12)
    at doWrite (_stream_writable.js:307:12)
    at writeOrBuffer (_stream_writable.js:293:5)
    at Multipart.Writable.write (_stream_writable.js:220:11)
    at eachSeries ($proj/src/node_modules/ipfs-api/src/utils/send-files-stream.js:52:35)
    at $proj/src/node_modules/async/internal/withoutIndex.js:9:16
```
Does anyone have any idea why this happens?
Thanks :)

joaosantos15:
[...]

PR #651, opened 2017-12-20
Buffer shim is no longer working in the Browser as expected

joaosantos15:
For #649 
Adds a failing test for when adding an NPM Buffer:
```js
let Buffer = require('buffer/').Buffer
let expectedBufferMultihash = 'QmWfVY9y3xjsixTgbd9AorQxH7VtMpzfx2HaWtsoUYecaX'
let file = Buffer.from('hello')

ipfs.files.add(file, ...
```


joaosantos15:
cc @diasdavid

test/files.spec.js
```diff
@@ -62,6 +62,21 @@ describe('.files (the MFS API part)', function () {
       })
     })
 
+    it('file.add with NPM Buffer', (done) => {
+      let Buffer = require('buffer/').Buffer
```
diasdavid: Why the `/`?

joaosantos15: It's to depend on that module specifically. I tested without it and it was using Nodejs core Buffer module.
I read about the [usage here](https://www.npmjs.com/package/buffer).
> To require this module explicitly, use require('buffer/') which tells the node.js module lookup algorithm (also used by browserify) to use the npm module named buffer instead of the node.js core module named buffer!

diasdavid:
@hacdias mind checking this one?

test/files.spec.js
hacdias: There is no `buffer` package [being imported](https://github.com/joaosantos15/js-ipfs-api/blob/master/package.json). Am I missing some point?

hacdias:
@joaosantos15 @diasdavid I've got one thing to ask: we support the native Node.js Buffer. So, why are trying to support `feross/buffer`? This way, we should support every single Buffer package which is on NPM.

Why don't we do something similar to what `js-ipfs` [does](https://github.com/ipfs/js-ipfs/blob/58ded8d2c3a3580dd79f77e02c9c47a4b99e6ed0/src/core/index.js#L55-L62)? Just make Buffer part of the type so when someone uses it in the browser it just works?

hacdias:
@diasdavid could you take a look at my previous comment? 😄

diasdavid:
@hacdias the feross/Buffer package is the browser shim that every Bundler (browserify, webpack, etc) uses to shim the Node.js Buffer.

I started getting this issue as well on #688 

image

I believe @vmx had found problems with feross/Buffer package. @vmx will https://github.com/feross/buffer/pull/178 fix this issue?

vmx:
@diasdavid I don't think https://github.com/feross/buffer/pull/178 will fix the issue. Though I think that once https://github.com/feross/buffer/issues/177 is done, it might fix the issue (178 is just a small part of that. But I won't spend more time on it until I got feedback on it).

diasdavid:
merging into a branch so that we all can edit.

diasdavid:
Follow on https://github.com/ipfs/js-ipfs-api/pull/689

PR #648, opened 2017-12-18
Updating CI files

VictorBjelkholm:
This commit updates all CI scripts to the latest version

Creating new PR as the old one was closed without being merged for some reason.

Ref: https://github.com/ipfs/js-ipfs-api/pull/644

PR #656, opened 2017-12-29
test: HTTP weirdness

richardschneider:
First reported in https://github.com/ipfs/js-ipfs-api/pull/648
```
4) 'deal with HTTP weirdness' tests does not crash if no content-type header is provided:
     Uncaught AssertionError: expected [Error: read ECONNRESET] to not exist 
      at ipfsAPI.config.replace (test\request-api.spec.js:28:28)
```

This is just a work around and does not fix the underlying problem.  I've changed the test to use `id` instead of `config.replace` and the test passes.

@diasdavid I'm think the `ECONNRESET` could be coming from push stream not being read by the server.  But perhaps it something else.  More eyeballs might help.  I think it's more of a timing/race issue than a specific windows issue

### DIFF
--- a/test/request-api.spec.js
+++ b/test/request-api.spec.js
@@ -24,7 +24,8 @@ describe('\'deal with HTTP weirdness\' tests', () => {
     })
 
     server.listen(6001, () => {
-      ipfsAPI('/ip4/127.0.0.1/tcp/6001').config.replace('test/fixtures/r-config.json', (err) => {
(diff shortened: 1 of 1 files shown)


richardschneider:
Fixed it by consuming the entire request, before responding.

diasdavid:
Interesting, this looks like a Node.js issue and not a js-ipfs-api. Thanks @richardschneider

PR #661, opened 2018-01-05
fix: name and key tests

hacdias:


test/name.spec.js
```diff
@@ -102,9 +100,7 @@ describe('.name', function () {
       return ipfs.name.resolve(name.Name)
         .then((res) => {
           expect(res).to.exist()
-          expect(res).to.be.eql({
-            Path: name.Value
-          })
+          expect(res).to.be.eql(name.Value)
```
diasdavid: @hacdias should it be `name.value`?

hacdias: Thanks for noticing 😄

hacdias:
I think it's done @diasdavid. For some reason, some testes (not related to name nor key api) failed on some builds but not on others.

diasdavid:
Let's fix the non related issues on a separate PR

PR #669, opened 2018-01-16
fix: normalize stats fields

hacdias:
That. Let's put them starting by a lower case letter like all of the others.

/cc @diasdavid


hacdias:
Somehow I managed to do this unwanted change (https://github.com/ipfs/js-ipfs-api/pull/669/files#diff-e2d5a5a9f16cf7bf8904992b82632eaf). How can I reverse it? I did copy the one's from master to this branch but for my computer they are equal files 😞

hacdias:
Is there something missing @diasdavid ?

PR #670, opened 2018-01-18
breaking change: normalize fields and flatten output

hacdias:
This normalizes the object fields names by lower casing them. It also flattens the output of `files.ls`.

`files.ls` now returns an array of objects like:

```
{
	name: 'something',
    type: 0,
    size: 0,
    hash: 'something'
}
```

`files.stat` now returns an object like:

```
{
    type: 'file',
    size: 0,
    hash: 'something',
	blocks: 152,
	cumulativeSize: 0
}
```

### DIFF
--- a/src/files/ls.js
+++ b/src/files/ls.js
@@ -2,16 +2,27 @@
 
 const promisify = require('promisify-es6')
 
+const transform = function (res, callback) {
+  callback(null, res.Entries.map((entry) => {
+    return {
+      name: entry.Name,
+      type: entry.Type,
+      size: entry.Size,
+      hash: entry.Hash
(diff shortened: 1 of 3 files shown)


hacdias:
@diasdavid mind taking a look at this?

diasdavid:
@hacdias a "refactor" that changes the output is not a refactor, it is a breaking change. It is justified? Do we need it?

hacdias:
@diasdavid the justification is to make it all consistent.

hacdias:
Ping @diasdavid

diasdavid:
sounds good @hacdias

PR #675, opened 2018-01-22
feat(breaking change): stream to value on files read

hacdias:
Use `streamToValue` on `files.read` to return a beautiful Buffer instead of a stream.


hacdias:
The tests seem to be passing (some not related to files.read randomly fail). ping @diasdavid  😄

diasdavid:
@hacdias let's sync up first with all these API changes. We can't simply break the API without understanding why and have a good way to notify users.

Spec first out the Files API and then propose the changes.

hacdias:
Sure, makes sense @diasdavid

hacdias:
@diasdavid the spec is done and merged 😄

diasdavid:
@hacdias can we get CI to pass here?

hacdias:
image

@diasdavid Travis seems to be passing

PR #680, opened 2018-01-26
fix: exception when dir is empty

hacdias:
bugfix @diasdavid


hacdias:
The failing tests are not related to this bugfix.

Issue #683, opened 2018-01-27
Understand why Webpack is running out of Memory


diasdavid:
Currently, CI fails with:

```
Test Browser
<--- Last few GCs --->
  332336 ms: Mark-sweep 1341.7 (1418.5) -> 1335.0 (1416.2) MB, 403.5 / 0.0 ms [allocation failure] [GC in old space requested].
  332750 ms: Mark-sweep 1335.0 (1416.2) -> 1335.0 (1416.2) MB, 414.8 / 0.0 ms [allocation failure] [GC in old space requested].
  333142 ms: Mark-sweep 1335.0 (1416.2) -> 1335.0 (1401.2) MB, 391.3 / 0.0 ms [last resort gc].
  333522 ms: Mark-sweep 1335.0 (1401.2) -> 1335.0 (1401.2) MB, 380.5 / 0.0 ms [last resort gc].
<--- JS stacktrace --->
==== JS stack trace =========================================
Security context: 0x25fbd6fcf781 <JS Object>
    1: toString [buffer.js:490] [pc=0x83681c5517c] (this=0x3368e86806a9 <an Uint8Array with map 0x580e41ffe69>)
    2: /* anonymous */(aka /* anonymous */) [/home/travis/build/ipfs/js-ipfs-api/node_modules/karma-webpack/lib/karma-webpack.js:255] [pc=0x83681e0f89f] (this=0x25fbd6f04381 <undefined>,err=0x25fbd6f04381 <undefined>,content=0x3368e86806a9 <an Uint8Array with map 0x580e41ffe69>)
    3...
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [node]
 2: 0x10a03cc [node]
 3: v8::Utils::ReportApiFailure(char const*, char const*) [node]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [node]
 5: v8::internal::Factory::NewRawTwoByteString(int, v8::internal::PretenureFlag) [node]
 6: v8::internal::Factory::NewStringFromUtf8(v8::internal::Vector<char const>, v8::internal::PretenureFlag) [node]
 7: v8::String::NewFromUtf8(v8::Isolate*, char const*, v8::String::NewStringType, int) [node]
 8: node::StringBytes::Encode(v8::Isolate*, char const*, unsigned long, node::encoding) [node]
[...]

PR #684, opened 2018-01-29
Some fixes regarding the Stats API

hacdias:
- [x] Make `bitswap.stat` **equal to** `stats.bitswap`

This is also a breaking change.


diasdavid:
CI is failing due to #683 . Thanks @hacdias

PR #686, opened 2018-01-30
feat(breaking change): use stream on stats.bw

hacdias:
`stats.bw` has an option called `poll` which allows the user to "tell" IPFS to send bandwidth stats each interval of time. We were collecting the response before calling the callback so that option would be useless. 

This way it sends a stream to the callback so the user can listen to `data` event which contains the Object.

Interface IPFS core update: https://github.com/ipfs/interface-ipfs-core/pull/211

### DIFF
--- a/src/stats/bw-pull-stream.js
+++ b/src/stats/bw-pull-stream.js
@@ -5,19 +5,11 @@ const pull = require('pull-stream')
 const transformChunk = require('./bw-util')
 const deferred = require('pull-defer')
 
-const transform = () => (read) => (abort, cb) => {
-  read(abort, (err, data) => {
-    console.log(data)
-    if (err) return cb(err)
-    cb(null, transformChunk(data))
-  })
-}
-
 module.exports = (send) => {
   return (hash, opts) => {
     opts = opts || {}
 
-    const p = deferred.through()
+    const p = deferred.source()
 
     send({
       path: 'stats/bw',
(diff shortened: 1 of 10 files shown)


hacdias:
Needed for https://github.com/ipfs/js-ipfs/pull/1198 /cc @diasdavid 😄

VictorBjelkholm:
@hacdias this should be a argument to the function that changes if it just returns the values once or if you get a stream back, rather than just having one of them.

hacdias:
Note to self: Pull Stream version not working. Still don't know why.

src/stats/bw-pull-stream.js
```diff
@@ -0,0 +1,36 @@
...
+
+    send({
+      path: 'stats/bw',
+      qs: opts
+    }, (err, stream) => {
+      if (err) {
+        return p.end(err)
+      }
+
+      pull(toPull(stream), p)
```
dignifiedquire: this could be simply

pull(pull.map(transformChunk, p)

if I understand it correct

hacdias: Thanks :D

hacdias:
This is passing with the tests from ipfs/interface-ipfs-core#211 😄 

I think it's ready /cc @diasdavid

diasdavid:
There you go :)

image

hacdias:
Thanks @diasdavid! Once the tests pass I'll ping you.

hacdias:
@diasdavid the failing tests are unrelated to Stats (stats are passing). I think we can merge this and release a new version 😄 so I can continue https://github.com/ipfs/js-ipfs/pull/1198

PR #688, opened 2018-02-07
Fix 683

diasdavid:


test/bitswap.spec.js
```diff
@@ -9,15 +9,16 @@ chai.use(dirtyChai)
 const IPFSApi = require('../src')
 
 const DaemonFactory = require('ipfsd-ctl')
-const df = DaemonFactory.create()
+const df = DaemonFactory.create({remote: true, port: 30003})
```
diasdavid: @dryajov I'm confused. From docs seems that this is the way to spawn remove daemons, however, it seems it didn't use it even once for all the tests. Is there any magic trick? Truth be told, tests are failing and so I suspect this to be it.

package.json
```diff
@@ -9,12 +9,14 @@
...
-    "ipfs": false
+    "ipfs": false,
+    "./node_modules/ipfsd-ctl/src/ipfsd-in-proc.js": false,
+    "./node_modules/ipfsd-ctl/src/factory-in-proc.js": false
   },
   "scripts": {
     "test": "aegir test",
     "test:node": "aegir test -t node",
-    "test:browser": "aegir test -t browser",
+    "test:browser": "node --max-old-space-size=8192 ./node_modules/.bin/aegir test -t browser",
```
diasdavid: Still no luck. Kind of running out of my options here :(

diasdavid: @vmx @pgte @dignifiedquire any Webpack wisdom to solve this issue?

diasdavid: Actually, spoke too soon, IT RUUUNS :D 

image

test/bitswap.spec.js
dryajov: By default, it detects if not running in node, it will assume we're in the browser and use the remote daemon. So, no arguments to `create` will do what you expect depending on env. This was a specific requirement we discussed during the revamp of `ipfsd`.

dryajov: So no need for explicit `remote` unless you have to specify the port.

diasdavid:
remaining issues are being handled on #651

PR #693, opened 2018-02-16
Make `ls` links' depth dynamic.

JonKrone:
To be able to prettily print `ipfs ls -r`, we need to know the depth of each file from the `ls` command.

It was previously hardcoded a 1, though I don't know why. I'm passing the depth along from a change I made in js-ipfs [here](https://github.com/ipfs/js-ipfs/pull/1222/files#diff-dcaf16a1df33e0e8f40195e527b9294dR293).

The b58 hash test for `ls` should fail in CI because the above change to js-ipfs is required to receive the links' depth. Unfortunately this is a cyclical dependency. I can fix it by breaking out the change to js-ipfs or we can merge and fix if something comes up. lemme know

ref: https://github.com/ipfs/js-ipfs/pull/1222

### DIFF
--- a/src/ls.js
+++ b/src/ls.js
@@ -41,7 +41,7 @@ module.exports = (arg) => {
         size: link.Size,
         hash: link.Hash,
         type: typeOf(link),
-        depth: link.Depth
+        depth: link.Depth || 1
       }))
 
       callback(null, result)
--- a/src/ls.js
+++ b/src/ls.js
@@ -36,12 +36,12 @@ module.exports = (arg) => {
       }
 
       result = result.map((link) => ({
-        depth: 1,
         name: link.Name,
         path: args + '/' + link.Name,
         size: link.Size,
         hash: link.Hash,
-        type: typeOf(link)
+        type: typeOf(link),
+        depth: link.Depth
       }))
 
       callback(null, result)


diasdavid:
@JonKrone CI fails because tests fail: https://travis-ci.org/ipfs/js-ipfs-api/jobs/342470865#L6136-L6193

JonKrone:
> The b58 hash test for ls should fail in CI because [the above change ](https://github.com/ipfs/js-ipfs/pull/1222/files#diff-dcaf16a1df33e0e8f40195e527b9294dR293)to js-ipfs is required to receive the links' depth. Unfortunately this is a cyclical dependency. I can fix it by breaking out the change to js-ipfs or we can merge and fix if something comes up. lemme know

Preferred approach?

diasdavid:
>  js-ipfs is required to receive the links' depth. 

@JonKrone js-ipfs-api is tested against go-ipfs, not js-ipfs. No change in js-ipfs should be required.

JonKrone:
@diasdavid Thanks! The failing tests are from interface-ipfs-core. I did something that made these tests pass at one point but I must've mislead myself somehow. Lots to these projects. Anyhow, I just made a change, let me know if that's not the right way.

When a feature of js-ipfs diverges from go-ipfs and requires a change in a dependency that uses both go-ipfs and js-ipfs. I know we want to maintain as much similarity as possible but are the interfaces a core that we can extend from or do we want to strictly maintain them? Should we update any interfaces for this change?

JonKrone:
Tests failed this time due to an OOM, heap out of memory: https://travis-ci.org/ipfs/js-ipfs-api/jobs/343565256#L3547

diasdavid:
Thanks @JonKrone!

> Tests failed this time due to an OOM, heap out of memory

Just the browser ones and on CI given https://github.com/ipfs/js-ipfs-api/issues/683. Still figuring that one out.

PR #695, opened 2018-02-19
add support for ipfs files stat --with-local

MichaelMure:
So, I upgraded my js-ipfs-api and found out that the support for `ipfs files stat --with-local` was broken after 98000af3fe9c15aa7f6a594f760bff2f3d373095 (the response object is not relayed entirely anymore, so the new values get dropped).

Another problem I had and that is not solved yet by this PR is that due to the migration in go-ipfs to the new `go-ipfs-cmds` framework (https://github.com/ipfs/go-ipfs/pull/4638), the command `ipfs files stat` now reply with a stream which is not converted properly in values by js-ipfs-api. There is two way to solve that problem:
- convert the stream in a single value in js-ipfs-api
- use `cmds.EmitOnce` in go-ipfs to hint the ReponseEmitter and reply without streaming

Do you have a preference ?

CC @whyrusleeping, @Stebalien

### DIFF
--- a/src/files/stat.js
+++ b/src/files/stat.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const promisify = require('promisify-es6')
+const _ = require('lodash')
 
 const transform = function (res, callback) {
   callback(null, {
@@ -8,7 +9,10 @@ const transform = function (res, callback) {
     blocks: res.Blocks,
     size: res.Size,
     hash: res.Hash,
-    cumulativeSize: res.CumulativeSize
+    cumulativeSize: res.CumulativeSize,
+    withLocality: res.WithLocality || false,
+    local: res.Local || undefined,
+    sizeLocal: res.SizeLocal || undefined
   })
 }
 
@@ -18,6 +22,9 @@ module.exports = (send) => {
       callback = opts
       opts = {}
     }
+
+    opts = _.mapKeys(opts, (v, k) => _.kebabCase(k))
+
     send.andTransform({
       path: 'files/stat',
       args: args,
--- a/src/files/stat.js
+++ b/src/files/stat.js
@@ -2,17 +2,24 @@
 
 const promisify = require('promisify-es6')
 const _ = require('lodash')
+const streamToValue = require('../utils/stream-to-value')
 
 const transform = function (res, callback) {
-  callback(null, {
-    type: res.Type,
-    blocks: res.Blocks,
-    size: res.Size,
-    hash: res.Hash,
-    cumulativeSize: res.CumulativeSize,
-    withLocality: res.WithLocality || false,
-    local: res.Local || undefined,
-    sizeLocal: res.SizeLocal || undefined
+  return streamToValue(res, (err, data) => {
+    if (err) {
+      return callback(err)
+    }
+
+    callback(null, {
+      type: data[0].Type,
+      blocks: data[0].Blocks,
+      size: data[0].Size,
+      hash: data[0].Hash,
+      cumulativeSize: data[0].CumulativeSize,
+      withLocality: data[0].WithLocality || false,
+      local: data[0].Local || null,
+      sizeLocal: data[0].SizeLocal || null
+    })
   })
 }
 
--- a/test/files.spec.js
+++ b/test/files.spec.js
@@ -282,7 +282,10 @@ describe('.files (the MFS API part)', function () {
         size: 12,
         cumulativeSize: 20,
         blocks: 0,
-        type: 'file'
+        type: 'file',
+        withLocality: false,
+        local: undefined,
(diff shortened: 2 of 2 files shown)


MichaelMure:
If the best choice is to adjust go-ipfs, it would be nice to choose before the upcoming 0.4.14.

MichaelMure:
Someone to answer my question ?

diasdavid:
> after 98000af (the response object is not relayed entirely anymore, so the new values get dropped).

What are the values missing? Mind updating that with the docs + tests on https://github.com/ipfs/interface-ipfs-core ?

>  the command ipfs files stat now reply with a stream which is not converted properly in values by js-ipfs-api.

The endpoint should be setting the headers so that js-ipfs-api figures that on the fly. Is that not the case?

MichaelMure:
> What are the values missing? Mind updating that with the docs + tests on https://github.com/ipfs/interface-ipfs-core ?

New values comes from https://github.com/ipfs/go-ipfs/pull/4638. I will update interface-ipfs-core as well.

> The endpoint should be setting the headers so that js-ipfs-api figures that on the fly. Is that not the case?

Actually, I'm not really sure what's going on. This is what I get when I do a request with the last go-ipfs:

image

Here, `res` is a `DestroyableTransform` which is part of the `through2` npm package. This doesn't happen if `cmds.EmitOnce` is used in go-ipfs.
go-ipfs current master
js-ipfs-api v17.5.0

vmx:
@MichaelMure I had a problem which was similar, though after a quick look at the stats source I think I was triggering a different code path. Nonetheless it might be worth to check. Could you set a breakpoint at https://github.com/ipfs/js-ipfs-api/blob/f81dce5bf54df3a3578e4786bfb741d761cd1297/src/utils/send-files-stream.js#L135 and see if you come across it? If yes, then it might be related to https://github.com/ipfs/js-ipfs-api/issues/696.

Another hack that worked in my case when I didn't get any value back from a stream was hard-coding the `chunkedObjects` to `false` at https://github.com/ipfs/js-ipfs-api/blob/f81dce5bf54df3a3578e4786bfb741d761cd1297/src/utils/send-request.js#L34.

MichaelMure:
The breakpoint at https://github.com/ipfs/js-ipfs-api/blob/f81dce5bf54df3a3578e4786bfb741d761cd1297/src/utils/send-files-stream.js#L135  is not triggered.

Hardcoding `chunkedObjects` to false indeed solve the problem. It's the same effect as using `cmds.EmitOnce` in go-ipfs.

The root problem might linked to 210dabb3a857bd6e832f37c64b90ad0fcdba6519 that introduce changes on how data is decoded when `chunkedObjects` is true.

vmx:
@pgte There seem to be more issues with streamed responses.

MichaelMure:
@diasdavid does that make sense for you to set the optional values of the http request to default value as I did in this PR ? Or should the keys be missing entirely ?

For reference, the go struct:
```go
type statOutput struct {
	Hash           string
	Size           uint64
	CumulativeSize uint64
	Blocks         int
	Type           string
	WithLocality   bool   `json:",omitempty"`
	Local          bool   `json:",omitempty"`
	SizeLocal      uint64 `json:",omitempty"`
}
```

MichaelMure:
@diasdavid spec completed here: https://github.com/ipfs/interface-ipfs-core/pull/227
I'll add some test when I'm sure what the interface should look like.

MichaelMure:
PR updated with better default values and a fixed test

hacdias:
@MichaelMure the code looks code. Could you fix [the tests](https://github.com/MichaelMure/js-ipfs-api/blob/statslocal/test/stats.spec.js) please?

MichaelMure:
@hacdias please note that you need go-ipfs master and https://github.com/ipfs/js-ipfs-api/issues/696 fixed (I used the `cmds.EmitOnce` workaround in go-ipfs to have the test pass).

hacdias:
@MichaelMure curiously enough, the tests passed here. Anyway, I'm approving this changes because they seem to be correct. We can wait until those issues are also fixed to merge this one.

MichaelMure:
So, huuuu, I actually had trouble when the http reply was a stream because I ... didn't unwrap the stream with `streamToValue`. So #696 might have been a problem but my code was incorrect as well. Well ...

I still have trouble running the tests properly (lots of unrelated tests are failing), so I'm not sure everything is correct. That said I'm currently using a fork of js-ipfs-api in my project (https://github.com/MichaelMure/js-ipfs-api/tree/fork) and it works properly now.

diasdavid:
@MichaelMure released interface-ipfs-core with your changes, please rebase master onto this branch and update the interface-ipfs-core dep.

PR #697, opened 2018-02-20
Chore: set package.json's main to dist

sohkai:
The `main` field should probably be pointing to the built version, rather than the original src. Without this, some module bundlers (and uglify) will choke.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ipfs-api",
   "version": "18.1.1",
   "description": "A client library for the IPFS HTTP API",
-  "main": "src/index.js",
+  "main": "dist/index.js",
   "browser": {
     "glob": false,
     "fs": false,


diasdavid:
@sohkai are you using the latest uglify-es? The issue is that uglify2 didn't support ES6, uglify3 will do it, it is what we use to get that dist version https://github.com/ipfs/aegir/blob/master/package.json#L81

sohkai:
Even so, I'm of the belief that the main output should always be the transpiled version so that end users don't have to build it themselves.

Getting tree-shaking from `import` can be done using the pseudostandard [`module`](https://github.com/rollup/rollup/wiki/pkg.module) field.

diasdavid:
Closing in favor of tracking this discussion at https://github.com/ipfs/js-ipfs/issues/938 The solution won't require this step. //cc @vmx

PR #699, opened 2018-02-22
fix: remove argument from .stats.bw*

hacdias:
Removing an argument that wasn't supposed to exist.

### DIFF
--- a/src/stats/bw-readable-stream.js
+++ b/src/stats/bw-readable-stream.js
@@ -5,7 +5,7 @@ const pump = require('pump')
 const transformChunk = require('./bw-util')
 
 module.exports = (send) => {
-  return (hash, opts) => {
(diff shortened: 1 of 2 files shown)


src/stats/bw-readable-stream.js
```diff
@@ -5,7 +5,7 @@ const pump = require('pump')
 const transformChunk = require('./bw-util')
 
 module.exports = (send) => {
-  return (hash, opts) => {
+  return (opts) => {
```
diasdavid: Hm.. how doesn't this change break tests?

hacdias: No. Although the build failed due to 'JavaScript heap out of memory'.

diasdavid: got it

Issue #509, opened 2017-01-16
A function that returns the IPFS addresses of files without uploading them


yxliang01:
This would be useful for many applications, especially for experiments.

jbenet:
should be able to be done with `ipfs add --only-hash` (so `ipfs.add({"only-hash": true}, ...)` i think? cc @diasdavid @dignifiedquire

diasdavid:
Exactly, you can do that with the `only-hash` option.

yxliang01:
Thanks. But, when I pass opts with the key `only-hash` to `ipfs.util.addFromFs`, this function never calls the callback.. Help needed.

yxliang01:
By try and error and some analysis, I think has nothing deal with `ipfs.util.addFromFs`. If the file was uploaded before or run with `ipfs add --only-hash` before, `ipfs.util.addFromFs` calls the `cb` almost immediately. But, if the file was brand new to the IPFS, it just hanged without error. I tried to understand the source code, but I haven't really understood it yet...

aakilfernandes:
@yxliang01 you can take a look at this for inspiration

https://github.com/SafeMarket/ux/blob/master/lib/utils/getUnixMultihash.js

I believe the problem is that the multihash depends on the filesystem.

yxliang01:
Thanks @aakilfernandes . While this is handy, I still think `only-hash` should work and do the job perfectly because it can guarantee that no matter how IPFS changes its way calculating the hash, the result is still accurate...

yxliang01:
So... I think this is a bug instead of a question. Anyone has idea?

yxliang01:
Okay, I think I know why. After dipping into the code, I found that add.js doesn't have options support which contradicts the documentation `ipfs.files.add(data, [options], [callback])` at https://github.com/ipfs/interface-ipfs-core/tree/master/API/files#add .

santiagorp:
What happens with this? I saw a branch containing with this which was deleted today.

sterlu:
Any news on this?

JonKrone:
It looks like `files.add` does not support `onlyHash` right now but the unix dagNode creator thing (still trying to figure it all out), [ipfs/js-ipfs-unixfs-engine](https://github.com/ipfs/js-ipfs-unixfs-engine), does so I think we might just need to connect the two.

[...]

PR #700, opened 2018-02-23
feat: only-hash option for add/addFromFS/addFromURL

JonKrone:
This adds `only-hash` as a potential querystring arg for files.add, util.addFromFS, and util.addFromURL: `ipfs.add(<content>, { only-hash: true })`.

only-hash allows users to generate a hash for some content without actually adding the content to ipfs.

related issue: https://github.com/ipfs/js-ipfs-api/issues/509

needed for js-ipfs: https://github.com/ipfs/js-ipfs/pull/1233


lgierth:
Does this also support `-n`?

```
  -n,          --only-hash           bool   - Only chunk and hash - do not write to disk.
```

JonKrone:
> Does this also support -n?

I was thinking that alias support like that is only available through a CLI, (js-ipfs [here](https://github.com/ipfs/js-ipfs/pull/1233/files#diff-606d3a23b2aff2a65d1e977760345e05R143)). Should this API also support aliases? Please explain if I'm missing something, but would you expect `ipfs.add(<content>, { n: true })` to be valid?

lgierth:
Ah I'm sorry, I didn't notice we're in js-ipfs-api

JonKrone:
Haha, no worries 😉. Thanks for the look-over, though.

JonKrone:
CI question: Does anyone know why ESLint finds errors but doesn't report them [here](https://travis-ci.org/ipfs/js-ipfs-api/jobs/345360314#L2822)? This has tricked me a few times, I think.

lgierth:
> Should this API also support aliases?

Actually: maybe. go-ipfs supports option aliases on the API, e.g. `/api/v0/refs?r=true&arg=QmFoo` (because that's where most of the CLI options eventually end up). @diasdavid's call.

lgierth:
But again this is just options objects passed to an API function yeah? Then probably not.

JonKrone:
> Actually: maybe. go-ipfs supports option aliases on the API

Cool, I think js-ipfs could change that as a mild `chore` across the route handlers later on. Do you know why go decided to? Aliases make sense for a CLI because you get lots of human interaction so shortcuts are handy, but most code through an API is generated so aliases don't have that big of an impact.

JonKrone:
CI is failing on object.patch tests: https://travis-ci.org/ipfs/js-ipfs-api/jobs/345383716#L3541

These tests do not fail locally but are also failing on VMX's most recent PR: https://github.com/ipfs/js-ipfs-api/pull/701.

diasdavid:
Alias are supported for some methods (early days) but we decided to bet on the full options names for docs. (Aliases mean + docs and + dup tests)

diasdavid:
@vmx could https://github.com/ipld/js-ipld-dag-pb/releases/tag/v0.13.1 be breaking the tests? HAven't looked too much into it.

vmx:
@diasdavid: Nope, I've just publish js-ipld-dag-pb 0.13.1 just 30mins ago.

JonKrone:
Current status: ready for review / merge.

CI failure is [`pubsub multiple nodes connected: empty() list when no topics are subscribed`](https://travis-ci.org/ipfs/js-ipfs-api/jobs/346327672#L3336).

The `object.patch` CI tests that were failing for vmx and me are passing now for both PRs.

diasdavid:
@JonKrone any reason why to use a different github account for your commits?

See how things are commited by @jkrone rather than @JonKrone 
image

JonKrone:
@diasdavid No reason. It is not a different Github account, just a local `git config user.name` setting that was set to jkrone. I've synced the names and pushed a commit that reflects it.

diasdavid:
let's get https://github.com/ipfs/js-ipfs-api/pull/705 by @vmx in first to that we can see CI green in every PR :)

diasdavid:
@JonKrone please rebase master onto this branch when you have a chance.

diasdavid:
@JonKrone merging this branch into a branch of the repo itself so that we can collaborate 
image

PR #703, opened 2018-02-28
chore: disable Browser tests on CI for now

vmx:
Currently the Browser tests fail to a known issue. Therefore disable
them for now and run the tests only on node. Some CI is better
than none CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
 
 script:
   - npm run lint
-  - npm run test
+  - npm run test:node
   - npm run coverage
 
 before_script:
--- a/circle.yml
+++ b/circle.yml
@@ -4,6 +4,8 @@ machine:
     version: stable
 
 test:
+  override:
+    - npm run test:node
   post:
     - npm run coverage -- --upload --providers coveralls
 


vmx:
Does Travis need to run coverage reports? The tests pass, but if they are re-run for the coverage report they randomly fail :-/

VictorBjelkholm:
@vmx yeah, coverage tests runs slower than normal tests. I'm unsure of why we run tests, then we run tests again with coverage. We should just run tests with coverage or just run tests, not both IMHO.

VictorBjelkholm:
I am not super happy about disabling a whole class of tests. js-ipfs-api browser tests have been passing in the past, and there was a bad commit (or two) that changed that. That no one has had enough time to fix them yet is bad as it is, disabling browser tests will just make us ignore it longer.

vmx:
@VictorBjelkholm Hence the Browser tests are not disabled on Jenkins. Those will fail until it is properly fixed. But at least we are not totally blocked until it is fixed. I think it is a good trade-off. What do you think?

diasdavid:
@vmx seems that Jenkins is not being able to handle the memory request from it.

vmx:
I'm closing this one as I'm really close to a proper fix.

PR #705, opened 2018-03-05
fix: don't create one webpack bundle for every test file

vmx:
This commit introduces a new experimental branch of AEgir which
should fix the issue with taking to much memory when running the
Browser based tests. If it turns out to be successful, there
will be a new AEgir release containing that fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,17 +12,17 @@
     "ipfs": false
   },
   "scripts": {
-    "test": "cross-env NODE_OPTIONS='--max-old-space-size=8192' aegir test",
+    "test": "aegir test",
     "test:node": "aegir test -t node",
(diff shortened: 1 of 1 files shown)


dryajov:
@vmx the latest `ipfsd-ctl` should make the `9999` used port error go away.

package.json
```diff
@@ -65,15 +65,15 @@
     "url": "https://github.com/ipfs/js-ipfs-api"
   },
   "devDependencies": {
-    "aegir": "^13.0.0",
+    "aegir": "https://github.com/ipfs/aegir#karma-webpack-bundle-alt2",
```
diasdavid: Can you release aegir and update it here?

vmx: I'll do once I'm done with my Windows fixes.

diasdavid: Which ones?

vmx: The ones I'm working on atm.

PR #707, opened 2018-03-06
fix: validates and serializes cid argument to dht.provide

alanshaw:
This PR validates that the `cid` argument passed to provide is a CID and also serializes it to a base58 encoded string as the serialization job done by `qs` appears to be ignored by go-ipfs.

Please see [this discussion](https://github.com/ipfs/interface-ipfs-core/pull/221#issuecomment-370938018) for the context.

### DIFF
--- a/src/dht/provide.js
+++ b/src/dht/provide.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const promisify = require('promisify-es6')
+const CID = require('cids')
 
 module.exports = (send) => {
   return promisify((cids, opts, callback) => {
@@ -18,6 +19,13 @@ module.exports = (send) => {
 
     if (!Array.isArray(cids)) {
       cids = [cids]
+    }
+
+    // Validate CID(s) and serialize
+    try {
+      cids = cids.map(cid => new CID(cid).toBaseEncodedString('base58btc'))
+    } catch (err) {
+      return callback(err)
     }
 
     send({


src/dht/provide.js
```diff
@@ -20,6 +21,13 @@ module.exports = (send) => {
       cids = [cids]
     }
 
+    // Validate CID(s) and serialize
+    try {
+      cids = cids.map(cid => new CID(cid).toBaseEncodedString('base58btc'))
```
diasdavid: I believe that the DHT will only like CIDs V0 and not CIDs V1. Could you double check with both?

alanshaw: I've added a test https://github.com/ipfs/interface-ipfs-core/pull/221/commits/bde22b2db3dd72e895619c5694b5dd6ed14318c7

go-ipfs (0.4.13) does not error when provide a CIDv1:

screen shot 2018-03-07 at 10 11 54

In js-ipfs, [`repo.blocks.has`](https://github.com/ipfs/js-ipfs/blob/0920ff4712cd979192519deec432d3413fa1a992/src/core/components/dht.js#L119) returns false, so the test fails in https://github.com/ipfs/js-ipfs/pull/1250 but [`libp2pNode.contentRouting.provide`](https://github.com/ipfs/js-ipfs/blob/0920ff4712cd979192519deec432d3413fa1a992/src/core/components/dht.js#L133) does not error.

PR #709, opened 2018-03-09
fix regression on files.add

diasdavid:


diasdavid:
verified that tests pass locally. cant wait for #705

Issue #706, opened 2018-03-06
Tests fail on Windows


vmx:
On Jenkins the Windows tests constantly fail: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/PR-705/3/pipeline/15/. It looks like an actual issue, not like a CI problem.

I'll take a look.

vmx:
More tests fail due to: https://github.com/ipfs/js-ipfs-api/pull/711

vmx:
Often on Windows (but not exclusively) stopping the go-ipfs fails. In the Browsers the error message is different from Node, hence the test fails: https://github.com/ipfs/interface-ipfs-core/pull/235

vmx:
I'm closing this on in favour of https://github.com/ipfs/js-ipfs-api/issues/713

PR #710, opened 2018-03-09
fix: disable Browser test on Windows

vmx:
`process.platform` is only available in Node.js and not in the Browsers.
Hence the `diag.spec.js` tests weren't correctly skipped on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
   },
   "devDependencies": {
     "aegir": "^13.0.6",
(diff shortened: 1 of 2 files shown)


diasdavid:
@vmx can I get a list of the "PRs to get CI green"?

vmx:
@diasdavid As it's mostly Windows related, i'll use #706 for further information on getting CI green.

PR #711, opened 2018-03-09
Use a different remote server for test

vmx:
The previous server had issues, so the test timed out. Hopefully
this on is more stable.

### DIFF
--- a/test/util.spec.js
+++ b/test/util.spec.js
@@ -113,9 +113,9 @@ describe('.util', () => {
   })
 
   it('.urlAdd http with redirection', (done) => {
-    ipfs.util.addFromURL('http://covers.openlibrary.org/book/id/969165.jpg', (err, result) => {
+    ipfs.util.addFromURL('https://coverartarchive.org/release/6e2a1694-d8b9-466a-aa33-b1077b2333c1', (err, result) => {
       expect(err).to.not.exist()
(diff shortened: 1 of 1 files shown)


vmx:
Some tests fail, but at least not the `.urlAdd http with redirection` one anymore.

test/util.spec.js
```diff
@@ -113,9 +113,9 @@ describe('.util', () => {
   })
 
   it('.urlAdd http with redirection', (done) => {
-    ipfs.util.addFromURL('http://covers.openlibrary.org/book/id/969165.jpg', (err, result) => {
+    ipfs.util.addFromURL('https://coverartarchive.org/release/6e2a1694-d8b9-466a-aa33-b1077b2333c1', (err, result) => {
```
diasdavid: @vmx can we use one file used from the ipfs gateway?

vmx: If you give me a link, for sure.

I don't know how to find URLs that redirect. I wanted a quick solution to the problem. The real fix would be one that doesn't need an Internet connection at all, but that's out of scope for me atm.

diasdavid: nvm. Missed the need for a redirect :)

vmx:
CI failed due to this issue: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/PR-705/21/tests

vmx:
For comparison. With this pull request, this test didn't fail: https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/PR-711/4/tests

Issue #713, opened 2018-03-12
Getting CI green


vmx:
This is a meta issue to collect all the PRs to get the CI Green

 - [x] https://github.com/ipfs/js-ipfs-api/pull/711
 - [x] https://github.com/ipfs/interface-ipfs-core/pull/235
 - [x] https://github.com/ipfs/interface-ipfs-core/pull/236
 - [ ] Fix of https://github.com/ipfs/interface-ipfs-core/pull/227#issuecomment-372696424
 - [ ] https://github.com/ipfs/jenkins/issues/95
 - [ ] Fix of https://github.com/ipfs/js-ipfs-api/issues/715
 - [x] https://github.com/ipfs/interface-ipfs-core/pull/238
 - [x] https://github.com/ipfs/js-ipfs-api/pull/716
 - [x] https://github.com/ipfs/interface-ipfs-core/pull/239

diasdavid:
@vmx double check, but I think you also need https://github.com/ipfs/js-ipfs-api/pull/695

vmx:
@diasdavid So far it doesn't seem that I need #695, except if it fixes https://github.com/ipfs/interface-ipfs-core/pull/227 (i haven't tried yet)

vmx:
With the outstanding PRs merged and these three reverted:

 - https://github.com/ipfs/js-ipfs-api/commit/a10e3d57f0d71e934faa26ba63d371edcd86df92
 - https://github.com/ipfs/js-ipfs-api/commit/b08f21a8dea1852064518d045cf17706bf90b821
 - https://github.com/ipfs/interface-ipfs-core/commit/de72184d45aab5153818d96e6d0d3b38ea8b6076

CI is quite often green. It's still not reliable, but as different tests are failing I'd consider them environment issues.

vmx:
@diasdavid Everything from my side is merged now. The test runs are quite stable now. Now only the tests broken by the commits mentioned above need to be fixed. Who will fix those? The authors of those commits?

diasdavid:
@vmx would you have the chance to fix those too? thanks!

vmx:
@diasdavid Not sure when, but I can have a look if it isn't fixed in the meantime.

diasdavid:
@vmx I went through it. the files.stat was failing because it was expecting features that haven't been released on go-ipfs. ALl good now :)

vmx:
[...]

PR #717, opened 2018-03-15
Only hash

diasdavid:


diasdavid:
Tested things locally, outstanding issues are related to https://github.com/ipfs/js-ipfs-api/issues/713

JonKrone:
@disdavid Woohoo! Thanks a ton for the merge! 🎆 I'll go ahead and update any related work.

src/utils/send-files-stream.js
```diff
@@ -75,23 +75,10 @@ module.exports = (send, path) => {
...
-
-    if (options.hash != null) {
-      qs.hash = options.hash
-    } else if (options.hashAlg != null) {
-      qs.hash = options.hashAlg
-    }
+    qs['cid-version'] = propOrProp(options, 'cid-version', 'cidVersion')
+    qs['raw-leaves'] = propOrProp(options, 'raw-leaves', 'rawLeaves')
+    qs['only-hash'] = propOrProp(options, 'only-hash', 'onlyHash')
+    qs.hash = propOrProp(options, 'hash', 'hashAlg')
```
AtkinsChang: Please don't direct assign to `qs` which is a options that user pass to api.
This will cause error if user's options is read only

AtkinsChang: Please don't directly assign `qs` which is a function parameters in API,
, this will cause error if user pass read only object as option.

JonKrone: Thanks @AtkinsChang. Did this break something that used to work? We've been assigning to `qs` like this for a few months now so perhaps it is not just the assignment if you've used any of those versions.

> which is a function parameters in API

I'm sorry, I don't get what you mean here. Can you give a clear example?

AtkinsChang: For example, in `src/files/add.js`:v`options` is passed  directly to `createAddStream({ qs: options })`.
In `src/utils/send-files-stream`:
line 34 `options = options ? Object.assign({}, options, options.qs) : {}`
line 76 `const qs = options.qs || {}`
So basically the qs is the options user pass to the api.

JonKrone: Because of line 34 like you mention, it looks like `qs` here is actually a copy of the options passed. I think that this avoids problems with a read-only options object. For example, the below doesn't cause any problems:

```
const obj = {}
Object.defineProperty(obj, 'color', {
  writable: false,
  value: 'red'
})
const options = Object.assign({}, obj)
options.color= 'rouge'
console.log(options.color) // prints 'rouge'
```

Did this break something that worked for you before?

AtkinsChang: Yes this is the currently workaround I use.
But I think the internal ipfs api should use a copy of user's option instead of modifying it directly. It is a good practice to avoid reassigning function arguments, which may cause user hard to debug in some situation, for example: shared options between different apis.

JonKrone: @AtkinsChang ipfs-api internally **does** use a copy of the user's options because of [line 34](https://github.com/ipfs/js-ipfs-api/pull/717/files#diff-81838fe306a1416094e9e2d3e33b5db6R34) and does not mutate the options you pass in. Given that, I'm not sure this is the source of the problem you were having.

If you can reproduce it in some code please let me know or submit a PR with your proposed changes.

PR #721, opened 2018-03-21
Allow files.cat to accept a cid instance

jacobheun:
Implements the spec addition at ipfs/interface-ipfs-core#243, allowing users to use the cid instance itself to cat files.

This has been tested against the added tests from ipfs/interface-ipfs-core#243 via `npm link`

connects ipfs/js-ipfs#1229


diasdavid:
image
@VictorBjelkholm we really need to get Mac OS X stable. It keeps failing randomly for some reason.

PR #723, opened 2018-03-24
feat: streamable ping and optional packet number

JGAntunes:
This should close #556 and add two additional methods to stream the ping results.


diasdavid:
Thank you so much @JGAntunes <3 

image

One of the Mac OS runs didn't really go through but that is another issue.

PR #729, opened 2018-04-02
feat: add awareness of wrapWithDirectory option to utils/send-files-stream

wraithgar:
For https://github.com/ipfs/js-ipfs/issues/1178

Adds awareness of new wrapWithDirectory parameter which will be passed along to the api via querystring.


diasdavid:
merging into a repo branch so that I can update deps

PR #730, opened 2018-04-03
Wrap with dir

diasdavid:
from @wraithgar. just adding latest interface-ipfs-core


diasdavid:
@wraithgar seems that tests on wrap-with-directory are not passing here:

image

wraithgar:
That test can't pass without the code in https://github.com/ipfs/js-ipfs/pull/1290

diasdavid:
@wraithgar the js-ipfs-api runs its tests against go-ipfs. It doesn't use js-ipfs at all. It means that the client library is not interacting with go-ipfs properly which might mean there is some mismatch on how the http endpoint in js-ipfs is working from go-ipfs.

wraithgar:
Ah ok I see. The test that's failing is the new one added in interface-ipfs-core which passes the new `wrapWithDirectory` parameter. So a go equivalent to that js-ipfs PR needs to exist first then right?

diasdavid:
This feature already exists on go-ipfs, it seems that the param for the endpoint was just not implemented in the same way.

image

See also the http-api docs https://ipfs.io/docs/api/#apiv0add

diasdavid:
@wraithgar this is currently blocking me from release js-ipfs-api given that interface-ipfs-core shipped with the test before this was complete (next time we need to run all the tests and make sure it all passes).

I'm just going to tackle it quick to unblock other work.

PR #731, opened 2018-04-03
feat: public readonly method for getting host and port

### LINKED ISSUE #580: Public readonly method to get host and port from API instance
Currently, if you have an instance of the ipfs-api object, it seems that there's no good way to tell what the configured port and hostname is. It'd be nice to expose these as read-only getters on the API object instance.

vasco-santos:
Added a util getter for obtaining the host and port.

Closes #580


diasdavid:
Will wait for CI to merge

PR #734, opened 2018-04-10
feat: Add offset/length arguments to files.cat

achingbrain:
Adds `offset/length` args to ipfs-api.

See:

- ipfs/js-ipfs#1231
- ipfs/interface-ipfs-core#242
- ipfs/js-ipfs-unixfs-engine#209


diasdavid:
@achingbrain can you check on why CI is failing terribly ?

achingbrain:
@diasdavid It's looking slightly healthier now but will need the fixes from ipfs/js-ipfs-api#746

I encountered some problems with duplicated dependencies while working on multiple libraries simultaneously and `~` and `^` are used interchangeably.  Do you think we could standardise on one or the other across all js-ipfs projects?  Better yet, can we release v1 of all modules that are currently 0.x?

diasdavid:
> I encountered some problems with duplicated dependencies while working on multiple libraries simultaneously and ~ and ^ are used interchangeably. Do you think we could standardise on one or the other across all js-ipfs projects? Better yet, can we release v1 of all modules that are currently 0.x?

Our rule is: Use ~ for everything below 1.0.0 and ^ for everything above 1.0.0. If you find a package.json that is not following this rule, please submit a PR.

As for v1 of all the modules. We've discussed it in the past and agreed that until IPFS reaches 1.0.0, we won't be releasing 1.0.0 modules.

diasdavid:
Many things -- https://github.com/ipfs/ipfs/blob/master/ROADMAP-TO-1.0.0.md --, the first questions should be, what is missing to reach Beta, see https://github.com/ipfs/ipfs/issues/314

Issue #724, opened 2018-03-24
Ability to Modify Default Headers


OR13:
As far as I am aware, best practice for securing IPFS is to use an api gateway or proxy to restrict access to the ipfs api.

https://github.com/ipfs/go-ipfs/issues/1532

Bearer Authentication is common, here are some more details on the practice:

https://swagger.io/docs/specification/authentication/bearer-authentication/

I have tested using Kong to secure IPFS in this way and had success, but I find myself wanting to add a header to every network request made by js-ipfs-api.

We could support adding a header to the config:

```
const ipfsConfig = {
  host: 'localhost',
  port: 5001,
  protocol: 'https',
  headers: {
    authorization: 'Bearer ' + ACCESS_TOKEN
  }
};
```

Here where the user agent is added:

https://github.com/ipfs/js-ipfs-api/blob/master/src/utils/send-request.js#L111

We could add the authorization header to every request.

Axios http client supports this:

https://github.com/axios/axios/issues/209

The api gateway (kong/ tyk) would be responsible for validating the jwt token.

I'm happy to submit a PR for this work, but I'm not sure the potential security implications for the rest of the api.

At a minimum it would seem wise to only allow the authorization header (assuming it is formatted correctly).

Interested to hear if this would be useful for others.

wraithgar:
(Assuming jwt is the bearer schema being used)

Since the API is still on `hapi@16` this would be implemented in the server by [`hapi-auth-jwt@7`](https://github.com/dwyl/hapi-auth-jwt2) (version 8 is when it moved to `hapi@17`.

If the API were to move to `hapi@17` we would probably want to use [`@now-ims/hapi-now-auth`](https://github.com/now-ims/hapi-now-auth)

prettymuchbryce:
I think this would be appropriate as I find myself in the same situation as @OR13 . Perhaps as a more general solution would be to allow configuration of custom headers to be sent with any request. This could then extend beyond authentication and a more tailored authentication solution could be implemented if/when the API moves to `hapi@17`.

Thoughts?

OR13:
Here is a PR which should support custom headers for all requests made by the library.

https://github.com/ipfs/js-ipfs-api/pull/741

eolszewski:
@diasdavid this pull request allows IPFS consumers to use Bearer Authentication with a half-line of code. 

This will allow us and others to remove our forks and stay up to date with the progress y'all are making. Can you please merge this?

[...]

PR #740, opened 2018-04-14
Add authorization header config option

prettymuchbryce:
One potential solution for #724, allowing for clients to optionally send an `Authorization` header. This would allow an API gateway to authenticate each request before proxying it to an IPFS daemon.


lgierth:
> @lgierth any comments here?

Nope I think this probably makes sense. One could use it e.g. to authenticate the pin endpoint.

prettymuchbryce:
Closing in favor of #741

PR #749, opened 2018-04-25
fix: increase timeout for name.publish and fix setup code

alanshaw:
`name.publish` takes forever! I've increased the timeout to 5 minutes to accommodate.

I also moved the test (which adds a file for future tests) into the `before` block, I've removed the assertions that this test was making (there should be adequate tests for adding files elsewhere) and instead of hard coding the CID I've extracted it from the return value of the `files.add` call.

### DIFF
--- a/test/name.spec.js
+++ b/test/name.spec.js
@@ -21,6 +21,7 @@ describe('.name', () => {
   let other
   let otherd
   let name
+  let testFileCid
 
   before(function (done) {
     this.timeout(20 * 1000)
@@ -48,6 +49,13 @@ describe('.name', () => {
           const ma = id.addresses[0]
           other.swarm.connect(ma, cb)
         })
+      },
+      (cb) => {
+        ipfs.files.add(testfile, (err, res) => {
+          expect(err).to.not.exist()
+          testFileCid = res[0].hash
+          cb()
+        })
       }
     ], done)
   })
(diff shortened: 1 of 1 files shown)


VictorBjelkholm:
Could we instead of raising the timeout, make sure the daemon is started in offline mode for this test? In offline mode, the operation takes about 0.01 seconds, and gives the same output as-if it was online.

VictorBjelkholm:
> I'm guessing the name.resolve test wouldn't work if we did this?

It would, as long as we're using the same daemon.

alanshaw:
hmm, interesting bug, `ipfs.name.publish` after `ipfs.id` in offline mode causes error:

```
$ ipfs daemon --offline &
[1] 47216
$ Initializing daemon...
Swarm not listening, running in offline mode.
API server listening on /ip4/127.0.0.1/tcp/5001
Gateway (readonly) server listening on /ip4/127.0.0.1/tcp/8090
Daemon is ready
$ ipfs id
{
	"ID": "QmahispzgXbQ5ewaxSjdrst8KNLzREhVHdhxZwAUHQWqLb",
	"PublicKey": "CAASpgIwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDC4Q6B6burDPTlfIkjlF72Q/G72GT7CDBzlZsGI9XKf3XCC5PQiTBcpvKEnUCoBvquiClWE1yJiM4bu1dH/YR5NDBlx9EQzDDmiVvS909cqokRHxA0NY5t83hs8gthoMt6wn15euG2e6LVohqefq6YiiFWgakxDcRyMclLWKiF2t7U6OhMMSd2YdHp15wcHac13ZmdNKhZ/TBhjeKowl4DjAixxLE0kmcK5MsABeQEA+BHukz425K3929onEriSkG7xkSrd0YwKOx6J2Apub7Df8yhcG1PFoTyfTUD/gxkYNrB2jjqUJaNXrK4QDbaDinBNS3KZUn4SQtmxzj2gi4PAgMBAAE=",
	"Addresses": null,
	"AgentVersion": "go-ipfs/0.4.14/5db3846",
	"ProtocolVersion": "ipfs/0.1.0"
}
$ ipfs name publish zdpuArHMUAYi3VtD3f7iSkXxYK9xo687SoNf5stAQNCMzd77k
Error: private key already loaded
```

VictorBjelkholm:
I'm really not a fan of having us use live IPNS in tests when it's as slow as it is. Can we investigate and see if providing the `--local` flag can work instead of `--offline` for now?

diasdavid:
Note it is not "live" as in "live network", the tests spin a mini IPFS network of 3 nodes.

PR #751, opened 2018-04-30
get it ready for release

diasdavid:
Blocked on:
- https://github.com/jbenet/node-subcomandante/issues/8 to get Circle Green


diasdavid:
The compression-level test is now borked as the stream never starts flowing. Investigating. @vmx @pgte any clues here? (I need to hop off now, will be back in a bit)

PR #755, opened 2018-05-03
[WIP] feat: add IPLD to dag.get

### LINKED ISSUE #738: Use js-ipld for DAG API
Basic support for the DAG API landed with https://github.com/ipfs/js-ipfs-api/pull/534.

Though not all formats are supported and you can't traverse through different types of formats. I.e. you couldn't follow a link from a dag-cbor node to a dag-pb node.

The solution is to use js-ipld instead of hard-coding the formats (currently dag-cbor and dag-pb).

Using js-ipld is not straight forward. The problem is that js-ipld needs a Block Service which in turn needs a Repository to work. But js-htt-api doesn't have the blocks available in the JS environment, but receives them manually over the HTTP API.

So I guess you would need a Block Service (or Repository) that uses the HTTP API to retrieve the blocks. But I leave the details to whoever will implement this.

### LINKED ISSUE #747: dag.get for cid with codec not dag-pb or dag-cbor never calls callback
https://github.com/ipfs/js-ipfs-api/blob/563baa3c2204e266755a70c22978e01ac4bd6d41/src/dag/get.js#L42-L49

I have a `git-raw` CID. What should happen? Callback with error? Related to #738?

alanshaw:
This PR is a proposal to resolve #738 and will also fix #747.

In the `dag.get` function we create an temporary in-memory IPLD resolver with a custom bitswap exchange hooked up to the `block.get` API call.

This is kinda awesome!

I've not implemented all the methods for the exchange, but you don't need to because it only uses `get`.

I created a utility module for extracting the CID and path from the `cid` arg...the spec for `dag.get` is flexible in that you can pass it a `Qmhash`, or a `Qmhash/with/path` or a CID instance but the IPLD module isn't as flexible and they need to be split into two args: `CID(Qmhash)` and `/with/path`.

I've also added `explain-error` because it's really good at adding context to error messages whilst also retaining the original error & stack. It also means that in tests we can assert on error messages that are generated within this module rather than (possibly) randomly changing messages from dependencies.

---

**BLOCKED**

This is currently a breaking change to the way dag-pb nodes are traversed via path.

In go-ipfs you can do this:

```
ipfs dag get /ipfs/QmWgCbdDVhsXA8UzpaQwFgw71iGaLrAfyyy51XWpHzNu8C/gif
```

but with this pull request, to get the same object, you'd need to:

```
jsipfs dag get /ipfs/QmWgCbdDVhsXA8UzpaQwFgw71iGaLrAfyyy51XWpHzNu8C/Links/gif
```

Note "Links".


magik6k:
Note that go-ipfs intends to add `/ipld` namespace which will behave the same way as `/ipfs` does here. See https://github.com/ipfs/go-ipfs/pull/4672.

I'd also use `ipfs dag relolve` to resolve the path on the daemon side to limit api cals.

alanshaw:
@magik6k thanks for looking at this!

I'd love to use `dag.resolve` but `js-ipfs` daemon doesn't have it. Is `dag.resolve` going away in `go-ipfs` or is it worth implementing it in `js-ipfs` and adding it to the spec?

diasdavid:
dag.resolve should not exist as it stands per all the conversations I, @whyrusleeping and @jbenet had in the past, both in person and async. It has been almost 2 years since the work on this API started, not worth going through all the data points, the tl;dr; is that there wasn't the opportunity to commit to an API for both implementations. 

I took a lot of notes along the way here, find them at:

- https://github.com/ipfs/interface-ipfs-core/issues/81
- https://github.com/ipfs/interface-ipfs-core/issues/121
- https://github.com/ipfs/interface-ipfs-core/issues/125

My suggestion is for everyone to get a revisited dag API and implement it for both implementations that respects all the constraints (i.e. we can't just use JSON serialization for responses over HTTP as it messes up with types).

test/utils-ipfs-path.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  it('should throw on invalid CID in buffer', () => {
+    const input = Buffer.from('notaCID')
+    expect(() => ipfsPath(input)).to.throw('invalid CID')
+  })
+
+  it('should throw on invalid path', () => {
+    const input = 42
+    expect(() => ipfsPath(input)).to.throw('invalid path')
+  })
+})
```
vmx: Additional test cases:

 - what happens if a path has two slashes in a row (`/ipfs/CID/some//thing`)
 - ipfs namespace without a path (`/ipfs/CID`)
 - CID-only with a trailing slash, with and without namespace (`CID/`, `/ipfs/CID/`)

src/dag/get.js
```diff
@@ -22,31 +24,20 @@ module.exports = (send) => {
...
-          dagCBOR.resolver.resolve(ipfsBlock.data, path, cb)
-        }
-        if (ipfsBlock.cid.codec === 'dag-pb') {
-          dagPB.resolver.resolve(ipfsBlock.data, path, cb)
-        }
-      }
-    ], callback)
+    IPLDResolver.inMemory((err, ipld) => {
+      if (err) return callback(explain(err, 'failed to create IPLD resolver'))
+      ipld.support.rm(dagPB.resolver.multicodec)
```
vmx: Can you please add a comment here, why this is needed? It's clear to me, but I don't think it's clear when you read the code for the first time.

alanshaw:
unblocked thanks to https://github.com/ipld/js-ipld-dag-pb/pull/78

src/dag/get.js
```diff
@@ -22,31 +24,20 @@ module.exports = (send) => {
...
-      (ipfsBlock, path, cb) => {
-        if (ipfsBlock.cid.codec === 'dag-cbor') {
-          dagCBOR.resolver.resolve(ipfsBlock.data, path, cb)
-        }
-        if (ipfsBlock.cid.codec === 'dag-pb') {
-          dagPB.resolver.resolve(ipfsBlock.data, path, cb)
-        }
-      }
-    ], callback)
+    IPLDResolver.inMemory((err, ipld) => {
```
diasdavid: I've been thinking that this should be done by passing the `ipfs.block.{get,put}` as the Block Service, rather than doing an inMemory resolver and then monkey patching the Block Service Instance with a fake exchange.

You should be able to do:

```
const ipld = new IPLD(ipfs.block)
```

or close

```
const ipld = new IPLD(shimBS(ipfs.block))
```

PR #760, opened 2018-05-09
fix: make pubsub.unsubscribe async and alter pubsub.subscribe signature

alanshaw:
depends on: https://github.com/ipfs/interface-ipfs-core/pull/260


diasdavid:
image

Windows tests are having a hard time. Being fixed at #756

PR #761, opened 2018-05-13
feat: Bitswap wantlist peerid

wraithgar:
Adds peerid parameter to bitswap wantlist. Required for https://github.com/ipfs/js-ipfs/pull/1349


src/bitswap/unwant.js
```diff
@@ -1,13 +1,33 @@
...
+      if (CID.isCID(args)) {
+        cid = args
+        args = cid.toBaseEncodedString()
+      } else if (Buffer.isBuffer(args)) {
+        cid = new CID(args)
+        args = cid.toBaseEncodedString()
+      } else if (typeof args === 'string') {
+        cid = new CID(args)
+      } else {
+        return callback(new Error('invalid argument'))
```
alanshaw: It's best not to call a callback in a try block. If the callback throws for whatever reason then it'll be caught and then the callback will be called again....which is really hard to debug.

You can just `throw` here, it'll be caught and the callback will be called.

src/bitswap/wantlist.js
```diff
@@ -1,11 +1,38 @@
...
+          cid = new CID(args)
+          args = cid.toBaseEncodedString()
+        } else if (typeof args === 'string') {
+          cid = new CID(args)
+        } else {
+          return callback(new Error('invalid argument'))
+        }
+      } catch (err) {
+        return callback(err)
+      }
```
alanshaw: Please see comments in `unwant.js` about this try block

src/bitswap/unwant.js
```diff
@@ -1,13 +1,33 @@
...
+      }
+    } catch (err) {
+      return callback(err)
+    }
```
alanshaw: I'm just wondering if there's a reason for the type detection here? It's a duplication of what's [done in the CID constructor](https://github.com/ipld/js-cid/blob/4d7bcf860214fef04700ebe338908d36182c931e/src/index.js#L56-L88). Would it be possible to do this instead?:

```js
let cid

try {
  cid = new CID(args)
} catch (err) {
  return callback(err)
}

send({
  path: 'bitswap/unwant',
  args: cid.toBaseEncodedString(),
  qs: opts
}, callback)
```

alanshaw:
Windows test failures are unrelated to this PR and are known swarm connection issues on Windows. I am investigating and am close to figuring out what the issue is.

PR #762, opened 2018-05-14
fix(ping): convert the ping messages to lowercase

JGAntunes:
See https://github.com/ipfs/js-ipfs/pull/1342#discussion_r186710133


JGAntunes:
@alanshaw about the ping spec to ipfs-core, it's one thing missing yes, I was considering opening an issue on interface-ipfs-core just to track it and work on it afterwards if that's ok.

alanshaw:
@JGAntunes see https://github.com/ipfs/js-ipfs/pull/1342#issuecomment-388813381

PR #767, opened 2018-05-16
chore: Add api path docs and test

### LINKED ISSUE #581: ipfs behind proxy
I am running my IPFS behind an ngnix proxy and redirected port 5001 of IPFS to` http://<IPAddress>/ipfs`

We are developing a Web Interface from which we want to access `http://<IPAddress>/ipfs`, however js-ipfs-api supports ipfsAPI which seems to be accepting only IP address / port as shown below
var ipfs = ipfsAPI(`'ip4/<IPAddress>/tcp/5001'`, {protocol: 'http'})

Is there any option that would allow me to connect to `http://<IPAddress>/ipfs` using ipfsAPI?

vasco-santos:
After @alanshaw [suggestion](https://github.com/ipfs/js-ipfs-api/pull/744), I added docs and a test for the `api-path` usage in the API constructor.

Closes #581

### DIFF
--- a/test/constructor.spec.js
+++ b/test/constructor.spec.js
@@ -56,6 +56,16 @@ describe('ipfs-api constructor tests', () => {
       clientWorks(ipfsAPI(splitted[2], splitted[4]), done)
     })
 
+    it('specify host, port and api path', (done) => {
+      const splitted = apiAddr.split('/')
+
+      clientWorks(ipfsAPI({
+        host: splitted[2],
+        port: splitted[4],
+        'api-path': '/api/v0/'
+      }), done)
+    })
+
     it('host, port, opts', (done) => {
       const splitted = apiAddr.split('/')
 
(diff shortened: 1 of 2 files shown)


test/constructor.spec.js
```diff
@@ -56,6 +56,16 @@ describe('ipfs-api constructor tests', () => {
       clientWorks(ipfsAPI(splitted[2], splitted[4]), done)
     })
 
+    it('specify host, port and api path', (done) => {
+      const splitted = apiAddr.split('/')
+
+      clientWorks(ipfsAPI({
+        host: splitted[2],
+        port: splitted[4],
+        'api-path': '/api/v0/'
```
alanshaw: Since this is the default, this test doesn't really prove anything except for you can pass `api-path` as an option without it breaking.

Ideally we'd actually setup a proxy for the test and point the API at the proxy with `api-path: '/ipfs/api/v0'`. Obviously this wouldn't be possible in the browser so you'd need to skip.

I'd accept a PR for that!

PR #772, opened 2018-05-17
fix(pubsub): callback from unsubscribe after stream ends

alanshaw:
This PR changes the `unsubscribe` logic to only callback on the last event we can listen for on a dropped connection.

This **still** does not guarantee that your topic won't appear when calling `pubsub.ls` in the callback to `pubsub.unsubscribe` - **but only when talking to go-ipfs**.

Adding an artificial timeout using `setTimeout` _seems_ to provide enough time for go-ipfs to actually do the unsubscribe.

### DIFF
--- a/src/pubsub.js
+++ b/src/pubsub.js
@@ -81,8 +81,29 @@ module.exports = (arg) => {
 
       // Drop the request once we are actually done
       if (ps.listenerCount(topic) === 0) {
-        subscriptions[topic].abort()
+        if (!callback) {
+          return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


diasdavid:
Now seems that a new fail on DHT is happening:

```
  1) .dht

       "before all" hook:

     Error: Uncaught AssertionError: expected [Error: Unexpected end of JSON input] to not exist (webpack:///node_modules/chai/lib/chai/assertion.js:141:0 <- node_modules/aegir/src/config/karma-webpack-bundle.js:93052)
```

PR #776, opened 2018-05-23
fix: files ls inconsisntency

hacdias:


src/files/ls.js
```diff
@@ -8,7 +8,7 @@ const transform = function (res, callback) {
   callback(null, entries.map((entry) => {
     return {
       name: entry.Name,
-      type: entry.Type,
+      type: (entry.Type === 0) ? 'file' : 'directory',
```
travisperson: The unixfs type for `file` is 2, and `directory` is 1.

Sources: 
[0] https://github.com/ipfs/js-ipfs-unixfs/blob/master/src/unixfs.proto.js#L6,L7
[1] https://github.com/ipfs/js-ipfs/blob/master/src/http/api/resources/files.js#L323,L332

Also per source [1], it appears that `js-ipfs-unixfs-engine` uses `dir`[2] instead of `directory`. However, `js-ipfs-unixfs` defines the string type as `directory`[3].

[2] https://github.com/ipfs/js-ipfs-unixfs-engine/blob/master/src/exporter/dir-flat.js#L32
[3] https://github.com/ipfs/js-ipfs-unixfs/blob/master/src/index.js#L11

I'm unsure if this is going to cause a conflict immediately though, but it looks like `js-ipfs-unixfs-engine` should be updated as well at some point.

hacdias: > The unixfs type for file is 2, and directory is 1.

That's strange, because I'm getting 0s for files and 1s for directories... Hmmm, well, I'll take a look at this later then. 😄

alanshaw: What are the numbers in go-ipfs?

hacdias:
@alanshaw what do you think about this?

hacdias:
@alanshaw what's the state of this PR?

alanshaw:
@hacdias did you see the comments here https://github.com/ipfs/interface-js-ipfs-core/pull/282#discussion_r232642237 ?

hacdias:
@alanshaw I believe I missed it!! Just read the comments. I'll work on those points. I'll close this PR as soon as I open a new one since this is a bit out of date.

PR #782, opened 2018-05-30
test: dht testing

diasdavid:
The dht.findprovs endpoint is inconsistent with the other endpoints when reporting errors. It sends a message like:

```
argument "key" is required
```

While others wrap it in a JSON object with more info:

```
{"Message":"Invalid base58 digit ('-')","Code":0,"Type":"error"}
```

This breaks our parser because it is expecting that the messages are standardized.

I'm testing these with the https://github.com/ipfs/interface-ipfs-core/pull/288

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,8 @@
     "coverage-publish": "aegir coverage --provider coveralls --timeout 100000"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


diasdavid:
Failing tests are from before (PubSub and Windows)

PR #784, opened 2018-06-06
fixed json-loader error

leekt216:
while checking out upload-file-via-browser example,
i encountered "cannot resolve module json-loader" error
it was simply solved by updating webpack from 1.9.6-> 2.0.0
since json works on default since webpack 2.0.0 https://github.com/webpack-contrib/json-loader
react-hot & babel loader had to be specified( add '-loader')


alanshaw:
Thanks @leekt216! Could you please remove the DS_Store files from your PR and also remove the `json-loader` from `package.json` dependencies since we don't need it anymore. Then we can merge!

leekt216:
Thanks @alanshaw i removed DS_Store and json-loader

PR #785, opened 2018-06-08
feat: modular interface tests

alanshaw:
This PR updates js-ipfs-api to use the new interface-ipfs-core tests. This gives us fine grained control over what tests we run.

Please see the [description on the PR](https://github.com/ipfs/interface-ipfs-core/pull/290) for more information on the reasons behind this change.

This also enables a bunch of tests that weren't previously being run 🎉 ...and removes a LOT of duplicated code for running the interface tests.

**requires https://github.com/ipfs/interface-ipfs-core/pull/290**


test/interface.spec.js
```diff
@@ -0,0 +1,179 @@
...
+      'should get a dag-pb node local value',
+      'should get dag-pb value via dag-cbor node',
+      'should get by CID string + path',
+      // dag.put
+      // FIXME This works in go-ipfs because dag-pb will serialize any object. If
+      // the object has neither a `data` nor `links` field it's serialized
+      // as an empty object
+      'should not put dag-cbor node with wrong multicodec'
+    ]
+  })
```
diasdavid: Hm... is this really the preferred approach? I foresee people changing the test title and not understanding why things are failing. I kind of liked when the tests explicitly communicated that it was known it wasn't working with go-ipfs or js-ipfs.

test/interface.spec.js
```diff
@@ -0,0 +1,179 @@
...
+  tests.object(defaultCommonFactory)
+
+  tests.pin(defaultCommonFactory)
+
+  tests.ping(defaultCommonFactory)
+
+  tests.pubsub(CommonFactory.create({
+    spawnOptions: {
+      args: ['--enable-pubsub-experiment'],
+      initOptions: { bits: 1024 }
```
diasdavid: Why not use `1024` for all tests? It would make things way faster.

test/interface.spec.js
```diff
@@ -0,0 +1,179 @@
...
+              nodes.push(_ipfsd)
+              cb(null, IPFSApi(_ipfsd.apiAddr))
+            })
+          }
+        })
+      }
+    }
+  }))
+
+  tests.types(defaultCommonFactory, { skip: true })
```
diasdavid: Why skip?

test/interface.spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+  tests.util(defaultCommonFactory, { skip: true })
```
diasdavid: Why skip?

test/interface.spec.js
alanshaw: These were not being run before and they fail right now - I'll add a comment

test/interface.spec.js
alanshaw: These were not being run before and they fail right now - I'll add a comment

test/interface.spec.js
alanshaw: They still are, the `defaultCommonFactory` passed to other suites uses 1024 by default https://github.com/ipfs/js-ipfs-api/pull/785/files/519bbf5c4374317d5355d9bddfb2fc5862ddd16d#diff-b4ff00e5aeab8b441e96f3d0b1082a2eR12

package.json
```diff
@@ -75,6 +74,7 @@
     "browser-process-platform": "~0.1.1",
     "chai": "^4.1.2",
     "cross-env": "^5.1.6",
+    "detect-node": "^2.0.3",
```
alanshaw: Note to self: this needs to be moved back into dependencies before merging as it's used in the main code

test/interface.spec.js
alanshaw: Also worth mentioning - we already have a scattering of where this information is held. We have implementation specific skips in interface-ipfs-core (go/js), but platform dependent (browser/node.js) skips in js-ipfs.

For example, some suites don't work in the browser so aren't included https://github.com/ipfs/js-ipfs/blob/2e9418465c863b129300ea5f0c8b32590edc7729/test/core/interface/interface.spec.js#L17-L23 and we have the same thing for functionality that isn't implemented e.g. repo tests simply just not included in js-ipfs because it's not all implemented yet.

These PRs would put all of that information in the same place, which I think is a win.

I'd like to at least try things this way around and if it becomes an issue then we change it back?

What do you think @diasdavid - is worth a try? I'd like to move forwards on these PRs soon because they're becoming difficult to keep up to date with master!

Thank you, much ❤️

diasdavid: I would say Option 1 given the example but since some test titles are actually long, Option 2 might serve us better.

alanshaw: @diasdavid is it nice?:

**skip specific test**

screen shot 2018-06-26 at 23 08 11

**skip entire subsystem**

screen shot 2018-06-26 at 23 07 26

test/interface.spec.js
diasdavid: Understood! Thanks :)

test/interface.spec.js
```diff
@@ -0,0 +1,235 @@
...
+
+describe('interface-ipfs-core tests', () => {
+  const defaultCommonFactory = CommonFactory.create()
+
+  tests.bitswap(defaultCommonFactory, {
+    skip: [
+      // bitswap.unwant
+      {
+        name: 'should remove a key from the wantlist',
+        reason: 'FIXME why is this skipped?'
```
diasdavid: @alanshaw does it fail when run?

alanshaw: Yes unfortunately. It was just skipped in interface-ipfs-core before with no explanation.

```
  1) interface-ipfs-core tests
       .bitswap.unwant
         should remove a key from the wantlist:

      Uncaught AssertionError: expected false to equal true
      + expected - actual

      -false
      +true
      
      at ipfsB.bitswap.wantlist (/Users/alan/Code/protocol-labs/interface-ipfs-core/js/src/bitswap/unwant.js:52:61)
      at f (node_modules/once/once.js:25:25)
      at streamToValue (src/utils/stream-to-json-value.js:30:5)
      at concat (src/utils/stream-to-value.js:12:22)
      at ConcatStream.<anonymous> (node_modules/concat-stream/index.js:37:43)
      at finishMaybe (node_modules/readable-stream/lib/_stream_writable.js:630:14)
      at afterWrite (node_modules/readable-stream/lib/_stream_writable.js:492:3)
      at process._tickCallback (internal/process/next_tick.js:63:19)
```

diasdavid:
y no green CI?

alanshaw:
@diasdavid I had this all lined up to be merged asap after `interface-ipfs-core@0.70` was released.

After 0.70 was released I saw that there were many issues on CI that didn't appear at all when run locally.

Jenkins had a bad day last week coinciding with the release so it's been difficult and slow to debug and resolve the issues. I now have a PR open on `interface-ipfs-core` to resolve them https://github.com/ipfs/interface-ipfs-core/pull/319 (🤞 hopefully all of them). It would be rad if you have time to take a quick look.

I'm super anxious to get this and https://github.com/ipfs/js-ipfs/pull/1389 merged asap because they're both depending on 0.69.x still. It's confusing and frustrating the community when trying to add tests to interface-ipfs-core right now 😞

alanshaw:
FYI @diasdavid the 2 DAG tests that are failing here will be fixed by this PR https://github.com/ipfs/js-ipfs-api/pull/801

alanshaw:
@diasdavid gr gr gr greeeeeeen!

PR #786, opened 2018-06-12
chore: update method signatures in line with ipfs/interface-ipfs-core#277

achingbrain:
See ipfs/interface-ipfs-core#277 for more information

### DIFF
--- a/src/files/mv.js
+++ b/src/files/mv.js
@@ -1,17 +1,19 @@
 'use strict'
 
 const promisify = require('promisify-es6')
+const findSources = require('../utils/find-sources')
 
 module.exports = (send) => {
-  return promisify((args, opts, callback) => {
-    if (typeof opts === 'function' &&
-      callback === undefined) {
-      callback = opts
-      opts = {}
-    }
+  return promisify(function () {
+    const {
+      callback,
+      sources,
+      opts
+    } = findSources(Array.prototype.slice.call(arguments))
+
     send({
       path: 'files/mv',
-      args: args,
+      args: sources,
       qs: opts
     }, callback)
   })
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "flatmap": "0.0.3",
     "glob": "^7.1.2",
     "ipfs-block": "~0.7.1",
-    "ipfs-unixfs": "~0.1.14",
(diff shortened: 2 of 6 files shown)


src/files/mv.js
```diff
@@ -3,12 +3,24 @@
...
+      opts = args.pop()
+    }
+
+    if (args.length === 1 && Array.isArray(args[0])) {
+      // support ipfs.file.mv([src, dest], opts, cb)
+      sources = args[0]
+    } else {
+      // support ipfs.file.mv(src, dest, opts, cb) and ipfs.file.cp(src1, src2, dest, opts, cb)
+      sources = args
     }
```
vasco-santos: Maybe we can have the above logic abstracted in a util function since the code is replicated in `mv` and `cp`. What do you think @achingbrain ?

achingbrain:
@vasco-santos I've updated the PR and implemented your suggestion but I'm having problems getting Jenkins to pass.  Lots of instability around the ping tests.

They seem to be failing for master on Jenkins too, any ideas?

vasco-santos:
Thanks @achingbrain !

I checked the CI log and it was a timeout for windows. So, I started the CI again to check if it gets green again.

achingbrain:
> I checked the CI log and it was a timeout in windows environment. So, I started the CI again to check if it gets green again.

Ha, I've been restarting it all morning.  We should probably quarantine flaky tests until we can make them stable.

vasco-santos:
@achingbrain yeah! It still fails

vasco-santos:
Yeah, I agree that it is the best solution here @achingbrain !

achingbrain:
Great, can this be merged & released?

achingbrain:
...and could there please also be a corresponding release of [`ipfs/js-ipfsd-ctl`](https://github.com/ipfs/js-ipfsd-ctl) with the new `js-ipfs-api` version?

Releasing v1 of this modules would make this all so much easier!

vasco-santos:
For me, it is good to go and be released. 

I can handle the `ipfsd-ctl` part, but we need someone for releasing the API.

achingbrain:
I think everyone's on holiday..  I'll merge this and hopefully someone can do the do when people are back.

PR #789, opened 2018-06-20
Fix for empty data in block.

mikeal:
Found this very strange bug when creating DAGNode's with empty buffers for data.

streamToValue returns an empty array when the data is empty, which down the line triggers an exception in ipfs-block because this sends an empty array instead of a buffer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "eslint-plugin-react": "^7.9.1",
     "go-ipfs-dep": "~0.4.15",
     "gulp": "^3.9.1",
-    "interface-ipfs-core": "~0.68.1",
+    "interface-ipfs-core": "~0.69.0",
     "ipfsd-ctl": "~0.37.3",
     "pull-stream": "^3.6.8",
     "socket.io": "^2.1.1",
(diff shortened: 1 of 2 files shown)


alanshaw:
Thanks @mikeal, do you have a sec to add a test for this to `interface-ipfs-core`?

mikeal:
Hrm... that will take some time. There's a lot of test framework and fixtures to figure out in those tests that I have no experience with. I can do it, but I won't find the time for a while.

alanshaw:
I've amended this to add the same check for when the response doesn't have the `X-Stream-Output` header.

go-ipfs currently sets this (which is why we drop into the `streamToValue` bit - where you've added your fix) but js-ipfs currently doesn't...which is why we also need to do the check for when `res` is a value (not a stream). :P Hope that makes sense.

I added a test for this (https://github.com/ipfs/interface-ipfs-core/pull/308) and verified that it doesn't work before your fix and it does work after. Hoorays \o/

There's also a PR against js-ipfs (https://github.com/ipfs/js-ipfs/pull/1408) to add the `X-Stream-Output` header to `block.get` and fix an issue that disallowed us to `block.put` an empty buffer.

Phew, fun times.

PR #796, opened 2018-06-27
chore: logs path & querystring for requests

achingbrain:
Also maps 'Type' property from error responses

### DIFF
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -9,6 +9,7 @@ const once = require('once')
 const streamToValue = require('./stream-to-value')
 const streamToJsonValue = require('./stream-to-json-value')
 const request = require('./request')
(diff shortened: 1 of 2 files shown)


achingbrain:
Not sure why CI keeps failing, you'd have thought these changes would be fairly innocuous.

achingbrain:
Looks like the browser tests are taking ages to run and eventually the job times out and gets killed.

alanshaw:
> Not sure why CI keeps failing, you'd have thought these changes would be fairly innocuous.

Apparently not! Not that you'd know from CI, but in the browser `res.req` is not available so _all_ requests are broken in 22.2.0. PR here https://github.com/ipfs/js-ipfs-api/pull/798 - please can has quick review?

PR #798, opened 2018-06-29
fix: res.req only in Node.js, in browser use res.url instead

alanshaw:


src/utils/send-request.js
```diff
@@ -37,7 +37,11 @@ function onRes (buffer, cb) {
     const isJson = res.headers['content-type'] &&
                    res.headers['content-type'].indexOf('application/json') === 0
 
-    log(res.req.method, `${res.req.getHeaders().host}${res.req.path}`, res.statusCode, res.statusMessage)
+    if (isNode) {
+      log(res.req.method, `${res.req.getHeaders().host}${res.req.path}`, res.statusCode, res.statusMessage)
+    } else {
+      log(res.url, res.statusCode, res.statusMessage)
```
achingbrain: Is the method available or just the URL?

alanshaw: I looked but did not see the method

achingbrain: Fair enough

achingbrain: It's all `POST` anyway..

alanshaw:
Ok so tests are still failing but for other reasons, they got a whole lot further than they did last time (for the ones that actually got started).

requilence:
@alanshaw @achingbrain When I run last version of ipfs-api inside Electron, `isNode` is true, but `res.req` is undefined. So it may be better to check  `if (res.req)` instead of using **detect-node**?

alanshaw:
Thanks @requilence - I've created an issue to track this https://github.com/ipfs/js-ipfs-api/issues/807 - would you be willing to send a PR to fix this?

PR #801, opened 2018-07-03
fix(dag): ensure `dag.put()` allows for optional options

PascalPrecht:
**Note: Only review last commit as it's rebased on top of #785 to make test runner work**

This is to align with API changes in 

https://github.com/ipfs/interface-ipfs-core/commit/011c417b97e58aceafaa4e98fce8f32217bc5cf7

and

https://github.com/ipfs/js-ipfs/pull/1415

**Temporarily rebased on top of #785**

Note: this is a WIP as I have to run now :D will make tests green once I'm back

### DIFF
--- a/src/dag/put.js
+++ b/src/dag/put.js
@@ -7,57 +7,65 @@ const CID = require('cids')
 const multihash = require('multihashes')
 const SendOneFile = require('../utils/send-one-file')
 
-function noop () {}
-
 module.exports = (send) => {
   const sendOneFile = SendOneFile(send, 'dag/put')
 
   return promisify((dagNode, options, callback) => {
     if (typeof options === 'function') {
       callback = options
-    } else if (options.cid && (options.format || options.hash)) {
+    }
+
+    options = options || {}
+
+    if (options.hash) {
+      options.hashAlg = options.hash
+      delete options.hash
+    }
+
+    if (options.cid && (options.format || options.hashAlg)) {
       return callback(new Error('Can\'t put dag node. Please provide either `cid` OR `format` and `hash` options.'))
-    } else if ((options.format && !options.hash) || (!options.format && options.hash)) {
+    } else if ((options.format && !options.hashAlg) || (!options.format && options.hashAlg)) {
       return callback(new Error('Can\'t put dag node. Please provide `format` AND `hash` options.'))
     }
 
-    callback = callback || noop
+    if (options.cid) {
+      let cid
+
+      try {
+        cid = new CID(options.cid)
+      } catch (err) {
+        return callback(err)
+      }
+
+      options.format = cid.codec
+      options.hashAlg = multihash.decode(cid.multihash).name
+      delete options.cid
+    }
 
     const optionDefaults = {
       format: 'dag-cbor',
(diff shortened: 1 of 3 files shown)


diasdavid:
@alanshaw following your comment on https://github.com/ipfs/js-ipfs-api/pull/785#issuecomment-402197142, I've changed the base of this PR to your branch. If you approve, let's get this one merged there and then merge one clean PR to master.

src/dag/put.js
```diff
@@ -15,29 +14,35 @@ module.exports = (send) => {
 
   return promisify((dagNode, options, callback) => {
     if (typeof options === 'function') {
-      return setImmediate(() => callback(new Error('no options were passed')))
+      callback = options
+    } else if (options.cid && (options.format || options.hash)) {
```
PascalPrecht: Notice that `js-ipfs-api`'s implementation expects `options.hash` instead of `options.hashAlg`. This seems to be inconsistent with the speck as per https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DAG.md#javascript---ipfsdagputdagnode-options-callback

Changing this would be a breaking change. @alanshaw how do you feel about this?

src/dag/put.js
```diff
@@ -15,29 +14,35 @@ module.exports = (send) => {
...
+      return callback(new Error('Can\'t put dag node. Please provide either `cid` OR `format` and `hash` options.'))
+    } else if ((options.format && !options.hash) || (!options.format && options.hash)) {
+      return callback(new Error('Can\'t put dag node. Please provide `format` AND `hash` options.'))
```
PascalPrecht: Turns out this is a breaking change as well (which is also one of the reasons CI fails at this time) Prior to this commit, it was okay to just have `format` with `hash` falling back to `sha2-256`.

Again, this would be a point of either introducing a breaking change and staying consistent with `js-ipfs` API, or leaving as is and live with the consequences of inconsistency. 

@alanshaw your call.

alanshaw:
@PascalPrecht I pushed a fix, I hope you don't mind, this PR is now on the critical path to getting the modular interface tests merged 😉 

I think the problem was that `hashAlg` was being passed to `dag.put` but it was expecting `hash`. The API often allows you to pass the HTTP API querystring name instead of the interface-ipfs-core name for the argument (they differ sometimes) so I've preserved this functionality.

test/util.spec.js
```diff
@@ -160,7 +160,9 @@ describe('.util', () => {
         .then(out => expectTimeout(ipfs.object.get(out[0].hash), 4000))
     })
 
-    it('with wrap-with-directory=true', (done) => {
+    it('with wrap-with-directory=true', function (done) {
```
PascalPrecht: I think the `function` keyword is a leftover here?

alanshaw: I'm using `function` here because on the next time we want to call `this.timeout`. Mocha provides an object with this method on it for the function context (`this`) but arrow functions do not allow you to change the context.

PascalPrecht: Got it. Thanks for clarifying!

PR #802, opened 2018-07-03
chore: update deps and test on CI

diasdavid:


diasdavid:
Oh I see, missed that #785 hasn't been merged yet.

PR #804, opened 2018-07-04
fix(files): ensure CID in `cat()` command is only cleaned when input …

### LINKED ISSUE #799: Error: multihash too short. must be > 3 bytes.
If you submit a request to get a file using ipfs.files.cat, and you have a hash that is greater than 3 bytes, you still get the error message: "Error: multihash too short. must be > 3 bytes."

For example:

`ipfs.files.cat(/ipfs/QmUzdHisA7ouM8DM5rVAKursVK7t7tKeqDYGvmaCdJw5jEeesdlfjsldkfjsldkfjsdlkfjskldfj` will return that error message

PascalPrecht:
…is valid

Fixes #799

License: MIT
Signed-off-by: Pascal Precht pascal.precht@gmail.com


PascalPrecht:
@alanshaw maybe you can leave a quick comment on https://github.com/ipfs/js-ipfs-api/issues/799#issuecomment-402466066

I'll put this PR is shape right away then :)

src/files/cat.js
```diff
@@ -13,7 +18,9 @@ module.exports = (send) => {
     }
 
     try {
-      hash = cleanCID(hash)
+      if (checkValidInput(hash)) {
+        hash = cleanCID(hash)
+      }
```
alanshaw: I think you're right, we should delegate the parameter checking to the IPFS node we're calling so not to duplicate the logic and have to keep two sets of the same code in sync. We just need to serialize valid inputs that aren't already serialized. Could we simply do something like this?:

```js
try {
  hash = typeof hash === 'string' ? hash : cleanCID(hash)
} catch (err) {
  return callback(err)
}
```

PascalPrecht: Sure that could work I'll update the and also look into if and how this can be tested.

PascalPrecht:
@alanshaw this is now updated. A test for verification can be found here https://github.com/ipfs/interface-ipfs-core/pull/330

PR #806, opened 2018-07-04
chore: use data-encoding arg to ensure data is not corrupted

achingbrain:
Requires ipfs/js-ipfs#1420 to be merged in order to allow the test in ipfs/interface-ipfs-core#302 to pass.

### DIFF
--- a/src/object/get.js
+++ b/src/object/get.js
@@ -41,11 +41,16 @@ module.exports = (send) => {
 
     send({
       path: 'object/get',
-      args: cidB58Str
+      args: cidB58Str,
+      qs: {
+        'data-encoding': 'base64'
+      }
     }, (err, result) => {
       if (err) {
         return callback(err)
       }
+
+      result.Data = Buffer.from(result.Data, 'base64')
 
       const links = result.Links.map((l) => {
         return new DAGLink(l.Name, l.Size, Buffer.from(bs58.decode(l.Hash)))


alanshaw:
Do you mean https://github.com/ipfs/js-ipfs/pull/1420 not https://github.com/ipfs/js-ipfs/pull/420 ?

achingbrain:
Yes, have updated the description.

diasdavid:
@achingbrain please rebase master onto this branch to get that CI green :)

achingbrain:
This requires go-ipfs 0.4.17 to be released first - not much point merging this until then.

achingbrain:
go-ipfs 0.4.17 has been released, should probably get this in so we can merge ipfs/js-ipfs#1420 and ipfs/interop#25.  Needs ipfs/interface-ipfs-core#342 merging first it seems.

achingbrain:
@alanshaw CI is green, can it be merged?

achingbrain:
Yes, it requires go 0.4.17.

achingbrain:
Will this stop this from being merged?

diasdavid:
0.4.17 has been released

alanshaw:
@achingbrain would you mind adding a BREAKING CHANGE message to the commit? Also, do you think it's more appropriate to prefix with "fix:" than "chore:"?

This change will mean that the API will only work with go-ipfs 0.4.17 and above and we need to make sure there's no confusion around that when it's released.

achingbrain:
Done, just trying to get Jenkins to pass.

achingbrain:
Hmm, [this new test](https://github.com/ipfs/js-ipfs-api/blob/ff7c7e569ce0749f76bd282b5d16edadbc7e2201/test/dag.spec.js#L73-L81) is failing, but only on Windows and only some of the time.  Any ideas?

achingbrain:
It times out so the repo might be missing the DAG node it's trying to get?  It's not clear to me where that node would come from so I've added it to the repo in the test.

achingbrain:
...aaaand one of the Jenkins workers is out of disk space.

src/object/get.js
```diff
@@ -41,12 +41,17 @@ module.exports = (send) => {
 
     send({
       path: 'object/get',
-      args: cidB58Str
+      args: cidB58Str,
+      qs: {
+        'data-encoding': 'base64'
```
lidel: @alanshaw  quick question: what values are allowed here? 
Ones from https://github.com/multiformats/multibase/blob/master/multibase.csv ?

achingbrain: At the moment `go-ipfs` [supports `text` and `base64`](https://github.com/ipfs/go-ipfs/blob/master/core/commands/object/object.go#L568-L579).

Go's `text` is the default (e.g. raw bytes), though `go-ipfs` doesn't escape the characters so it's http api can produce strings that are invalid in JSON and hard to replicate in JavaScript so `base64` is safer.

Once ipfs/js-ipfs#1420 is in, `js-ipfs` will support whatever [Buffer.toString([encoding[, start[, end]]])](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end) does.

If we want to add all of the different encodings from that csv it should be pretty straightforward once both implementations support the `data-encoding` argument.

lidel: Thanks @achingbrain!

This PR reminded me of [yEnc](https://en.wikipedia.org/wiki/YEnc) from Usenet days, so I was just looking at efficiency column [in encoding standards comparison](https://en.wikipedia.org/wiki/Binary-to-text_encoding#Encoding_standards) and things like [Ascii85](https://en.wikipedia.org/wiki/Ascii85) (used in PDFs and git patch) looked interesting (due to moderate adoption and smaller overhead, eg.  25% instead of 33%).

..but then i remembered that HTTP API is often gzipped which means any gains introduced by more exotic encoding than Base64 go away.

For example: [base-64 seems to compress better than base-122](http://blog.kevinalbs.com/base122#experimental_results) :upside_down_face:

PR #810, opened 2018-07-13
fix: add missing debug dependency

### LINKED ISSUE #809: Error: Cannot find module 'debug'
'debug' package is used at ipfs-api/src/utils/send-request.js:12:13, but is not listed as a dependency.

alanshaw:
fixes #809


diasdavid:
@alanshaw can you rebase master onto this branch so that we keep the good practice of merging only green PRs?

PR #813, opened 2018-07-16
fix: increase name after timeout

alanshaw:
Seeing this timeout on multiple environments and Node.js versions so have given it an extra 5s.

Also increases the `browserNoActivityTimeout` to give the `dht.query` test time to [fail but pass anyway](https://github.com/ipfs/interface-ipfs-core/blob/87a8f96b74987ab6d477dc73a91d9f32ff066267/js/src/dht/query.js#L46-L52) - previously the `browserNoActivityTimeout` wasn't taking into account the extra 60s allocated to the before hook, so I just added it here.

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -17,7 +17,7 @@ module.exports = {
       served: true,
       included: false
     }],
-    browserNoActivityTimeout: 150 * 1000,
+    browserNoActivityTimeout: 210 * 1000,
     singleRun: true
   },
   hooks: {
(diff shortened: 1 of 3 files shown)


alanshaw:
Sorry @diasdavid jenkins seems to be running these tests again...they passed here before I submitted the PR https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/fix%2Fincrease-name-after-timeout/2/pipeline - hopefully they'll pass again!

diasdavid:
go-ipfs PubSub still doesn't love us:

image

alanshaw:
I'm hoping it does now with the upgrade to `interface-ipfs-core`

Issue #790, opened 2018-06-21
dag.tree() is missing :(


mikeal:
Didn't see an issue tracking this yet. It's in the docs but isn't actually implemented yet.

PR #819, opened 2018-07-24
[WIP] add missing dag.tree()

pin3da:
Hi!

I was trying to solve #790 but I decided to stop here because I'm not sure if this is the right path and I decide to ask you what to do next.

Initially, I searched in the [HTTP-API](https://ipfs.io/docs/api/#apiv0dagget) for an endpoint "/dag/tree", but I did not find it. I also ran the go daemon and made some tests but it seems that the endpoint is not there yet. According to the [documentation](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/DAG.md) the method `dag.tree` is only implemented in js-ipfs.

After that, I decided to write some code that is able to get the paths (no recursion yet) in order to get some exposure to the ipfs ecosystem and try to understand some of its components. I also added some tests for that.

Should I continue in this direction and add the logic for the recursive following of the links or the solution is finishing the HTTP-API?

Thank you! (:

### DIFF
--- a/src/dag/tree.js
+++ b/src/dag/tree.js
@@ -1,0 +1,37 @@
+'use strict'
+const promisify = require('promisify-es6')
+const block = require('../block')
+
+const DAGFormats = {
+  'dag-cbor': require('ipld-dag-cbor'),
+  'dag-pb': require('ipld-dag-pb')
+}
+
+module.exports = (send) => {
+  return promisify((cid, path, options, callback) => {
+    if (typeof path === 'function') {
+      callback = path
+      path = undefined
+    }
+
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
+    }
+
+    options = options || {}
+    path = path || ''
+
+    // FIXME: handle case when options.recursive is true
+    block(send).get(cid, options, (err, ipfsBlock) => {
+      if (err) return callback(err, null)
+
+      let codec = ipfsBlock.cid.codec
+      if (codec in DAGFormats) {
+        DAGFormats[codec].resolver.tree(ipfsBlock.data, callback)
+      } else {
+        callback(new Error(`codec ${codec} is not valid`), null)
+      }
+    })
+  })
+}
--- a/test/dag.spec.js
+++ b/test/dag.spec.js
@@ -1,0 +1,77 @@
+/* eslint-env mocha */
+'use strict'
+
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+
+const IPFSApi = require('../src')
+const f = require('./utils/factory')
+
+describe('.dag', function () {
+  this.timeout(50 * 1000) // slow CI
+
+  let ipfs
+  let ipfsd
+
+  const obj = {
+    a: 1,
+    b: [1, 2, 3],
+    c: {
+      ca: [5, 6, 7],
+      cb: 'foo'
+    }
+  }
+
+  before((done) => {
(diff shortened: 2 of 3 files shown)


src/dag/tree.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    options = options || {}
+    path = path || ''
+
+    // FIXME: handle case when options.recursive is true
+    block(send).get(cid, options, (err, ipfsBlock) => {
+      if (err) return callback(err, null)
+
+      let codec = ipfsBlock.cid.codec
```
vmx: Nit-pick: This could be a `const`.

test/dag.spec.js
```diff
@@ -0,0 +1,77 @@
...
+const chai = require('chai')
+const dirtyChai = require('dirty-chai')
+const expect = chai.expect
+chai.use(dirtyChai)
+
+const IPFSApi = require('../src')
+const f = require('./utils/factory')
+
+describe('.dag', function () {
+  this.timeout(50 * 1000) // slow CI
```
vmx: I guess that was copy&pasted. Generally try to stick to the default timeouts and see if they are long enough,

test/dag.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      expect(err).to.not.exist()
+      ipfs.dag.tree(cid, {}, (err, paths) => {
+        expect(err).to.not.exist()
+        expect(paths).deep.equal(expectedPaths)
+        done()
+      })
+    })
+  })
+
+  it('.dag.put', (done) => {
```
vmx: Looks like a left-over from copy&pasting. Please remove it if that's the case.

test/dag.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    let expectedCidStr = 'zdpuArMWc9Ee3B7zUDucRjvA1bDgYpWt8rpUXXjY3tbmBw619'
+    ipfs.dag.put(obj, (err, cid) => {
+      expect(err).to.not.exist()
+      let cidStr = cid.toBaseEncodedString()
+      expect(cidStr).to.be.equal(expectedCidStr)
+      done()
+    })
+  })
+
+  it('.dag.get', (done) => {
```
vmx: Looks like a left-over from copy&pasting. Please remove it if that's the case.

src/dag/tree.js
```diff
@@ -0,0 +1,37 @@
...
+      callback = options
+      options = {}
+    }
+
+    options = options || {}
+    path = path || ''
+
+    // FIXME: handle case when options.recursive is true
+    block(send).get(cid, options, (err, ipfsBlock) => {
+      if (err) return callback(err, null)
```
vmx: In the error callbacks, the return value is normally `undefined` and not `null`. So this would then just be `return callback(err)`.

test/dag.spec.js
pin3da: I tried with the default timeout but it was not enough.

test/dag.spec.js
pin3da: I did not know how to name this test. I looked into `log.spec.js` and `repo.spec.js` and tried to follow the convention there.

vmx: I was just wondering as this test doesn't have a call to `tree()`.

test/dag.spec.js
vmx: See the comment above.

vmx:
Please have a look at the failures of the continuous integration. The code coverage decreased. It gives some indication on which test cases should be added.

PR #823, opened 2018-08-01
feat: expose mfs files.read*Stream methods

achingbrain:
See ipfs/interface-ipfs-core#341 for the supporting tests.


alanshaw:
Tests all pass locally - just waiting on CI

PR #825, opened 2018-08-03
fix: config get

### LINKED ISSUE #822: ipfs.config.get() returns different value types
If the API is connected to a Go node, calling `ipfs.config.get()` returns a `Uint8Array`.

If the API is connected to a JS node, calling `ipfs.config.get()` returns, well, an object with the config.

Interface IPFS Core says "[the result] is a JSON object containing the configuration of the IPFS node" which is also confusing to me.

Should we normalize it to a simple object everywhere?

hacdias:
Fixes config.get by always returning an object and never a buffer. It can be a possible breaking change for those using js-ipfs-api with go-ipfs, but not with js-ipfs.

Fixes https://github.com/ipfs/js-ipfs-api/issues/822.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "eslint-plugin-react": "^7.10.0",
     "go-ipfs-dep": "~0.4.17",
     "gulp": "^3.9.1",
-    "interface-ipfs-core": "~0.73.0",
+    "interface-ipfs-core": "~0.75.0",
     "ipfsd-ctl": "~0.39.0",
     "pull-stream": "^3.6.8",
     "socket.io": "^2.1.1",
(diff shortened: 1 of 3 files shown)


hacdias:
@diasdavid @alanshaw what do you think? 😄

hacdias:
[What interface says](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/CONFIG.md#configget).

daviddias (changes requested):
Needs tests to make sure this doesn't break again.

hacdias:
@diasdavid

alanshaw (approved):

PR #794, opened 2018-06-26
feat: added resolve command

jbenet:
- js-ipfs-api is lacking the `ipfs resolve` command.
- it is also lacking in the ipfs/specs/core-api docs.
- but `ipfs resolve` has been an intended command for a long time

PR #826, opened 2018-08-06
feat: add resolve cmd

alanshaw:
Just being super careful not to accidentally remove the commit in https://github.com/ipfs/js-ipfs-api/pull/794


alanshaw:
commitlint expected to fail - everything else looking 💯

PR #827, opened 2018-08-08
wrong path

x5engine:


alanshaw:
Are you sure this is right @meteorplus?

Looks like this bundle is accessible at the path already specified here: https://github.com/ipfs/js-ipfs-api/blob/master/examples/bundle-webpack/webpack.config.js#L16

PR #831, opened 2018-08-12
Fix for dag.get failing silently on missing multicodec

lidel:
Right now `ipfs.dag.get` fails silently for CIDs with codecs different than `dag-cbor` or `dag-pb`.
To reproduce, try `dag.get` for  `raw` one at  `bafkreigh2akiscaildcqabsyg3dfr6chu3fgpregiymsck7e7aqa4s52zy`.

This PR fixes the bug by returning error when codec resolver is missing. 

Notes:
- Problem occurs only in js-ipfs-api, js-ipfs already delegates resolvers to ipld which has proper error handling  [here](https://github.com/ipld/js-ipld/blob/19498948477af8e33018b7472e76f02d4df926f6/src/index.js#L163).
- Added regression tests for `dag.put+get`, not sure if it is the right place tho. Any feedback would be appreciated.
- Not sure what is the long term plan for resolvers,  I assume endgame will be https://github.com/ipfs/js-ipfs-api/pull/755 ?

### DIFF
--- a/src/dag/get.js
+++ b/src/dag/get.js
@@ -7,6 +7,11 @@ const CID = require('cids')
 const waterfall = require('async/waterfall')
 const block = require('../block')
 
+const resolvers = {
+  'dag-cbor': dagCBOR.resolver,
+  'dag-pb': dagPB.resolver
+}
+
 module.exports = (send) => {
   return promisify((cid, path, options, callback) => {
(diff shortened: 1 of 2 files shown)


lidel:
@alanshaw applied suggested changes and rebased, ok to merge?

PR #833, opened 2018-08-14
testing: add a basic config test

Stebalien:


alanshaw:
🚀 this is very cool - thank you for taking the time! However, there is already a test in interface-ipfs-core that is very similar (all of those tests are run as part of the test suite here too):

https://github.com/ipfs/interface-ipfs-core/blob/2fbf551dd541f49582b3b4db9ea44c21c30ef96f/js/src/config/get.js#L33-L40

That test ensures that the response is a plain object and not a buffer, the only thing missing is a check that it contains a property `Addresses`.

Since they're so similar, I rather add the extra assertion to the existing test...

That said, and I'm speculating here, but if it's just there to ensure that you didn't get a buffer back then we're all good - that case is already covered ;)

Stebalien:
Ah, I see. A simple grep of the repo didn't reveal any tests so I wrote this to test a change I made to go-ipfs. If it's tested, that's good enough for me (and the core interface repo is a better place for it anyways).

PR #834, opened 2018-08-14
fix: remove external urls from addFromURL tests

### LINKED ISSUE #803: Remove reliance on external services for util.addFromURL tests
The `addFromURL` tests call out to URLs on the internet to add content to IPFS. This makes our tests dependent on the availability and response time of these services. Instead we should spin up a mock node server to respond to these requests for each test.

https://github.com/ipfs/js-ipfs-api/blob/a8f9a43064930c816b0e73dbc1674a313cf1baca/test/util.spec.js#L115-L197

alanshaw:
Relying on external URLs makes our tests dependent on the availability and response time of these services as well as not being able to run without an internet connection. This PR removes those requests to external URLs and instead spins up test HTTP(S) servers as they are needed to serve the content that `ipfs.util.addFromURL` consumes.

This PR also strengthens some of the tests to make assertions that the result received by `ipfs.util.addFromURL` is the same as the result of adding the data directly using `ipfs.add`.

resolves #803


achingbrain:
One comment - you can use the [pem](https://www.npmjs.com/package/pem) module to generate SSL certs on the fly instead of hard coding them.  Though it calls out to the openssl cli tool in the background, not sure of Windows support off the top of my head.  As long as it's installed on our Windows Jenkins workers, I guess.

PR #841, opened 2018-08-29
WIP: chore: add ipns core interface tests

vasco-santos:
PR aiming to have the `ipns` core interface tests enabled on the API.


vasco-santos:
Currently, we cannot have the CI green here for two main reasons. 

The first one is that we cannot run the **offline go daemon** (there is some work being done in [go-ipfs#4978](https://github.com/ipfs/go-ipfs/issues/4978) and [go-ipfs#4980](https://github.com/ipfs/go-ipfs/pull/4980)). Without running the offline daemon, we end up with timeouts for the tests.

In addition, there are others tests (not related to this PR), which need a `go-ipfs` release to pass.

PR #843, opened 2018-08-29
Support adding files using filestore

skeary-immt:
Hi,
Here is a small change to support adding files using the filestore in go-ipfs via the HTTP API. 
As described in #562, In order to use the filestore, an `Abspath` header must be specified which indicates the full path of the file in the local file system. This change provides a way to specify this header by using client code of the form:

```
  const filesAdded = await ipfs.files.add(
    {
      path: 'filename',
      abspath: '/full/path/to/filename',
      content: fs.createReadStream('/full/path/to/filename')
    },
    {
      nocopy: true,
    }
  )
```
Based on my testing, when using the updated code, and specifying the `abspath` and `nocopy` options, the go-ipfs daemon then uses the filestore for the added file.

I'm not sure if the best approach but wanted to create the pull request for consideration. I'm happy to update to an alternative approach if there is a better way. I think an update would also need to be done to the IPFS API spec but I can later do that if it's thought this approach has merit.

Thanks,
Simon

### DIFF
--- a/src/utils/send-files-stream.js
+++ b/src/utils/send-files-stream.js
@@ -22,6 +22,10 @@ function headers (file) {
     header['Content-Type'] = 'application/octet-stream'
   }
 
+  if (file.abspath) {
+    header.Abspath = file.abspath
+  }
+
   return header
 }
 


skeary-immt:
Oh, I just realised there is another PR #564 that appears to try and solve this issue too....

I'll leave this open as the approaches are a bit different so it may be worth reviewing my changes to see how it compares and whether it is the basis of something better or not. (I'm new to IPFS so could quite easily be missing something in what I've done...)

alanshaw:
@skeary-immt this LGTM on the surface - is there any chance we can get a test?

parkan:
would it make sense to make sending the header conditional on `nocopy`, as well?

obo20:
@alanshaw What's all needed for the --nocopy functionality to get fully implemented? I'd really love to see this functionality as part of this library, but I imagine there's a few more things that need to happen for this to become a reality.

PR #844, opened 2018-08-30
fix: block.put options

### LINKED ISSUE #815: block.put ignores CID argument
Hello, I'm having an issue where `block.get` is hanging and never returns the block.

Here's the example code:

```js
const ipfsAPI = require('ipfs-api')
const CID = require('cids')
const multihashes = require('multihashes')
const crypto = require('crypto')

function sha256(data) {
  return crypto.createHash('sha256').update(data).digest()
}

const ipfs = ipfsAPI('localhost', '5001', {protocol: 'http'})
const buf = sha256('hello world!')
const mh = multihashes.encode(buf, 'sha2-256')
const cid = new CID(1, 'dag-cbor', mh)

console.log('cid', cid.toBaseEncodedString())

ipfs.repo.gc()
.then(() => {
  return ipfs.block.put(buf, cid)
  .then(block => {
    console.log('put', block.cid.toBaseEncodedString())

    return ipfs.block.get(cid)
    .then(block => {
      // Never gets called!
      console.log('get', block._data.toString())
    })
    .catch(error => console.error(error))
  })
})
```

Output

```bash
cid zdpuAtJJwFer98hHGdt2UQdoF1We7vQuBXSTDKjorNkYovyzQ
put QmPdbXGenToPFzmxJyoVNAAfh5eKvnu2f8KzSx2J8jhwQP
```

It works using the CLI though:

```
$ ipfs repo gc

$ echo -n 'hello world!' | ipfs block put -f cbor
zdpuAtJJwFer98hHGdt2UQdoF1We7vQuBXSTDKjorNkYovyzQ

$ ipfs block get zdpuAtJJwFer98hHGdt2UQdoF1We7vQuBXSTDKjorNkYovyzQ
hello world!
```

Any help is appreciated, thanks.

alanshaw:
Alternative to #793 

Should fix this according to [the spec](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/BLOCK.md#blockput).

fixes #815 
closes #793
~~tested by https://github.com/ipfs/interface-ipfs-core/pull/370~~ (released)

### DIFF
--- a/src/block/put.js
+++ b/src/block/put.js
@@ -3,35 +3,62 @@
 const promisify = require('promisify-es6')
 const Block = require('ipfs-block')
 const CID = require('cids')
-const once = require('once')
+const multihash = require('multihashes')
 const SendOneFile = require('../utils/send-one-file')
 
 module.exports = (send) => {
   const sendOneFile = SendOneFile(send, 'block/put')
 
-  return promisify((block, cid, _callback) => {
-    // TODO this needs to be adjusted with the new go-ipfs http-api
-    if (typeof cid === 'function') {
-      _callback = cid
-      cid = {}
+  return promisify((block, options, callback) => {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
     }
 
-    const callback = once(_callback)
+    options = options || {}
 
     if (Array.isArray(block)) {
       return callback(new Error('block.put accepts only one block'))
     }
 
-    if (typeof block === 'object' && block.data) {
-      block = block.data
+    if (Buffer.isBuffer(block)) {
+      block = { data: block }
     }
 
-    sendOneFile(block, {}, (err, result) => {
+    if (!block || !block.data) {
+      return callback(new Error('invalid block arg'))
+    }
+
+    const qs = {}
+
+    if (block.cid || options.cid) {
+      let cid
+
+      try {
+        cid = new CID(block.cid || options.cid)
+      } catch (err) {
+        return callback(err)
+      }
+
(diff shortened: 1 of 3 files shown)


alanshaw:
Please note that the code has a temporary workaround for https://github.com/ipfs/go-cid/issues/75

src/block/put.js
```diff
@@ -3,35 +3,72 @@
...
-    if (typeof cid === 'function') {
-      _callback = cid
-      cid = {}
+  return promisify((block, options, callback) => {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
     }
 
-    const callback = once(_callback)
```
diasdavid: I can't remember but I wonder if this was a protection to go-ipfs sometimes sending multiple errors ?!

alanshaw: Noted, I guess we'll find out soon ;)

src/block/put.js
```diff
@@ -55,7 +55,17 @@ module.exports = (send) => {
 
     sendOneFile(block.data, { qs }, (err, result) => {
       if (err) {
-        return callback(err) // early
+        // Retry with "protobuf" format for go-ipfs
+        // TODO: remove when https://github.com/ipfs/go-cid/issues/75 resolved
+        if (qs.format === 'dag-pb') {
+          qs.format = 'protobuf'
```
neodon: This is the key change for me that would have allowed me to complete #815.  Still learned a lot from trying to work that story.  Time to find a new issue to dig into. :)

Issue #842, opened 2018-08-29
Streaming versions of ipfs.files.add buffer entire data in memory before doing HTTP POST


lidel:
```
ipfs.files.addPullStream([options])
ipfs.files.addReadableStream([options])
```
## Problem Summary

Streaming over HTTP does not work the way people expect: it ends before HTTP transport and entire thing is buffered in memory before HTTP POST is called with entire thing in its payload.

This is bad because:
- breaks progress reporting (https://github.com/ipfs/js-ipfs-api/issues/748)
  (0% until entire thing is received by remote server, then 100% instantly)
- web browsers crash (https://github.com/ipfs/js-ipfs-api/issues/654)
  or return silent errors due to increased  memory allocation
  (https://github.com/ipfs-shipyard/ipfs-companion/issues/464#issuecomment-382446114)

## Potential Solutions

- Writable streams from  [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API):
   > The [`WritableStream`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream) interface of the the Streams API provides a standard abstraction for writing streaming data to a destination, known as a sink. This object comes with build-in backpressure and queuing.
   - Not supported by Firefox yet: https://caniuse.com/#feat=streams
- Add manual chunking between HTTP API and js-ipfs-api (?)
-  (?)

## Related Issues

cc https://github.com/ipfs/js-ipfs-api/issues/654, https://github.com/ipfs/js-ipfs-api/issues/788, https://github.com/ipfs/js-ipfs-api/issues/748, https://github.com/ipfs-shipyard/ipfs-companion/issues/464

hacdias:
cc https://github.com/ipfs-shipyard/ipfs-webui/issues/762

JiaJiaJiang:
I haven't read all code of the uploading logic, but if the data will not be processed before uploading ,here is an advice: 
**XHR** can send a **FormData** object, and the formdata object can carry files that gave into the input element. This way will not cause the memory problem and all thing are handled by the browser. The uploading progress can be got from the xhr object too, nothing will be blocked.
Here is the example of FormData object whih input files: [https://developer.mozilla.org/en-US/docs/Web/API/FormData/append](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)

Issue #797, opened 2018-06-28
ipfs.files.add is missing error handling on invalid response from go-ipfs


lidel:
### The Bug

Sometimes (for specific multipart payloads in browser context) the `ipfs.files.add`  backed provided by go-ipfs returns an error in the middle of payload:

```js
{"Name":"T-RexFaceEmoji.jpg_original","Hash":"QmR1G8Jq4rNiYwCKyNWEZhNS7i8EWrCpz2xDjKsB9ofi82","Size":"23723"}
{"Name":"FlyingSaucerEmoji.jpg_original","Hash":"QmVWxHcLVHYpFJP6R1ZLBZcpAZsxyFMfU7BU84D2LAMSed","Size":"13071"}
{"Message":"http: invalid Read on closed Body","Code":0,"Type":"error"}
```

Then, upon receiving this payload, `js-ipfs-api` does not throw an error, but silently ignores the last line and returns a partial result list with only two first items. ([screenshot](https://user-images.githubusercontent.com/157609/42034774-5ca014bc-7ae1-11e8-8af3-ba3401169513.png))

A similar problem occurs for some single file uploads (with `wrapWithDirectory: true`), js-ipfs-api returns an empty list of responses `[]` ([screenshot](https://user-images.githubusercontent.com/157609/42034747-4742caa6-7ae1-11e8-873b-42f3260255d1.png))

### How to Reproduce

Turns out the `.zip` with sample app triggers the issue :upside_down_face: , so its easy to reproduce:

1. Expose API of go-ipfs 0.4.15 or 0.4.16-rc1 locally on port `:5001`
2. Download demo app: [`upload-multiple-files-via-browser-ipfs-api-bug-demo.zip`](https://github.com/ipfs/js-ipfs-api/files/2145286/upload-multiple-files-via-browser-ipfs-api-bug-demo.zip)
  (uses `ipfs-api ^22.1.1` and uploads files via `ipfs.files.add` with `wrapWithDirectory: true`)
3. Build it and start via `npm install && npm start`, then open form at http://localhost:3000
4. Reproduce using samples known to trigger issue at go-ipfs
   - Sample 1: upload single file: `upload-multiple-files-via-browser-ipfs-api-bug-demo.zip`
       - This should produce [this error](https://user-images.githubusercontent.com/157609/42034747-4742caa6-7ae1-11e8-873b-42f3260255d1.png)
[...]

Issue #788, opened 2018-06-19
upload-file-via-browser can only upload file size less than 165MB


liao1k:
Hi JS-IPFS-API Development Team,
   I use upload-file-via-browser to upload file into IPFS, it can only upload the file size less than 160MB. Any file size large than 160MB, the web browser just stay here, no returning file hash id link showing on the web browser. Also Does "upload-file-via-browser" support upload multiple files?  The current version can only support upload one single file at a time. Any comments are highly appreciated! Thanks for your help.

Kevin

liao1k:
This is my demonstration web site: http://ipshare.info:3000/
Visit the above web site and upload a file which size is less than 160MB. The web browser can return the hash id, then try to upload a large file and you will see the local browser crashed. See the screenshot in the attachment. 
![uploadlargefile](https://user-images.githubusercontent.com/37547030/41617572-d6fcad8e-73ce-11e8-8a4d-d12cb450692d.JPG)

I use google chrome Version 67.0.3396.87 (Official Build) (64-bit) for the local browser to upload file into IPFS system. 

Webpack:
npm list webpack
upload-file-via-browser@1.0.0 /home/donovan/UploadFiles
└── webpack@2.7.0

node -v:
v8.11.2


Operating System (Ubuntu: Ubuntu 18.04 LTS)
 uname -a
Linux ipfs-dlchains 4.15.0-23-generic #25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

$ npm start
> upload-file-via-browser@1.0.0 start /home/donovan/UploadFiles
> node server.js

Listening at 172.16.0.254:3000
Hash: 6b3687b304ba5ea7451a
Version: webpack 2.7.0
Time: 2083ms
    Asset     Size  Chunks                    Chunk Names
bundle.js  4.78 MB       0  [emitted]  [big]  main
chunk    {0} bundle.js (main) 4.31 MB [entry] [rendered]
   [23] (webpack)/buildin/module.js 517 bytes {0} [built]
  [122] ./~/react-dom/lib/ReactMount.js 25.4 kB {0} [built]
  [335] ./~/react-hot-api/modules/index.js 57 bytes {0} [built]
  [336] ./~/react-hot-loader/RootInstanceProvider.js 1.18 kB {0} [built]
  [338] ./~/react-hot-loader/makeExportsHot.js 1.73 kB {0} [built]
[...]

Issue #654, opened 2017-12-24
Uploading big files crush page in Chrome


mattermoran:
So I'm doing as in this example https://github.com/ipfs/js-ipfs-api/tree/master/examples/upload-file-via-browser and everything works fine. But when I try to upload files that more than 3gb page crushes. 
![screen shot 2017-12-14 at 5 29 13 pm](https://user-images.githubusercontent.com/29712386/34324845-1615900c-e8bb-11e7-9245-448743bc334b.png)

kostysh:
I am facing the same issue

julientregoat:
I have the same issue, except with files over 50-100mb crashing it. I've tried ReadableStream, PullStream, buffer instances... nothing working. Smaller files get there just fine. Running node 8.10

diasdavid:
@julientregoat -- https://github.com/ipfs/js-ipfs/issues/952 -- says that things are good uploading over 1GB on js-ipfs. Here, you are using js-ipfs-api, the HTTP Client library and talking with a go-ipfs daemon, it is a different scenario.

This should just be a buffering/streaming issue when creating the HTTP request. If anyone is available to tackle this, it would be great to have it solved.

julientregoat:
@diasdavid ah got it! sorry, I didn't understand. still new to this but will try to play with it to see if I can get anything going.

channprj:
It works on Safari. Just crashes in Chrome. How can I handle this? Does anybody work on it?

diasdavid:
![](https://media.giphy.com/media/qIXVd1RoKGqlO/giphy.gif)

PR #851, opened 2018-09-03
[WIP] feat: add support for chunked uploads

hugomrdias:
This is still a work in progress to add support for chunked uploads (`ipfs.add`) and fix multiple issues related to adding big files.

Tests are filtered here https://github.com/ipfs/js-ipfs-api/blob/90c40363fbcd55d29307e51f4feabb8be867ded8/test/add-experimental.spec.js#L38-L46 to make review easy, just run ipfs daemon with https://github.com/ipfs/js-ipfs/pull/1540 

features/fixes in this PR together with https://github.com/ipfs/js-ipfs/pull/1540:
- big data add non-chunked (this will either break with browser memory or hit the maxBytes config in the daemon, see next)
- really big data add chunked (theoretically the limit is daemon disk space or maybe request timeouts)
- streaming progress reporting 
- error handling and reporting
- add multiple files with wrapWithDirectory
- improved browser support, handles `File`'s directly from the input 
```js
const files = document.getElementById('file').files;
        this.ipfsApi
            .add([...files], {
                wrapWithDirectory: true,
                experimental: true,
                progress: prog => console.log(`received back: ${prog}`)
                chunkSize: 10 * 1024 * 1024
            })
            .then(console.log)
            .catch(console.error);
```
- jsdoc for top level api and more

Notes: 
- trailers https://stackoverflow.com/questions/13371367/do-any-browsers-support-trailers-sent-in-chunked-encoding-responses

Needs: 
- https://github.com/ipfs/js-ipfs/pull/1540

Todo:

- [x] validate this example works after [https://github.com/ipfs/js-ipfs-api/tree/master/examples/upload-file-via-browser](https://github.com/ipfs/js-ipfs-api/tree/master/examples/upload-file-via-browser)
- [x] what to do with progress ? add another handler ?
- [x] multiple files return from the daemon only first hash
- [ ] ~~concurrent upload chunks~~ new PR for this
- [x] check uuid impl (maybe change to uuid v5 or [nano-id](https://github.com/ai/nanoid))
- [x] avoid preflight as much has possible
- [x] callbackify top level
- [x] try handling non chunked
- [x] fix multipart boundary handling for non chunked

Related: 
  - [https://github.com/ipfs/js-ipfs-api/issues/654](https://github.com/ipfs/js-ipfs-api/issues/654)
  - [https://github.com/ipfs/js-ipfs-api/issues/788](https://github.com/ipfs/js-ipfs-api/issues/788)
  - [https://github.com/ipfs/js-ipfs/issues/952](https://github.com/ipfs/js-ipfs/issues/952)
  - [https://github.com/ipfs/js-ipfs-api/issues/842](https://github.com/ipfs/js-ipfs-api/issues/842)
  - https://github.com/ipfs/js-ipfs-api/issues/797 - make sure this PR fixes this /cc @lidel 
  - https://github.com/ipfs-shipyard/ipfs-companion/issues/480 - will this PR fix this ? /cc @lidel

### DIFF
--- a/src/utils/multipart-experimental.js
+++ b/src/utils/multipart-experimental.js
@@ -156,6 +156,13 @@ class Multipart extends Duplex {
     return true
   }
 
+  /**
+   * Consume file and push to readable stream
+   *
+   * @param {(Buffer|Readable|PullStream)} file
+   * @param {function(): void} callback
+   * @return {void}
+   */
   pushFile (file, callback) {
     this.pushChunk(leading(file.headers, this._boundary), true)
 


lidel:
@hugomrdias thanks!

My view is that we should do our best to  to make it work without changing current `progress` API.
Details of the chunked upload should be abstracted away in best-effort fashion and hidden behind existing progress reporter.

What if we detect presence of `chunkSize`  parameter, and switch logic used for progress reporting behind the scenes?

For upload split into N chunks:
- uploading chunks 1 to (N-1) would show "upload only progress"
  (initially we could just return % based on the number of uploaded chunks, more resolution can be added later)
- uploading the last chunk  N could show real "add progress" but only when it is bigger than "upload progress" 

The end result would be a best-effort progress reporting that works with existing API and that is not stuck at 0% until the last chunk and behaves in expected manner (% always grows).

hugomrdias:
ipfs-chunked-add

hugomrdias:
the jsdoc should create some nice docs with documentation.js

api-docs

also should give code completion to anyone using editors with jsdoc support 
api-completion
this can bubble up to the top level public api with minimal changes to this [file](https://github.com/ipfs/js-ipfs-api/blob/3cd19e721bb529e36c53e76d835ac8625741e6eb/src/utils/load-commands.js)

src/utils/prepare-file.js
```diff
@@ -77,30 +84,79 @@ function prepareFile (file, opts) {
...
+    return prepare(file, opts)
+  })
+}
 
-      return loadPaths(opts, file)
-    }
+function prepare (file, opts) {
+  // probably it should be valid and would be handled below with Buffer.from
+  if (typeof file === 'string') {
+    throw new Error('String isn\'t valid as an input')
```
vasco-santos: Use [err-code](https://www.npmjs.com/package/err-code) here. We are trying to use errors with code everywhere.

src/utils/prepare-file.js
```diff
@@ -1,7 +1,14 @@
 'use strict'
 
 const isNode = require('detect-node')
+const { isSource } = require('is-pull-stream')
+const isStream = require('is-stream')
 const flatmap = require('flatmap')
+const fileReaderStream = require('filereader-stream')
+
+const isBrowser = typeof window === 'object' &&
```
vasco-santos: Is this not achievable with `isNode`?

src/utils/prepare-file.js
```diff
@@ -77,30 +84,79 @@ function prepareFile (file, opts) {
...
+      content: Buffer.from(file)
     }
-  })
+  }
+
+  return {
+    path: '',
+    symlink: false,
+    dir: false,
+    content: file
```
vasco-santos: I think it would be cleaner to have a single return. Then the above conditions would set only `path` and `content`. What do you think?

src/utils/send-stream-experimental.js
```diff
@@ -0,0 +1,199 @@
...
+const noop = () => {}
+
+/**
+ * Convert back to the proper schema
+ *
+ * @private
+ * @param {Object} data
+ * @returns {Object}
+ */
+const convert = (data) => {
```
vasco-santos: Since this is `private`, it should be named `_arrayToStream`. However, it can also be moved to a `util`. Same below with `prepareTransform`

src/utils/send-stream-experimental.js
```diff
@@ -0,0 +1,199 @@
...
+    super({ objectMode: true, highWaterMark: 1 })
+    this.options = options
+    this.send = send
+    this.multipart = new Multipart(options)
+    this.boundary = this.multipart._boundary
+    this.uuid = nanoid()
+    this.index = 0
+    this.rangeStart = 0
+    this.rangeEnd = 0
+    this.rangeTotal = 0
```
vasco-santos: What about using an object to handle all the range related properties?

src/utils/send-stream-experimental.js
```diff
@@ -0,0 +1,199 @@
...
+      }
+    })
+
+    if (options.chunkSize) {
+      this.multipart.on('data', this.onData.bind(this))
+    }
+  }
+
+  _read () {
+    // empty read
```
vasco-santos: Could you add the justification for the `empty read`?

test/add-experimental.spec.js
```diff
@@ -0,0 +1,516 @@
...
+        }
+      })
+
+      files.forEach((file) => stream.write(file))
+      stream.end()
+    })
+    // end runs
+  })
+
+  it.skip('files.add pins by default', (done) => {
```
vasco-santos: What is blocking this?

test/add-experimental.spec.js
```diff
@@ -0,0 +1,516 @@
...
+        ipfs.pin.ls((err, pins) => {
+          expect(err).to.not.exist()
+          expect(pins.length).to.eql(initialPinCount + 1)
+          done()
+        })
+      })
+    })
+  })
+
+  it.skip('files.add with pin=false', (done) => {
```
vasco-santos: What is blocking this?

test/add-experimental.spec.js
hugomrdias: Doesn't run connected to a js daemon, not related to this PR

test/add-experimental.spec.js
hugomrdias: Doesn't run connected to a js daemon, not related to this PR

alanshaw:
@Stebalien could we get your thoughts on adding this to go-ipfs?

This PR is adding a feature to the HTTP add endpoint that will allow big files to be uploaded to IPFS by making _multiple_ requests.

@lidel kindly put together a [good summary of the proposed process](https://github.com/ipfs/js-ipfs-api/pull/851#pullrequestreview-151877090):

> - Upload payload is split into small parts (`chunkSize = 256000`) 
> - Each part is sent as a a sequence of HTTP POST requests that have
>   -  a unique identifier for entire upload session  (uuid? – see below)
>   -  a sequential counter within upload session (a chunk index)
> - API backend needs to support additional HTTP headers to perform re-assembly of entire payload from chunks and passing it to the regular `ipfs.files.add` call in a transparent manner
>   - PR for js-ipfs: https://github.com/ipfs/js-ipfs/pull/1540
>   - PR for go-ipfs: (TODO)

Reasons for doing this:

1. It's not possible to stream a HTTP upload request (in _Firefox_) without buffering the entire payload into memory first
2. Has potential to allow resume for failed upload requests

alanshaw:
@hugomrdias @lidel I think that regardless of what happens with this PR we need to switch to using the streaming fetch API. Firefox is notably the only browser that hasn't shipped the streams API yet but it [sounds like this might happen soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1389628#c6). I think we can conditionally opt out of it for Firefox for the time being.

Switching to using the streaming fetch API will solve the buffering issue without any changes to the HTTP API and depending on priorities for go-ipfs we might be able to ship this before chunked uploads.

It's also worth noting that, streaming fetch will be way more efficient then multiple HTTP requests for chunked uploading.

hugomrdias:
That's only for response bodies not request bodies, this is the only way

alanshaw:
> That's only for response bodies not request bodies, this is the only way currently available to us. I didn't find any indication that request bodies will get streams soon on any browser

You're absolutely right - my bad. Thanks for clarifying!

lidel:
Sounds like worth mentioning here, in case concepts from this PR are revisited in the future:



   - a proposal of open protocol for resumable file uploads: 
     https://tus.io / https://tus.io/protocols/resumable-upload.html

hugomrdias:
> Sounds like worth mentioning here, in case concepts from this PR are revisited in the future:
> 
> * a proposal of open protocol for resumable file uploads:
>   [tus.io](https://tus.io) / [tus.io/protocols/resumable-upload.html](https://tus.io/protocols/resumable-upload.html)

yep i based the impl in tus

PR #854, opened 2018-09-10
fix: add bl module to package dependencies (#853)

### LINKED ISSUE #853: Add bl module to package.json dependencies
Hi, I'm using ipfs-api (and I don't use js-ipfs) in my electron project.

bl is used in https://github.com/ipfs/js-ipfs-api/blob/master/src/files/cat.js#L6

but in package.json, it is not in dependencies, so error occurs when executing my electron program builded. (Of course, if I add "bl": "^2.0.1" module to package.json dependencies in my project, it works)

Can you add it to package.json?

seungwon-kang:
Closes #853

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "dependencies": {
     "async": "^2.6.1",
     "big.js": "^5.1.2",
+    "bl": "^2.0.1",
     "bs58": "^4.0.1",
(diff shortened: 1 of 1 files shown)


alanshaw:
Merging as CI is failing because out of space, unrelated to this PR

PR #871, opened 2018-10-08
Fix/lodash dependency

nGoline:
Must add lodash as a dependency because it's required in 2 files.


alanshaw:
Hi @nGoline, thanks for the PR - our build tool generates changelogs, increments version numbers and adds contributors automatically - would you mind resubmitting with just the lodash dependency addition?

nGoline:
Hi @alanshaw. I just did it and the build is failing apparently with no regards to my code:
```No such property: currentOS for class: groovy.lang.Binding```

nGoline:
Ok, I'm having some problems with the size of my commit message.
I'll create a new branch and perform only one commit considering the actual change needed.

@alanshaw, I think you can delete this one...

PR #873, opened 2018-10-09
fix: add lodash dependency

### LINKED ISSUE #870: Should add lodash as a dependency
Lodash is required in [./files/stat.js](https://github.com/ipfs/js-ipfs-api/blob/master/src/files/stat.js) and [./pin/ls.js](https://github.com/ipfs/js-ipfs-api/blob/master/src/pin/ls.js) but is not referenced in package.json;

This is causing errors when the core project doesn't require lodash.

nGoline:
Add lodash as a dependency because it's required in 2 files.

This PR is related to [PR 871](https://github.com/ipfs/js-ipfs-api/pull/871).

@alanshaw this follows the requirements of the build tool. Sorry for that!

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "is-pull-stream": "0.0.0",
     "is-stream": "^1.1.0",
     "libp2p-crypto": "~0.13.0",
+    "lodash": "^4.17.11",
     "lru-cache": "^4.1.3",
     "multiaddr": "^5.0.0",
     "multibase": "~0.4.0",


nGoline:
Can someone explain to me why my commit breaks the "commitlint" process?
```bash
⧗   input: 
Add lodash dependency; Closes #870

✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
```

I'm commiting via bash in VSCode for Windows.

Also, why does it timeout that much and how can I signal jenkins to try again?

Regards.

hugomrdias:
pls check this link https://github.com/ipfs/community/blob/master/js-code-guidelines.md#commits

nGoline:
Sorry for I've missed that @hugomrdias.

All set now!

The only thing now is that I've changed nothing else than the dependency but I got an error on macos:10.11.0 test:node

hugomrdias:
everything is passing now

alanshaw:
Thanks @nGoline

PR #879, opened 2018-10-26
fix: deps, all deps

hugomrdias:


diasdavid:
a few CI fails

image


none related to this PR. Everything passes on my machine. 

@alanshaw shall I go ahead and merge?

PR #882, opened 2018-10-30
Fix log.tail by calling add after listening for events

magik6k:
This got broken in go-ipfs 0.4.18, I have no idea why it worked before..

### DIFF
--- a/test/log.spec.js
+++ b/test/log.spec.js
@@ -31,11 +31,16 @@ describe('.log', function () {
   })
 
   it('.log.tail', (done) => {
+    let i = setInterval(() => {
+      ipfs.files.add(Buffer.from('just adding some data to generate logs'))
+    }, 1000)
+
     const req = ipfs.log.tail((err, res) => {
       expect(err).to.not.exist()
       expect(req).to.exist()
 
       res.once('data', (obj) => {
+        clearInterval(i)
         expect(obj).to.be.an('object')
         done()
       })


achingbrain:
There are some linting failures:

1. Strings should use single quotes
2. The commit message should not be sentence case so I think `fix: Fix log.tail..` needs changing to be `fix: fix log.tail`

diasdavid:
@magik6k rebased master onto your branch so that it runs with latest go-ipfs.

Note, I didn't see it fail on master.

test/log.spec.js
```diff
@@ -31,7 +31,12 @@ describe('.log', function () {
   })
 
   it('.log.tail', (done) => {
+    let i = setInterval(() => {
+      ipfs.files.add(Buffer.from('just adding some data to generate logs'))
+    }, 1000)
+
     const req = ipfs.log.tail((err, res) => {
+      clearInterval(i)
```
alanshaw: I'm not sure if I'm mis-interpreting what is happening here. My understanding of this is:

1. We setup an interval to run once every second
2. Call `ipfs.log.tail` - which will callback almost immediately with a stream for us to consume
3. Clear the interval - presumably before the first time it runs (unless `ipfs.log.tail` takes > 1s to callback)

So I think adding this interval is either ineffectual, or it's relying on `ipfs.log.tail` to take > 1s to callback.

If this test is passing then it might be a false positive, which we need to get to the bottom of.

magik6k: Yeah, this should be in `res.once` callback.. It was quite late when I wrote this fix.

Looked at this more - with `ipfs daemon --offline`, doing `curl http://127.0.0.1:5001/api/v0/log/tail -v`, go-ipfs will not send response headers until first log message, so this is why this worked.

This test also works currently because go-ipfs tends to report some random events, but there is no guarantee that they will happen and this causes the test to hang for way too long.

alanshaw:
Merging as test failures are for ping tests - unrelated to this PR.

PR #887, opened 2018-11-06
fix: handle peer-info validation errors

### LINKED ISSUE #885: Handle errors when unpacking swarm.peers() response
The addition of the quic protocol means there are now peers on the network with a multiaddr that cannot be unpacked by older verions of js-multiaddr, and it throws.

In the current implementation of `ipfsApi.swarm.peers()` we don't handle any errors that could occur when trying to wrap the peer addr from the repsonse in a Multiaddr: 

https://github.com/ipfs/js-ipfs-api/blob/ead35992e9d5c60f85d74f0d80aad1b0e994ea5c/src/swarm/peers.js#L46-L51

the simplest solution would be to re-throw the error with an informative error message, but that still leaves us with the siutation when newer peers get added to the network, they can break my user-experience, as I suddenly start seeing 0 peers; js-ipfs-api throws an error if 1 peer appears invalid like in https://github.com/ipfs-shipyard/ipfs-webui/issues/878

A more helpful solution might be to provide some way to signal to the caller that some peers have information that we can't validate, but still do a best effort to return information about the response.

If we go the best-effort route, we've got

1. `easy` - remove peers from the list if we hit an error trying to validate their info, and write a debug log. This is simple, but leaves the peer list that you get from ipfs-api containing fewer items than the http repsonse did, which is undesireable.

2. `hard` - update the api response info to include peers that can't be validated. We could add a property to an idividual peer like `error` with the validation error, and the `raw` with the unvalidated response data. Or we could pull out unvalidtable peers into a sepearate list and return two arrays, could be `peers` (the valid ones) and `rawPeers` (all peers)...

olizilla:
BREAKING CHANGE. Previously swarm.peers would throw an uncaught error
if any peer in the reponse couldn't have its peerId or multiaddr validated.

This PR catches errors that occur while validating the peer info. The
returned array will contain an entry for every peer in the ipfs response.
peer-info objects that couldn't be validated, now have an `error` property
and a `rawPeerInfo` property. This at least means the count of peers in
the response will be accurate, and there the info is available to the caller.

This means that callers now have to deal with peer-info objects that may
not have a `peer` or `addr` property.

Adds `nock` tests to exercice the code under different error conditions.
Doing so uncovered a bug in our legacy go-ipfs <= 0.4.4 peer info parsing,
which is also fixed. The code was trying to decapusalate the peerId from
the multiaddr, but doing so trims the peerId rather than returning it.

fixes #885

License: MIT

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "libp2p-crypto": "~0.14.0",
     "lodash": "^4.17.11",
     "lru-cache": "^4.1.3",
-    "multiaddr": "^5.0.0",
+    "multiaddr": "^5.0.2",
     "multibase": "~0.5.0",
     "multihashes": "~0.4.14",
     "ndjson": "^1.5.0",
@@ -87,6 +87,7 @@
     "gulp": "^3.9.1",
     "interface-ipfs-core": "~0.84.3",
     "ipfsd-ctl": "~0.40.0",
+    "nock": "^10.0.2",
     "pull-stream": "^3.6.9",
     "stream-equal": "^1.1.1"
   },
--- a/src/swarm/peers.js
+++ b/src/swarm/peers.js
@@ -10,56 +10,65 @@ module.exports = (send) => {
       callback = opts
       opts = {}
     }
-
     const verbose = opts.v || opts.verbose
-
     send({
       path: 'swarm/peers',
       qs: opts
-    }, (err, result) => {
+    }, (err, response) => {
       if (err) {
         return callback(err)
       }
-
-      // go-ipfs <= 0.4.4
-      if (result.Strings) {
-        return callback(null, result.Strings.map((p) => {
-          const res = {}
-
-          if (verbose) {
-            const parts = p.split(' ')
-            res.addr = multiaddr(parts[0])
-            res.latency = parts[1]
-          } else {
-            res.addr = multiaddr(p)
-          }
-
-          res.peer = PeerId.createFromB58String(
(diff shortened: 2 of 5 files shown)


olizilla:
This also fixes https://github.com/ipfs-shipyard/ipfs-webui/issues/878

diasdavid:
Remove the `.spec` from the filename and add a node.js file that has a `require` for those tests. See https://github.com/ipfs/js-ipfs/blob/master/test/core/node.js for example

alanshaw (changes requested):
Code LGTM. Do you have a moment to PR https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/SWARM.md#swarmpeers to make a note of the change?

olizilla:
@alanshaw related to our conversations about the multiaddr validation problem, I found this https://github.com/multiformats/multiaddr/issues/70

olizilla:
The [`codelint`](https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/PR-887/4/pipeline/268#step-293-log-9) and [`linux/8.12/test:node`](https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-api/detail/PR-887/4/pipeline/361#step-491-log-122) CI failures are unrelated to this PR

There is a PR to update the interface-core spec here https://github.com/ipfs/interface-ipfs-core/pull/393

alanshaw (approved):

PR #888, opened 2018-11-07
fix: also retry with misnamed format "dag-cbor" as "cbor"

alanshaw:
Prevents error unrecognized format: 'dag-cbor'


alanshaw:
Merging as CI failures are unrelated - ping test failures

PR #889, opened 2018-11-08
fix: updates ipld-dag-pb dep to version without .cid properties

achingbrain:
Follows on from https://github.com/ipld/js-ipld-dag-pb/pull/99 and updates this module to not rely on DAGNodes having knowledge of their CIDs.


achingbrain:
The ping and pubsub tests will be the end of me.

PR #890, opened 2018-11-09
fix: dht validate if receiving stream

vasco-santos:
With `go-ipfs` we receive a stream of data. However, we currently do not receive a stream using `js-ipfs`. Types of notifications in `go-ipfs` are defined [here](https://github.com/libp2p/go-libp2p-routing/blob/master/notifications/query.go#L17-L26).

Needs:

- [x]  [interface-ipfs-core#389](https://github.com/ipfs/interface-ipfs-core/pull/389)
- [x]  [interface-ipfs-core#414](https://github.com/ipfs/interface-ipfs-core/pull/414)


src/dht/findpeer.js
```diff
@@ -17,10 +19,58 @@ module.exports = (send) => {
...
+
+      if (res.Responses) {
+        id = res.Responses[0].ID
+        addrs = res.Responses[0].Addrs
+      } else {
+        id = res.responses[0].id
+        addrs = res.responses[0].addrs
+      }
+
+      // inconsistencies js / go - go does not add `/ipfs/{id}` to the address
```
alanshaw: Do we _need_ to fix this here?

src/dht/findpeer.js
```diff
@@ -17,10 +19,58 @@ module.exports = (send) => {
...
+      let id
+      let addrs
+
+      if (res.Responses) {
+        id = res.Responses[0].ID
+        addrs = res.Responses[0].Addrs
+      } else {
+        id = res.responses[0].id
+        addrs = res.responses[0].addrs
+      }
```
alanshaw: Apologies if I wasn't clear, `interface-ipfs-core` should return lowered property names but the HTTP API should be compatible with go-ipfs.

Here I expect us to be converting "Uppered" property names to lowered form. In js-ipfs the HTTP API should take the lowered property names from core and "Upper" them to be compatible. I know this is a PITA but this is consistent with all our APIs.

vasco-santos: Will change it! Thanks for clarifying.

src/dht/findpeer.js
vasco-santos: So, `go-ipfs` response contains only address. As a consequence, it will fail the following test:

[ipfs/interface-ipfs-core/js/src/dht/findpeer.js#L48](https://github.com/ipfs/interface-ipfs-core/blob/master/js/src/dht/findpeer.js#L48)

I can change the test in `interface-ipfs-core` and remove this if you feel that it is less hacky.

alanshaw: What I'm concerned about is that we're making assumptions about the format of the address. What does go-ipfs return when it's a circulit address? Would that cause a false positive?

Ideally both implementations should return the same format! Do you know the reason why go-ipfs returns the addr without the peer ID? Perhaps for a smaller payload? It would be pretty easy for js-ipfs to decapsulate the peer id from the addrs wouldn't it?

alanshaw:
`interface-ipfs-core@0.91` is released

src/dht/findpeer.js
```diff
@@ -17,10 +22,44 @@ module.exports = (send) => {
...
+        r.Addrs.forEach((addr) => {
+          const ma = multiaddr(addr)
+
+          peerInfo.multiaddrs.add(ma)
+        })
+
+        return peerInfo
+      })
+
+      callback(null, responses)
```
alanshaw: Needs switch to returning a single response :D

PR #891, opened 2018-11-13
Update dependencies & config for `upload-file-via-browser` demo.

edsgerlin:
Packages used in the `upload-file-via-browser` demo is too old.
The PR update all of them and fixed all the warnings.


alanshaw:
Awesome! Thanks you @edsgerlin

PR #900, opened 2018-11-29
docs(readme): remove CORS Credentials header

lidel:
(afaik) HTTP API does not need this and we should not ask people to set it if they don't really need it.

See also:

- Potential problem with  `Access-Control-Allow-Credentials`: https://github.com/ipfs/go-ipfs/issues/5745
- HTTP Headers Cleanup: https://github.com/ipfs/in-web-browsers/issues/132


alanshaw:
Merging as CI failures are unrelated ping test failures (soon to be resolved in `interface-ipfs-core@0.89`)

PR #910, opened 2018-12-12
fix: added rm method for blocks

### LINKED ISSUE #792: block.rm() is missing :(
Was surprised to find that this API was missing. It's in the go ipfs command line but isn't in `js-ipfs-api` and isn't documented in https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/BLOCK.md

sweetpalma:
I've seen a command "ipfs block rm" in a CLI commands lists and HTTP, but seems like it is absent in `js-ipfs-http-client` package. I've opened this basic merge-request that adds it, but I am not sure what else should got here. Tested on my pet project, `ipfs.block.rm` worked like a charm.

resolves #792

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ const ipfs = ipfsClient({
 - [block](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/BLOCK.md)
   - [`ipfs.block.get(cid, [options], [callback])`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/BLOCK.md#blockget)
   - [`ipfs.block.put(block, [options], [callback])`](https://github.com/ipfs/interface-ipfs-core/blob/master/SPEC/BLOCK.md#blockput)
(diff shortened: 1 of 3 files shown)


sweetpalma:
As far as I can see, I need to upgrade https://github.com/ipfs/interface-ipfs-core too - I see nothing about that in the CONTRIBUTING guide. Could somebody help me with this? I really want this feature in JS client.

Prabhakar-Poudel:
@sweetpalma @alanshaw 
This has been open for long enough. Is it okay if I take over?

Prabhakar-Poudel:
@alanshaw 
Also the `/block/rm` endpoint, unlike other `block/` endpoints, takes one or more multihash of blocks to remove. That implies here too we need to support multiple arguments to `rm`?

alanshaw:
resolved in https://github.com/ipfs/js-ipfs-http-client/pull/1123

PR #925, opened 2019-01-14
chore: update deps

alanshaw:


alanshaw:
Merging as commitlint failure is due to missing dependency in `@commit/lint` - PR here https://github.com/marionebl/commitlint/pull/542

PR #934, opened 2019-01-25
fix: throw on invalid multiaddr to constructor

olizilla:
- dont assume that invalid multiaddrs are valid hosts
- if multiaddrOrHost starts with a /, assume its a multiaddr
- throw if that multiaddr is invalid.

Previously, we'd try and use invalid multiaddrs as the host,
which would lead to requests like http:///dnsaddr/foobar.com:5001/api

see: https://github.com/ipfs-shipyard/ipfs-webui/issues/925

License: MIT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+<a name="29.0.1"></a>
+## [29.0.1](https://github.com/ipfs/js-ipfs-http-client/compare/v29.0.0...v29.0.1) (2019-01-24)
+
+
+### Bug Fixes
+
+* bundle in meteor ([#931](https://github.com/ipfs/js-ipfs-http-client/issues/931)) ([431c442](https://github.com/ipfs/js-ipfs-http-client/commit/431c442)), closes [#10411](https://github.com/ipfs/js-ipfs-http-client/issues/10411)
+
+
+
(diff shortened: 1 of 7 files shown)


olizilla:
@alanshaw ok so, their was a bug that showed we were missing some test cases from the `test/constructor.spec` test suite. I've refactored the constructor more significantly and re-written the `test/constructor.spec` file to be unit test, checking that parameters are set as expected. The existing one often passes in default values for the `host` which means we werent really testing if that value was ever set becuase of the params passed in, or just becuase the default was being used.

To neatly check that the constructor params are updating the endpoint config that ends up being used, I need to additionally expose both the `protocol` and the `api-path` on the `util.getEndpointConfig` method. The other PR https://github.com/ipfs/js-ipfs-http-client/pull/935 already adds `protocol` and it's trivial to add `api-path` too, so I'm goint to update it in that PR. Once that is merged, I can update this PR to ensure that all the new constructor tests pass.

olizilla:
Ready for review. I've updated it to make use of the `protocol` and `apiPath` from `getEndpointConfig` from #935

alanshaw:
@olizilla looks like the browser tests are failing where the port is being detected from the browser URL - I guess karma is setup to run on port 9876.

PR #935, opened 2019-01-25
feat: return protocol from getEndpointConfig

olizilla:
The protocol is used to construct the request, and it's useful to be able to check it has been set correctly. This PR adds it to the `util.getEndpointConfig()` return value.

License: MIT

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ Complete documentation for these methods is coming with: https://github.com/ipfs
 
 > `ipfs.util.getEndpointConfig()`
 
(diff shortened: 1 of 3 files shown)


olizilla:
I have added `api-path` to the getEnpointConfig() return value as we can make use of it in the `test/constructor.spec` to check that the various parameter passing styles are all handled correctly.

See: https://github.com/ipfs/js-ipfs-http-client/pull/934#issuecomment-458089284

alanshaw:
LGTM - macos failures seem unrelated

PR #937, opened 2019-01-31
Update required Node version

filips123:
Readme says that minimum Node.js version is 10. This is true because of some dependencies which also require version 10. However, in `package.json` file, minimum version is 8. This is incorrect, so I'm fixing this now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "through2": "^3.0.0"
   },
   "engines": {
-    "node": ">=8.0.0",
+    "node": ">=10.0.0",
     "npm": ">=3.0.0"
   },
   "repository": {


alanshaw:
CI failure is unrelated.

filips123:
Ok, but don't forget to fix #926 which added Node.js 10 requirement to README.

andremedeiros:
@alanshaw is there anything I can help with bringing engine back to Node 8? It looks like the related PRs have already been merged, and I assume the offending library was taken out.

alanshaw:
@andremedeiros if you can do the work to verify the library is compatible with Node.js 8 and send a PR that would be great!

michaelsbradleyjr:
See PR #996.

PR #942, opened 2019-02-11
perf: use test profile

alanshaw:
Urh, this is kind of amazing, it took the node.js test time down from 13m49.208s to 5m15.992s 😱 (for me locally)

For reference, this is what using the "test" profile does: https://github.com/ipfs/go-ipfs-config/blob/master/profile.go#L68-L85

resolves https://github.com/ipfs/js-ipfs-http-client/issues/593

Depends on:

* [x] https://github.com/ipfs/js-ipfsd-ctl/pull/321

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "eslint-plugin-react": "^7.11.1",
     "go-ipfs-dep": "~0.4.18",
     "interface-ipfs-core": "~0.96.0",
-    "ipfsd-ctl": "~0.40.0",
+    "ipfsd-ctl": "github:ipfs/js-ipfsd-ctl#feat/support-profile",
     "nock": "^10.0.2",
     "pull-stream": "^3.6.9",
     "stream-equal": "^1.1.1"
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "eslint-plugin-react": "^7.11.1",
     "go-ipfs-dep": "~0.4.18",
     "interface-ipfs-core": "~0.96.0",
-    "ipfsd-ctl": "github:ipfs/js-ipfsd-ctl#feat/support-profile",
+    "ipfsd-ctl": "github:ipfs/js-ipfsd-ctl",
     "nock": "^10.0.2",
     "pull-stream": "^3.6.9",
     "stream-equal": "^1.1.1"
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -84,7 +84,7 @@
     "eslint-plugin-react": "^7.11.1",
     "go-ipfs-dep": "~0.4.18",
     "interface-ipfs-core": "~0.96.0",
-    "ipfsd-ctl": "~0.40.0",
+    "ipfsd-ctl": "github:ipfs/js-ipfsd-ctl#feat/support-profile",
```
daviddias: what's missing to have this shipped? :D

alanshaw: Nothing, I just didn't PR yet!

daviddias: I actually missed what this “test profile” was in the beginning. Super rad to see this go-IPFS feature.

Would love to see this on interop tests (and see if it doesn’t break my router anymore).

alanshaw: I should have explained better! For everyone else, you can pass a "profile" option to `ipfs init`:

> 'test' - Reduces external interference of IPFS daemon, this is useful when using the daemon in test environments.
> https://docs.ipfs.io/reference/api/cli/#ipfs-init

Kubuxu:
@alanshaw could you also try: `-p test,badgerds` it uses different datastore that might make it even a bit faster.

Kubuxu:
For reference posting what `test` does also here: https://github.com/ipfs/go-ipfs-config/blob/master/profile.go#L68-L85

alanshaw:
I updated this PR to use the "test" profile in all tests (not just the interface-ipfs-core tests). `npm run test:node` now takes:

✨ 2m43.800s ✨ 

@Kubuxu, I think I might be doing something wrong - passing "test,badgerds" for the test profile makes it slow again - I'm guessing these aren't being applied the way I'm specifying them - do I need multiple `-p` params or should it just work with `ipfs init -p test,badgerds`?

Kubuxu:
Checking it manually, they seem to be both applied when done in this way.
Maybe Badger has a flat overhead (but it scales better). Thanks for trying.'

> 2m43.800s

That is awesome.

alanshaw:
@Kubuxu thank you for the tip off :D

alanshaw:
Seems to be taking between 3-4 mins in Jenkins CI https://ci.ipfs.team/blue/organizations/jenkins/IPFS%2Fjs-ipfs-http-client/detail/PR-942/6/pipeline/257

PR #948, opened 2019-02-21
fix multipart/form-data

MichaelMure:
This PR is the counterpart of https://github.com/ipfs/go-ipfs-files/pull/8

According to https://tools.ietf.org/html/rfc7578#section-4.2, in a Content-Disposition header
of a part in a multipart/form-data body:

- the disposition must be of type `form-data`
- a `name` parameter (not explicitely specified by the RFC, but I suppose non-empty)

As this `name` parameter is unused by IPFS, this commit simply generate a placeholder with
a counter, to conform to the HTTP specification.

I couldn't test properly that this fix actually works properly, so please test thoroughly.

### DIFF
--- a/src/utils/send-files-stream.js
+++ b/src/utils/send-files-stream.js
@@ -7,12 +7,12 @@ const once = require('once')
 const prepareFile = require('./prepare-file')
 const Multipart = require('./multipart')
 
-function headers (file) {
-  const name = file.path
+function headers (file, i) {
+  const filename = file.path
     ? encodeURIComponent(file.path)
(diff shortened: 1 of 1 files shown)


hugomrdias:
for this to work ipfs/multipart needs to support this, or else js-ipfs api will break.

alanshaw:
> for this to work ipfs/multipart needs to support this, or else js-ipfs api will break.

I think it is already supported. https://github.com/ipfs/js-ipfs-multipart/blob/master/src/parser.js#L18

PR #949, opened 2019-02-28
refactor: default to CID v1 and encode with base32

alanshaw:
BREAKING CHANGE: Any v1 CIDs created by this module are base32 encoded by default when stringified.

Depends on:

* [ ] https://github.com/multiformats/js-cid/pull/73


alanshaw:
No longer required - `cids` already defaults to base32 encoding for v1 CIDs

Issue #928, opened 2019-01-16
ipfs.dht.findProvs() can't catch error with fake hash


Dendiom:
We want to implement a 'exist' function to judge if provided hash is exist in the private ipfs network. We use **ipfs.dht.findProvs()** to achieve this logic like this:
```
ipfs.dht.findProvs(hash)
.then(resp => {...})
.catch(e => {...})
```
but we found that if the hash is valid but not exist, it will throw an error that can't catch in promise:
` Uncaught TypeError: Cannot read property 'Type' of undefined`  
after see the source code, we found this:
```
const handleResult = (res, callback) => {
      // Inconsistent return values in the browser vs node
      if (Array.isArray(res)) {
        res = res[0]
      }

      // Type 4 keys
      if (res.Type !== 4) {
        const errMsg = `key was not found (type 4)`

        return callback(errcode(new Error(errMsg), 'ERR_KEY_TYPE_4_NOT_FOUND'))
      }
.....
```
It means that the res is undefined when hash is not exist?
or would you please give us a better way to achieve the 'exist' function

alanshaw:
@Dendiom would you be willing to submit a PR to callback with an empty array if `res` is null?

KrishnaPG:
Looks like this still does not handle the case where `res` is empty array. For example, in one case we are receiving empty array (for whatever reason), and it is causing error:
````
TypeError: Cannot read property 'Type' of undefined
    at handleResult (\node_modules\ipfs-http-client\src\dht\findprovs.js:38:15)
    at streamToValueWithTransformer (\node_modules\ipfs-http-client\src\utils\stream-to-value-with-transformer.js:14:5)
    at send (\node_modules\ipfs-http-client\src\dht\findprovs.js:68:7)
    at f (\node_modules\once\once.js:25:25)
    at concat (\node_modules\ipfs-http-client\src\utils\stream-to-value.js:12:22)
    at ConcatStream.<anonymous> (\node_modules\concat-stream\index.js:32:43)
    at ConcatStream.emit (events.js:187:15)
    at finishMaybe (\node_modules\readable-stream\lib\_stream_writable.js:620:14)
    at endWritable (\node_modules\readable-stream\lib\_stream_writable.js:629:3)
    at ConcatStream.Writable.end (\node_modules\readable-stream\lib\_stream_writable.js:567:22)
````
[...]

PR #950, opened 2019-03-02
fix: dht.findProvs.js handle valid hash but no providers

niinpatel:
This is a fix for #928

### DIFF
--- a/src/dht/findprovs.js
+++ b/src/dht/findprovs.js
@@ -23,6 +23,12 @@ module.exports = (send) => {
     }
 
     const handleResult = (res, callback) => {
+      // callback with an empty array if no providers are found
+      if (!res) {
+        const responses = []
+        return callback(null, responses)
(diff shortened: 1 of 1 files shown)


src/dht/findprovs.js
```diff
@@ -23,6 +23,12 @@ module.exports = (send) => {
     }
 
     const handleResult = (res, callback) => {
+      // callback with an empty array if no providers are found
+      if (res === null) {
```
alanshaw: According to the issue, `res` is `undefined` so this change won't resolve the issue. Suggestion below (which will catch both `null` _and_ `undefined`):

```suggestion
      if (res == null) {
```

A small unit test to assert this, that mocks out `send` would have caught this!

niinpatel: or instead of double equals, we can do `if (!res) { ... }`

PR #951, opened 2019-03-05
refactor: export types and utilities statically

### LINKED ISSUE #902: RFC: expose types and utils in static context
> I just found [this](https://github.com/ipfs/js-ipfs-http-client/issues/406), I'm curious why it was defined based on an instance of ipfs, rather than something static since it points at the statically defined types.
> 
> My specific use case is to do something like `foo instanceof ipfs.types.CID` but it happens in static functions or places where the ipfs instance isn't passed as a variable, i.e. where I don't easily have access to the "ipfs" variable, but have easy access to the result of 
> `IPFS = require('ipfs')`, 
> 
> The current code does `CID=require('cids')` solely to allow this type check, which seems exactly the kind of example this change was intended to avoid.
> 
> _Originally posted by @mitra42 in https://github.com/ipfs/js-ipfs-http-client/issues/406#issuecomment-443955150_

I remember having similar thoughts at least once (had to make fake instance just to access utils).
 
Perhaps we could at least support it in the static context by [exposing data types and util functions](https://github.com/ipfs/js-ipfs-http-client/blob/v27.0.0/README.md#domain-data-types) on `IPFS` (the static constructor) as well.

Thoughts?

alanshaw:
Allows users to access these additional types and utilities without having to create an instance of the client first.

resolves #902
refs https://github.com/ipfs/js-ipfs/pull/1908

BREAKING CHANGE: `ipfs.util.isIPFS` has moved to a static export and should be accessed via `const { isIPFS } = require('ipfs-http-client')`.

The modules available under `ipfs.types.*` have also become static exports.

`ipfs.util.crypto` has been removed as it is not a dependency of `ipfs-http-client` so reduces the bundle size. If you need to use libp2p crypto primitives then please see the [js-libp2p-crypto](https://github.com/libp2p/js-libp2p-crypto) project for info on how to use it in your project.

Finally `ipfs.util.getEndpointConfig` is now a direct instance method, `ipfs.getEndpointConfig`


README.md
```diff
@@ -348,41 +348,32 @@ This means:
...
+- `ipfs.getEndpointConfig()`
 
-- [`ipfs.types.Buffer`](https://www.npmjs.com/package/buffer)
-- [`ipfs.types.PeerId`](https://github.com/libp2p/js-peer-id)
-- [`ipfs.types.PeerInfo`](https://github.com/libp2p/js-peer-info)
-- [`ipfs.types.multiaddr`](https://github.com/multiformats/js-multiaddr)
-- [`ipfs.types.multibase`](https://github.com/multiformats/multibase)
-- [`ipfs.types.multihash`](https://github.com/multiformats/js-multihash)
-- [`ipfs.types.CID`](https://github.com/ipld/js-cid)
+Call this on your client instance to return an object containing the `host`, `port` and `protocol`.
```
lidel:
```suggestion
Call this on your client instance to return an object containing the `host`, `port`, `protocol` and `api-path`.
```

PR #954, opened 2019-03-13
feat: provide access to multicodec

niinpatel:
resolves: #955 

1. Included `multicodec` library in the static exports.
2. Removed `src/types.js`. I don't think it is needed anymore because we're exporting the utilities and types statically and not with the ipfs object. (#951 )

3. Also, I think we should also remove the tests for `ipfs.types.` in the [`interface-ipfs-core`](https://github.com/ipfs/interface-js-ipfs-core/blob/master/src/types.js) as they are no longer used. Should I also make a PR there for it? 

License: MIT
Signed-off by: Nitin Patel <nitinpatel278@gmail.com>


niinpatel:
> Perfect thanks ✨
> 
> > Also, I think we should also remove the tests for ipfs.types. in the interface-ipfs-core as they are no longer used. Should I also make a PR there for it?
> 
> Yes please!

Here it is: https://github.com/ipfs/interface-js-ipfs-core/pull/450

PR #966, opened 2019-04-09
fix: fix missing buffer bundling with browserify

### LINKED ISSUE #964: Browserify Buffer is not defined
Hello,
I noticed on one of my project that current version of js-ipfs-http-client doesn't work in browser (using Browserify). The same error happens when I try to load example browserify project from https://github.com/ipfs/js-ipfs-http-client/tree/master/examples/bundle-browserify . After page load I get this error
```
bundle.js:26254 Uncaught ReferenceError: Buffer is not defined
    at Object.191../utils/default-config (bundle.js:26254)
    at o (bundle.js:1)
    at bundle.js:1
    at Object.<anonymous> (bundle.js:5)
    at Object.1.buffer (bundle.js:40)
    at o (bundle.js:1)
    at r (bundle.js:1)
    at bundle.js:1
```
That happens on this line
```
Object.assign(module.exports, { isIPFS, Buffer, CID, multiaddr, multibase, multicodec, multihash, PeerId, PeerInfo })
```
When I tried to include `buffer` package manually to the project it had no effect. Does anybody know what is wrong?

hugomrdias:
Closes #964

We should backport this to js-ipfs.

Explicit require of buffer can actually be enforce by linter to avoid these and other issues with node/browser stuff.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "bignumber.js": "^8.0.2",
     "bl": "^3.0.0",
     "bs58": "^4.0.1",
+    "buffer": "^5.2.1",
     "cids": "~0.5.5",
     "concat-stream": "hugomrdias/concat-stream#feat/smaller",
     "debug": "^4.1.0",
--- a/src/index.js
+++ b/src/index.js
@@ -2,6 +2,7 @@
 /* global self */
 
 const isIPFS = require('is-ipfs')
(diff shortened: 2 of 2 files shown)


alanshaw:
Would you please add `buffer` to dependencies @hugomrdias?

Issue #968, opened 2019-04-12
Babel error in upload-file-via-browser example


autonome:
Follow the readme directions, and I see this:

MacBook:upload-file-via-browser dietrich$ npm start

> upload-file-via-browser@1.0.0 start /Users/dietrich/misc/js-ipfs-http-client/examples/upload-file-via-browser
> node server.js

Listening at localhost:3000
✖ ｢wdm｣: Hash: a12059f5b30f6dcc8c8e
Version: webpack 4.25.1
Time: 566ms
Built at: 04/12/2019 3:30:42 PM
    Asset     Size  Chunks             Chunk Names
bundle.js  379 KiB    main  [emitted]  main
Entrypoint main = bundle.js
[0] multi webpack-dev-server/client?http://localhost:3000 webpack/hot/only-dev-server ./src/index 52 bytes {main} [built]
[./node_modules/ansi-html/index.js] 4.16 KiB {main} [built]
[./node_modules/ansi-regex/index.js] 135 bytes {main} [built]
[./node_modules/loglevel/lib/loglevel.js] 7.68 KiB {main} [built]
[./node_modules/strip-ansi/index.js] 161 bytes {main} [built]
[./node_modules/url/url.js] 22.8 KiB {main} [built]
[./node_modules/webpack-dev-server/client/index.js?http://localhost:3000] (webpack)-dev-server/client?http://localhost:3000 7.78 KiB {main} [built]
[./node_modules/webpack-dev-server/client/overlay.js] (webpack)-dev-server/client/overlay.js 3.58 KiB {main} [built]
[./node_modules/webpack-dev-server/client/socket.js] (webpack)-dev-server/client/socket.js 1.05 KiB {main} [built]
[./node_modules/webpack/hot sync ^\.\/log$] (webpack)/hot sync nonrecursive ^\.\/log$ 170 bytes {main} [built]
[./node_modules/webpack/hot/emitter.js] (webpack)/hot/emitter.js 75 bytes {main} [built]
[./node_modules/webpack/hot/log-apply-result.js] (webpack)/hot/log-apply-result.js 1.27 KiB {main} [built]
[./node_modules/webpack/hot/log.js] (webpack)/hot/log.js 1.11 KiB {main} [built]
[./node_modules/webpack/hot/only-dev-server.js] (webpack)/hot/only-dev-server.js 2.55 KiB {main} [built]
[./src/index.js] 1 KiB {main} [not cacheable] [built] [failed] [1 error]
[...]

PR #970, opened 2019-04-15
Fix: Babel error in upload-file-via-browser example #968

autonome:
Update to latest Babel fixed ze problem.


autonome:
@alanshaw this codecov problem looks unrelated to my change. any ideas?

alanshaw:
Thanks @autonome ❤️

PR #973, opened 2019-04-17
chore: update dependencies

alanshaw:


olizilla:
The test that is failing is

```
  1) interface-ipfs-core tests
       .dht.provide
         should allow multiple CIDs to be passed:
     Uncaught TypeError: Cannot read property 'hash' of undefined
      at ipfs.add (node_modules/interface-ipfs-core/src/dht/provide.js:68:26)
      at f (node_modules/once/once.js:25:25)
      at result (src/files-regular/add.js:7:642)
```

In that test we call ipfs.add with an array of 2 buffers, and expect the result to contain an array of 2 ipfs file references, but it only contains 1.

https://github.com/ipfs/interface-js-ipfs-core/blob/fa21abb5ed9ff379e3ac5fb517592f8ea3acb1d7/src/dht/provide.js#L62-L74

```js
ipfs.add([Buffer.from('t0'), Buffer.from('t1')], (err, res) => {
        if (err) return done(err)
        console.log({res})
        ipfs.dht.provide([
          new CID(res[0].hash),
          new CID(res[1].hash)
```

`Cannot read property 'hash' of undefined` is coming from the `new CID(res[1].hash)` line, as `res[1]` does not exit. `res[0]` does exist and has a hash property as expected. Why are we only getting 1 result back and not 2?

olizilla:
hmm, if i edit the test `should add a Buffer as tuple` to pass an array of more than 1 thing, it also fails

**Fails** - `filesAdded.length === 1` not 2 as expected.
```js
    it('should add a Buffer as tuple', (done) => {
      ipfs.add([
        { path: 'testfile.txt', content: fixtures.smallFile.data },
        { path: 'other.txt', content: fixtures.bigFile.data }
      ], (err, filesAdded) => {
        expect(err).to.not.exist()
        expect(filesAdded).to.have.length(2)
        done()
      })
    })
```

olizilla:
Dropping back to `go-ipfs-dep@0.4.19` doesn't change anything. Test still fails. Time to go deeper.

olizilla:
Spoke to soon. Switching back to `go-ipfs-dep@0.4.19` does cause the test to pass. Investigating 🔎

olizilla:
Cant add multiple files via the CLI in go-ipfs@0.4.20 https://github.com/ipfs/go-ipfs/issues/6254

PR #978, opened 2019-04-24
refs endpoint

dirkmc:
- make refs endpoint compatible with js-ipfs
- add refs pull stream
- move refs test to [interface-js-ipfs-core](https://github.com/ipfs/interface-js-ipfs-core) (see https://github.com/ipfs/interface-js-ipfs-core/pull/460)


dirkmc:
@lidel I didn't see it used anywhere so I didn't add it. Do you know if it will be needed?

FYI I am refactoring this PR heavily to
- move most of the tests into interface-ipfs-core, so they can be used to test against go and js
- traverse the refs DAG in js-ipfs instead of relying on js-ipfs-unix-exporter, which is specifically aimed at unix format files (turns out this code is quite simple to implement with [pull-traverse](https://pull-stream.github.io/#pull-traverse))

dirkmc:
I added `refsReadableStream` in case it will be needed

lidel:
Thanks! FYSA we always provide both pull-stream and readable-stream versions (see [interface-js-ipfs-core/SPEC/FILES](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md)), and if streaming is supported users expect to find both.

test/sub-modules.spec.js
```diff
@@ -183,6 +183,9 @@ describe('submodules', () => {
     expect(filesRegular.ls).to.be.a('function')
     expect(filesRegular.lsReadableStream).to.be.a('function')
     expect(filesRegular.lsPullStream).to.be.a('function')
+    expect(filesRegular.refs).to.be.a('function')
+    expect(filesRegular.refs.local).to.be.a('function')
+    expect(filesRegular.refsPullStream).to.be.a('function')
```
lidel: missing `refsReadableStream`  ?

dirkmc: Good catch, thanks! 👍

src/files-regular/refs-local-pull-stream.js
```diff
@@ -0,0 +1,26 @@
...
+
+module.exports = (send) => {
+  send = moduleConfig(send)
+
+  return (opts) => {
+    opts = opts || {}
+
+    const p = deferred.source()
+
+    send({ path: 'refs/local', qs: opts }, (err, stream) => {
```
alanshaw: `maxDepth` option name here needs to be converted to `max-depth` querystring param name.

...and for the other endpoints.

src/files-regular/refs.js
```diff
@@ -0,0 +1,48 @@
...
+  const send = moduleConfig(arg)
+
+  const refs = promisify((args, opts, callback) => {
+    if (typeof (opts) === 'function') {
+      callback = opts
+      opts = {}
+    }
+
+    try {
+      args = cleanCID(args)
```
alanshaw: Does `cleanCID` take into account this can be an array of paths?

src/files-regular/refs-local-pull-stream.js
```diff
@@ -0,0 +1,26 @@
...
+      if (err) { return p.resolve(pull.error(err)) }
+
+      p.resolve(toPull.source(stream))
```
alanshaw: Needs to convert response objects from PascalCase to camelCase i.e.

`{ Ref, Err }` to `{ ref, err }`

alanshaw:
@dirkmc I think this just needs a rebase to get the tests passing - go-ipfs 0.4.20 is broken for adding files so we've reverted to testing against 0.4.19.

dirkmc:
@alanshaw thanks, rebased

PR #980, opened 2019-04-26
fix: Uncaught Error when adding files bigger than 1MB

### LINKED ISSUE #967: add with 1MB file triggers Uncaught Error: stream.push() after EOF
Uploading a file > 1MB, as a pull-stream fails, with the error:

```
Uncaught Error: stream.push() after EOF
    at new NodeError (errors-browser.js:30)
    at readableAddChunk (_stream_readable.js:269)
    at PullDuplexStream.push../node_modules/readable-stream/lib/_stream_readable.js.Readable.push (_stream_readable.js:240)
    at next (index.js:55)
    at FileReader.loaded (index.js:22)
```

The issue is occuring in master on webui.ipfs.io since upgrading v30.1.2 from v29.1.0

See: https://github.com/ipfs-shipyard/ipfs-webui/issues/1010

lidel:
This switches pull-to-stream to v0.1.1 which includes a fix `Uncaught Error: stream.push() after EOF` and truly closes #967 in all build pipelines (semver seems to be  not enough when yarn lockfile is used)

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "promisify-es6": "^1.0.3",
     "pull-defer": "~0.2.3",
     "pull-stream": "^3.6.9",
-    "pull-to-stream": "~0.1.0",
+    "pull-to-stream": "~0.1.1",
     "pump": "^3.0.0",
     "qs": "^6.5.2",
     "readable-stream": "^3.1.1",


alanshaw:
@lidel I've rebased for green CI

PR #992, opened 2019-05-10
GC endpoint

dirkmc:
BREAKING CHANGE: repo.gc response objects have changed to `{ err, cid }`

Depends on:

* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/462

### DIFF
--- a/src/repo/gc.js
+++ b/src/repo/gc.js
@@ -2,11 +2,12 @@
 
 const promisify = require('promisify-es6')
 const streamToValueWithTransformer = require('../utils/stream-to-value-with-transformer')
+const CID = require('cids')
 
 const transform = function (res, callback) {
   callback(null, res.map(r => ({
-    err: r.Err,
-    cid: (r.Key || {})['/']
+    err: r.Err ? new Error(r.Err) : null,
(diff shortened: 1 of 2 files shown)


src/repo/gc.js
```diff
@@ -1,16 +1,32 @@
 'use strict'
 
 const promisify = require('promisify-es6')
+const streamToValueWithTransformer = require('../utils/stream-to-value-with-transformer')
+
+const transform = function (res, callback) {
+  callback(null, res.map(r => ({
+    err: r.Err,
```
alanshaw:
```suggestion
    err: r.Err ? new Error(r.Err) : null,
```

src/repo/gc.js
```diff
@@ -1,16 +1,32 @@
...
+    cid: (r.Key || {})['/']
```
alanshaw:
```suggestion
    cid: r.Key ? new CID(r.Key) : null
```

dirkmc:
repo.gc() docs fix: https://github.com/ipfs/interface-js-ipfs-core/pull/492

PR #996, opened 2019-05-14
build: change min node version to 8.3.0

### LINKED ISSUE #983: Should the required NodeJS version be >=8.0.0?
See: https://github.com/ipfs/js-ipfs-http-client/pull/937#issuecomment-486422623.

michaelsbradleyjr:
Also adjust Travis CI config to perform builds and tests with node 8.

`npm install`, `npm run build`, and `npm run test` all succeeded locally with node version `8.3.0`.

The reason for version `8.3.0` instead of `8.0.0` is that `8.3.0` introduced support for object rest/spread, which is used by at least one of this package's devDependencies (`go-ipfs-dep`, in production install of that devDep). It may be the case that at runtime older versions of node would work correctly with the built package, but it seems simpler to settle on the minimum supported version being a version that can build, test, and use the package.

Closes #983.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ stages:
 
 node_js:
   - '10'
+  - '8'
 
 os:
   - linux
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "through2": "^3.0.1"
   },
   "engines": {
-    "node": ">=10.0.0",
+    "node": ">=8.3.0",
     "npm": ">=3.0.0"
   },
   "repository": {


michaelsbradleyjr:
I'm not sure how to adjust the Travis CI config so that tests are run with node version 8 on Windows.

I'm also not sure, if this PR is to be merged, whether it should include changes to the README to the effect that the Maintenance LTS of node is also supported.

filips123:
@alanshaw @michaelsbradleyjr There is again problem with support for Node.js 8. It now fails because of `peer-id` which requires Node.js 10. See [my CI test](https://travis-ci.com/filips123/EthAvatar.JS/jobs/215604668) for details.

michaelsbradleyjr:
@filips123, indeed. The problem is especially apparent when you use `yarn` instead of `npm` in a project that has `ipfs-http-client` as a dependency because `yarn` always enforces `"engines"` at any depth in the dependency tree whereas `npm` does not.

So, in hindsight, this PR was actually mostly pointless and should probably be reverted.

PR #997, opened 2019-05-15
Fix error handling for refs/refs local

dirkmc:


src/utils/stream-to-value.js
```diff
@@ -7,14 +7,12 @@ const concat = require('concat-stream')
...
     response,
-    concat((data) => callback(null, data)),
-    (err) => {
-      if (err) {
-        callback(err)
-      }
-    })
+    concat((d) => { data = d }),
+    (err) => callback(err, data)
+  )
```
alanshaw: Good catch!

PR #1001, opened 2019-05-16
JSON.parse error reporting: with `invalid json:` in the error string

### LINKED ISSUE #912: Better handling for unexpected responses
[Issues like this](https://github.com/ipfs/js-ipfs-http-client/issues/905) will continue to be reported until we can communicate better to the user the cause of the error and/or the response from the server.

### LINKED ISSUE #1000: JSON.parse(data) errors
It looks like the IPFS gateways return plain text errors, while the clients are expecting JSON, which is causing the [json.parse](https://github.com/ipfs/js-ipfs-http-client/blob/master/src/utils/stream-to-json-value.js#L25) to throw errors many times, (such as `404 not found`, `method not allowed` etc.)

The problem is, while the `json.parse` [line 27](https://github.com/ipfs/js-ipfs-http-client/blob/master/src/utils/stream-to-json-value.js#L27) throws the error it discards the original data, leaving caller have no clue why the parse failed.

For example, here is one error received on the client side returned by IPFS deamon: if you put a breakpoint and observe the data, it has some valuable information such as: `argument "key" is required`, which gets lost when the exception is thrown at line 27

image

The rethrown error at the call site contains stack with incorrect (and useless) error message:
````
"SyntaxError: Unexpected token a in JSON at position 0
    at JSON.parse (<anonymous>)
    at streamToValue (\node_modules\ipfs-http-client\src\utils\stream-to-json-value.js:25:18)
    at concat (\node_modules\ipfs-http-client\src\utils\stream-to-value.js:12:22)
    at ConcatStream.<anonymous> (\node_modules\concat-stream\index.js:32:43)
    at ConcatStream.emit (events.js:187:15)
    at finishMaybe (\node_modules\readable-stream\lib\_stream_writable.js:620:14)
    at afterWrite (\node_modules\readable-stream\lib\_stream_writable.js:466:3)
    at process._tickCallback (internal/process/next_tick.js:63:19)"
````
Suggestion:
   1. If  clients are expecting JSON data, the gateways should return errors in JSON format, Else
   2. on the client side, e.g. at line 27, please include the original data received from the server in the error thrown so that callers get to know what the problem is, something like:
````
    try {
      res = JSON.parse(data)
    } catch (err) {
      return cb(new Error(data))   // <---- throw the data part of the error
    }
````
or something like `return cb(new Error("Invalid JSON: ${data}"))`

For beginners, being able to understand the server errors is very important, since it is not always possible get things right the first time.

### PR DESCRIPTION
**Pull request by KrishnaPG**
Better error reporting in case of `json.parse` failures

resolves #912
resolves #1000

### DIFF
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -13,9 +13,21 @@ const log = require('debug')('ipfs-http-client:request')
 
 // -- Internal
 
-function parseError (res, cb) {
+function parseError (res, cb, isJson = true) {
   const error = new Error(`Server responded with ${res.statusCode}`)
 
+  if (!isJson) {
+    return streamToValue(res, (err, data) => {
+      // the `err` here refers to errors in stream processing, which
+      // we ignore here, since we already have a valid `error` response
+      // from the server above that we have to report to the caller.
+      if (data) {
+        error.message = data.toString()
+      }
+      cb(error)
+    })
+  }
+  
   streamToJsonValue(res, (err, payload) => {
     if (err) {
       return cb(err)
@@ -44,7 +56,7 @@ function onRes (buffer, cb) {
     }
 
     if (res.statusCode >= 400 || !res.statusCode) {
-      return parseError(res, cb)
+      return parseError(res, cb, isJson)
     }
 
     // Return the response stream directly
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -13,10 +13,15 @@ const log = require('debug')('ipfs-http-client:request')
 
 // -- Internal
 
-function parseError (res, cb, isJson = true) {
+function hasJSONHeaders (res) {
+  return res.headers['content-type'] &&
+    res.headers['content-type'].indexOf('application/json') === 0
+}
+
+function parseError (res, cb) {
   const error = new Error(`Server responded with ${res.statusCode}`)
 
-  if (!isJson) {
+  if (!hasJSONHeaders(res)) {
     return streamToValue(res, (err, data) => {
       // the `err` here refers to errors in stream processing, which
       // we ignore here, since we already have a valid `error` response
@@ -46,8 +51,7 @@ function onRes (buffer, cb) {
   return (res) => {
     const stream = Boolean(res.headers['x-stream-output'])
     const chunkedObjects = Boolean(res.headers['x-chunked-output'])
-    const isJson = res.headers['content-type'] &&
-                   res.headers['content-type'].indexOf('application/json') === 0
+    const isJson = hasJSONHeaders(res)
 
     if (res.req) {
       log(res.req.method, `${res.req.getHeaders().host}${res.req.path}`, res.statusCode, res.statusMessage)
@@ -56,7 +60,7 @@ function onRes (buffer, cb) {
     }
 
     if (res.statusCode >= 400 || !res.statusCode) {
-      return parseError(res, cb, isJson)
+      return parseError(res, cb)
     }
 
     // Return the response stream directly
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -32,7 +32,7 @@ function parseError (res, cb) {
       cb(error)
     })
   }
-  
+
   streamToJsonValue(res, (err, payload) => {
     if (err) {
       return cb(err)
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -29,6 +29,7 @@ function parseError (res, cb) {
       if (data && data.length) {
         error.message = data.toString()
       }
+      error.code = res.statusCode
       cb(error)
     })
   }
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -21,7 +21,7 @@ function parseError (res, cb, isJson = true) {
       // the `err` here refers to errors in stream processing, which
       // we ignore here, since we already have a valid `error` response
       // from the server above that we have to report to the caller.
-      if (data) {
+      if (data && data.length) {
         error.message = data.toString()
       }
       cb(error)
--- a/src/utils/stream-to-json-value.js
+++ b/src/utils/stream-to-json-value.js
@@ -24,7 +24,7 @@ function streamToJsonValue (res, cb) {
     try {
       res = JSON.parse(data)
     } catch (err) {
-      return cb(err)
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by hugomrdias** [comment]
@alanshaw this is only failing on commitlint should be good to merge

**Comment by alanshaw** [comment]
I think @KrishnaPG is going to update this to be even better ;)

https://github.com/ipfs/js-ipfs-http-client/issues/1000#issuecomment-493087784

File: src/utils/send-request.js [hunk 1]
```diff
@@ -13,9 +13,21 @@ const log = require('debug')('ipfs-http-client:request')
 
 // -- Internal
 
-function parseError (res, cb) {
+function parseError (res, cb, isJson = true) {
```
alanshaw: `isJson` should be required and passed before `cb` - it's very unusual for a function that takes a callback for it to be passed in any position other than the last.

```suggestion
function parseError (res, isJson, cb) {
```

src/utils/send-request.js
```diff
@@ -13,9 +13,21 @@ const log = require('debug')('ipfs-http-client:request')
...
   const error = new Error(`Server responded with ${res.statusCode}`)
 
+  if (!isJson) {
+    return streamToValue(res, (err, data) => {
+      // the `err` here refers to errors in stream processing, which
+      // we ignore here, since we already have a valid `error` response
+      // from the server above that we have to report to the caller.
+      if (data) {
```
alanshaw: So we don't get a blank error message!

```suggestion
      if (data && data.length) {
```

src/utils/send-request.js
```diff
@@ -44,7 +56,7 @@ function onRes (buffer, cb) {
     }
 
     if (res.statusCode >= 400 || !res.statusCode) {
-      return parseError(res, cb)
+      return parseError(res, cb, isJson)
```
alanshaw: See previous comment for explanation :)

```suggestion
      return parseError(res, isJson, cb)
```

src/utils/send-request.js
KrishnaPG: Changing the parameter positions would be a breaking change. Is that ok? Is there no where else this `parseError` method being called currently? We may have to change it in all places.

If we look into it, this `isJson` value is actually being computed from the `res`, which itself is being passed as first parameter - so I have no objection to *deduce* the `isJson` again from the `res` inside the `parseError`. That way we would not need to pass the `isJson` as additional parameter (and also avoids the breaking change).

Now, to avoid the code-duplication of computing the `isJson` from `res`, I can extract it into a new method, something like:
````
function isJson(res) {
    return res.headers['content-type'] && res.headers['content-type'].indexOf('application/json') === 0;
}
function parseError(res, cb) {
    if(!isJson(res)) return streamToValue(...);
    streamToJsonValue(...);
}
````
This puts the burden of checking the data on the method itself, rather than on the caller (who should not worry about how the `parseError` does its job).

Though, I initially had the temptation to *fix* the `streamToJsonValue` itself to make it foolproof (instead of modifying `parseError`), since `streamToJsonValue` is internally calling `streamToValue` anyway. After all, `parseError` may not be the only client the `streamToJsonValue` may have, and every one cannot keep checking which variation of `streamToJsonValue` or `streamToValue` should be called in every place. Instead, the `streamToJsonValue` calls `JSON.parse(data)` only for what it believes as JSON data, and behaves like `streamToValue` in all other, non-JSON cases. For example, the below code works perfectly fine without having to modify `parseError` or any other existing `streamToJsonValue` calls:
````
function streamToJsonValue (res, cb) {
  streamToValue(res, (err, data) => {

   ...

    let res = data    // <- proposed change
    try {
      res = JSON.parse(data)
    } catch (err) {
      ; // <- proposed change, no error callback
    }

    cb(null, res)  //<- this will send `data` itself as the result
  })
}
````
This should not break any existing *working* code + more resilient. The `parseError` implementation need not be touched at all (i.e. no need for `isJson` checks, since `streamToJsonValue` takes care of it), though not sure how compatible this behavior is with overall design.

Let me know which approach should be taken forward, will modify accordingly.

src/utils/send-request.js
KrishnaPG: Yes - had that nagging feeling `data` should be non-empty (in addition to being non-null), but for some reason was hesitant to assume the `.length` property exists on the `data` object in all valid cases. The hesitation came from this [Line 19](https://github.com/ipfs/js-ipfs-http-client/blob/master/src/utils/stream-to-json-value.js#L19) where data is being checked if it is a buffer: `if (Buffer.isBuffer(data))`

But now that I think about it, you are right. Even for empty strings the `.length` is a valid value in Javascript (and confirmed it in Chrome console). Committing your suggestion. Hope `buffer` and `string` are the only two cases the `data` can be of type.

src/utils/send-request.js
alanshaw: I'm happy for `isJson` to be extracted into ta function and be called again from `parseError` 👍

`parseError` isn't exported from this module AFAIK so this is the only place that is using it.

Please leave `streamToJsonValue` as it is in this PR - IMHO it's unexpected for it to silently ignore a parse error and return a value that hasn't been parsed as JSON.

KrishnaPG: Done. The local variable `isJson` was clashing with the function name, so named the method as `hasJSONHeaders`.

Reverted back the `parseError` method signature to its earlier form. Kept the conditional routing to `streamToJsonValue`

KrishnaPG:
The remaining _lint_ error is about handling the error in callback
````
/home/travis/build/ipfs/js-ipfs-http-client/src/utils/send-request.js
   25:31  error    Expected error to be handled  handle-callback-err
````
By design we are ignoring that error, since we are already inside an error handler (with a valid error response from server).

Is there a way to turn this lint error off, for this specific case, so that this can be merged?

KrishnaPG:
The changes in this PR now enables usage such as:
````
ipfs.block.put(Buffer.from("Hello " + Date()), (err, block) => {
  if (err) {
     if (err.code == 401 || err.code == 403) 
        return Retry_With_Auth(block); // retry this whole thing with renewed AUTH Headers somehow
  }
  // success
})
````
Earlier this was not possible, since the textual response from gateways was crashing the app.

alanshaw:
@KrishnaPG I've rebased, added tests and changed `code` to `statusCode` (since it conflicts with the code in the JSON response) here: https://github.com/ipfs/js-ipfs-http-client/pull/1016

PR #1002, opened 2019-05-16
JSON.parse error reporting: without `invalid json:` in the error string

KrishnaPG:
Better Error reporting for `json.parse` errors with `data` in the error


alanshaw:
See https://github.com/ipfs/js-ipfs-http-client/issues/1000#issuecomment-493087784

PR #1016, opened 2019-05-21
fix: error reporting for non-JSON responses

### LINKED ISSUE #912: Better handling for unexpected responses

### LINKED ISSUE #1000: JSON.parse(data) errors

alanshaw:
Better error reporting by detecting `Content-Type` of the response and not attempting to parse JSON for non-`application/json` responses.

resolves #912
resolves #1000
closes #1001

### DIFF
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -20,16 +20,16 @@ function hasJSONHeaders (res) {
 
 function parseError (res, cb) {
   const error = new Error(`Server responded with ${res.statusCode}`)
+  error.statusCode = res.statusCode
 
   if (!hasJSONHeaders(res)) {
-    return streamToValue(res, (err, data) => {
+    return streamToValue(res, (err, data) => { // eslint-disable-line handle-callback-err
       // the `err` here refers to errors in stream processing, which
       // we ignore here, since we already have a valid `error` response
       // from the server above that we have to report to the caller.
       if (data && data.length) {
         error.message = data.toString()
       }
-      error.code = res.statusCode
       cb(error)
     })
   }
--- a/src/utils/send-request.js
+++ b/src/utils/send-request.js
@@ -13,9 +13,21 @@ const log = require('debug')('ipfs-http-client:request')
 
 // -- Internal
 
-function parseError (res, cb) {
+function parseError (res, cb, isJson = true) {
   const error = new Error(`Server responded with ${res.statusCode}`)
 
+  if (!isJson) {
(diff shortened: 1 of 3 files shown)


alanshaw:
PR build is failing because of commit lint in earlier commits. Branch build is passing. Lets get this merged! Thanks again @KrishnaPG

PR #1026, opened 2019-06-01
feat(block.rm): add block remove functionality

### LINKED ISSUE #792: block.rm() is missing :(

Prabhakar-Poudel:
docs: update readme

fixes #792

License: MIT
Signed-off-by: Prabhakar-Poudel <yuvrajzohan@gmail.com>

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -5,7 +5,7 @@ const createServer = require('ipfsd-ctl').createServer
 const server = createServer()
 
 module.exports = {
-  bundlesize: { maxSize: '236kB' },
+  bundlesize: { maxSize: '250kB' },
   webpack: {
     resolve: {
       mainFields: ['browser', 'main']
--- a/README.md
+++ b/README.md
@@ -34,17 +34,28 @@
 
 ## Table of Contents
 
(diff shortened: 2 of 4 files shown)


Prabhakar-Poudel:
@alanshaw 

> FAIL  ./dist/index.min.js: 243.68KB > maxSize 236KB (gzip)

Bundle size has exceeded the maximum. How can I resolve this?

achingbrain:
`maxSize` is set in [`.aegir.js`](https://github.com/ipfs/js-ipfs-http-client/blob/master/.aegir.js#L8), you may just need to increase that.

Prabhakar-Poudel:
> maxSize is set in .aegir.js, you may just need to increase that.

Will that break other parts of codebase?
And if we keep it increasing every so often, why not have a decent enough of an estimate?
@achingbrain

achingbrain:
This was implemented in #1123.  Sorry, I forgot this was here!

PR #1038, opened 2019-07-10
chore: use test profile for swarm tests

alanshaw:
Saves 1 minute running them locally.


alanshaw:
Tests are all still passing...

PR #1039, opened 2019-07-10
fix: response for findpeer and findprovs

alanshaw:
Pick out the correct item from the response, do not assume the first is the one we want.

Also calls `toString` on the CID/PeerID so that we can pass CID instances directly to these functions.

### DIFF
--- a/src/dht/findpeer.js
+++ b/src/dht/findpeer.js
@@ -6,7 +6,7 @@ const streamToValueWithTransformer = require('../utils/stream-to-value-with-tran
 const multiaddr = require('multiaddr')
 const PeerId = require('peer-id')
 const PeerInfo = require('peer-info')
-const errcode = require('error-code')
+const errcode = require('err-code')
 
 module.exports = (send) => {
   return promisify((peerId, opts, callback) => {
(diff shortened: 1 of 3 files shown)


src/dht/findpeer.js
```diff
@@ -25,14 +25,15 @@ module.exports = (send) => {
...
       // Type 2 keys
-      if (res.Type !== 2) {
-        const errMsg = `key was not found (type 2)`
-
-        return callback(errcode(new Error(errMsg), 'ERR_KEY_TYPE_2_NOT_FOUND'))
+      // 2 = FinalPeer
+      // https://github.com/libp2p/go-libp2p-core/blob/6e566d10f4a5447317a66d64c7459954b969bdab/routing/query.go#L18
+      if (!res || res.Type !== 2) {
+        const errMsg = `key was not found (type 4)`
+        return callback(errcode(new Error(errMsg), 'ERR_KEY_TYPE_4_NOT_FOUND'))
```
dirkmc: Should this be `ERR_KEY_TYPE_2_NOT_FOUND`

PR #1041, opened 2019-07-10
fix: make findprovs return all responses

jacobheun:
Find providers was not consistently returning all of the provider responses. This was causing flakey test failures for https://github.com/libp2p/js-libp2p-delegated-content-routing/pull/9.

It will now properly return all response records for each result. 

* [x] requires https://github.com/ipfs/interface-js-ipfs-core/pull/493

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "cross-env": "^5.2.0",
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "~0.4.21",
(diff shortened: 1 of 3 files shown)


alanshaw:
Code change LGTM - can we get a test in interface-ipfs-core that tests for multiple provider results? 🙏🙏🙏

jacobheun:
Added the findProvs test over at https://github.com/ipfs/interface-js-ipfs-core/pull/493. I also removed the skip here, as it wasnt running findProvs at all.

PR #1055, opened 2019-07-18
Added opts to object stat command

obo20:
The ipfs object stat command couldn't handle the {timeout: '1s'} option (or any other options). We're already checking for the opts parameter so it makes sense that we should be sending it as well. 

Things work as desired now with this one line addition.

Depends on:

* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/506

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "cross-env": "^5.2.0",
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "~0.4.21",
-    "interface-ipfs-core": "^0.107.1",
+    "interface-ipfs-core": "github:ipfs/interface-js-ipfs-core#fix/resolve-ipns-test",
     "ipfsd-ctl": "~0.43.0",
     "nock": "^10.0.2",
     "stream-equal": "^1.1.1"
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "cross-env": "^5.2.0",
     "dirty-chai": "^2.0.1",
     "go-ipfs-dep": "~0.4.21",
-    "interface-ipfs-core": "github:ipfs/interface-js-ipfs-core#fix/resolve-ipns-test",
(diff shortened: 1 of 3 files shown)


obo20:
@alanshaw Is there anything keeping this from going in? We can fork on our end if needed, but if we could stick to using the official API that would be great.

alanshaw:
My apologies for not responding. The code change LGTM, it just needs a test in https://github.com/ipfs/interface-js-ipfs-core/blob/master/src/object/stat.js and then we can get it merged. Any chance you can send a quick PR?

obo20:
@alanshaw How do you want that process to work considering they're separate repositories?

alanshaw:
This repo depends on `interface-ipfs-core`. [This file](https://github.com/ipfs/js-ipfs-http-client/blob/master/test/interface.spec.js) pulls them in and runs them.

To write a test and run locally:

1. `git clone git@github.com:ipfs/js-ipfs-http-client.git`
1. `git clone git@github.com:ipfs/interface-js-ipfs-core.git`
1. `cd interface-js-ipfs-core && npm i && npm link`
1. Write your test [here](https://github.com/ipfs/interface-js-ipfs-core/blob/master/src/object/stat.js)
1. `cd ../js-ipfs-http-client && npm i && npm link interface-ipfs-core`
1. Run just the `object.stat` tests: `npm test -- --grep object.stat`

Next:

1. Send a PR to `ipfs/interface-js-ipfs-core` (please also add to the documentation!)
1. I'll review and do the same steps as above
1. When merged I'll release a new version of `interface-ipfs-core`
1. This PR needs to be updated to depend on the new version

HTH, shout if you have any issues.

obo20:
@alanshaw see: https://github.com/ipfs/interface-js-ipfs-core/pull/506

alanshaw:
Thanks @obo20 😘

PR #1058, opened 2019-07-25
docs: how to add an interface core test

alanshaw:


obo20:
Thanks for adding this in. It should help with a lot of confusion when contributing.

PR #1078, opened 2019-08-21
feat: support adding files via async iterator

achingbrain:
Adds a method called `ipfs._addAsyncIterator`, prefixed with an underscore to mark it as sort of unofficial (though it's what `ipfs.add` will become once the async iterator migration is complete) and we don't want to implement symmetrical `ipfs.catAsyncIterator` etc methods, or at least not yet.

Also adds support for sending the `pin` and (non-standard, JS-specific) `preload` arguments via HTTP.

### DIFF
--- a/src/files-regular/add-async-iterator.js
+++ b/src/files-regular/add-async-iterator.js
@@ -1,0 +1,23 @@
+'use strict'
+
+const SendFilesStream = require('../utils/send-files-stream')
+const FileResultStreamConverter = require('../utils/file-result-stream-converter')
+
+module.exports = (send) => {
+  return async function * (source, options) {
+    options = options || {}
+    options.converter = FileResultStreamConverter
+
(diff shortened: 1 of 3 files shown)


alanshaw:
This is super cool, note that https://github.com/ipfs/js-ipfs-http-client/pull/1063 is in the pipeline :)

src/files-regular/add-async-iterator.js
```diff
@@ -0,0 +1,23 @@
...
+
+module.exports = (send) => {
+  return async function * (source, options) {
+    options = options || {}
+    options.converter = FileResultStreamConverter
+
+    const stream = SendFilesStream(send, 'add')(options)
+
+    for await (const entry of source) {
+      stream.write(entry)
```
alanshaw: Do you need to do some conversion of `entry.content`?

PR #1101, opened 2019-09-10
refactor: use fetch for ipfs.cat

alanshaw:


src/lib/error-handler.js
```diff
@@ -2,13 +2,25 @@
...
+  if (response.ok) {
+    // FIXME: remove when fixed https://github.com/sindresorhus/ky-universal/issues/8
+    //
+    // ky clones the response for each handler. In Node.js the response body is
+    // piped to 2 PassThroughs, one becomes the real body and the other is used
+    // in the clone.
+    //
+    // If the body in the clone is not consumed or destroyed the highwater mark
+    // will be reached (for large payloads) and stop the real body from flowing.
+    if (isNode) response.body.destroy()
```
alanshaw: This...sucks...

achingbrain: `ky` has some, uh, interesting caveats.

achingbrain: See also: https://github.com/ipfs/js-ipfs/pull/2304/files#r322675702

src/cat.js
```diff
@@ -0,0 +1,34 @@
...
+      searchParams.set('arg', new CID(path).toString())
+    }
+
+    if (options.offset) searchParams.set('offset', options.offset)
+    if (options.length) searchParams.set('length', options.length)
+
+    const res = await ky.get('cat', {
+      timeout: options.timeout,
+      signal: options.signal,
+      headers: options.headers,
```
hugomrdias: docs only has offset and length for options https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/FILES.md#cat.

src/cat.js
```diff
@@ -0,0 +1,34 @@
...
+      searchParams
+    })
+
+    for await (const chunk of toIterable(res.body)) {
+      yield Buffer.from(chunk)
+    }
+  })()
```
hugomrdias: why is this a IIEF ?

src/cat.js
```diff
@@ -0,0 +1,34 @@
...
+    if (options.length) searchParams.set('length', options.length)
+
+    const res = await ky.get('cat', {
+      timeout: options.timeout,
+      signal: options.signal,
+      headers: options.headers,
+      searchParams
+    })
+
+    for await (const chunk of toIterable(res.body)) {
```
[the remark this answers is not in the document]
alanshaw: Because of catReadableStream and catPullStream - only cat concats the chunks

src/cat.js
alanshaw: These are not options for the core interface but they will be available to all methods in the HTTP client. They need to be documented when they are supported globally. I’ll create an issue to track this and ensure it gets done.

src/cat.js
alanshaw: Its a generator function that needs to be called to obtain the generator. I can split this out into a definition and a return of the called function if you think it’ll be easier to understand.

PR #1105, opened 2019-09-13
fix: fix electron renderer tests and a couple more bugs

hugomrdias:
- temp fix for electron renderer until ky-universal is fixed
- Blob can accept Buffer directly
  - Buffer.buffer triggered a super weird bug in electron renderer
- enable a bunch of tests in the browser
- xvfb-run makes electron go crazy
- detect-node needs to be removed everywhere and ipfs-utils/src/env.js should be used
- If you mix promises and callbacks inside, call the callback with setImmediate

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,9 +1,13 @@
 'use strict'
 
+const promisify = require('promisify-es6')
 const createServer = require('ipfsd-ctl').createServer
-
+const EchoServer = require('interface-ipfs-core/src/utils/echo-http-server')
 const server = createServer()
+const echoServer = EchoServer.createServer()
 
+const echoServeStart = promisify(echoServer.start)
+const echoServeStop = promisify(echoServer.stop)
 module.exports = {
   bundlesize: { maxSize: '245kB' },
   webpack: {
@@ -22,9 +26,24 @@ module.exports = {
     singleRun: true
   },
   hooks: {
+    node: {
+      pre: () => echoServeStart(),
+      post: () => echoServeStop()
+    },
     browser: {
-      pre: () => server.start(),
-      post: () => server.stop()
+      pre: () => {
+        return Promise.all([
+          server.start(),
+          echoServeStart()
+
+        ])
+      },
+      post: () => {
+        return Promise.all([
+          server.stop(),
+          echoServeStop()
+        ])
+      }
     }
   }
 }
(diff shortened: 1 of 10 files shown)


src/add/form-data.js
```diff
@@ -40,3 +41,9 @@ exports.toFormData = async input => {
 
   return formData
 }
+
+// TODO remove this when upstream fix for ky-universal is merged
```
achingbrain: Could you link to the pending PR please?

test/interface.spec.js
```diff
@@ -102,40 +102,30 @@ describe('interface-ipfs-core tests', () => {
...
+      // // .addPullStream
+      // isNode ? null : {
+      //   name: 'should add pull stream of valid files and dirs',
+      //   reason: 'FIXME https://github.com/ipfs/js-ipfs-http-client/issues/339'
+      // },
+      // // .addReadableStream
+      // isNode ? null : {
+      //   name: 'should add readable stream of valid files and dirs',
+      //   reason: 'FIXME https://github.com/ipfs/js-ipfs-http-client/issues/339'
+      // },
```
achingbrain: Please delete commented code.

test/interface.spec.js
```diff
@@ -148,27 +138,27 @@ describe('interface-ipfs-core tests', () => {
...
+      // // .lsPullStream
+      // isNode ? null : {
+      //   name: 'should pull stream ls with a base58 encoded CID',
+      //   reason: 'FIXME https://github.com/ipfs/js-ipfs-http-client/issues/339'
+      // },
+      // // .lsReadableStream
+      // isNode ? null : {
+      //   name: 'should readable stream ls with a base58 encoded CID',
+      //   reason: 'FIXME https://github.com/ipfs/js-ipfs-http-client/issues/339'
+      // }
```
achingbrain: Please delete commented code.

.aegir.js
```diff
@@ -1,9 +1,13 @@
 'use strict'
 
+const promisify = require('promisify-es6')
 const createServer = require('ipfsd-ctl').createServer
-
+const EchoServer = require('interface-ipfs-core/src/utils/echo-http-server')
 const server = createServer()
+const echoServer = EchoServer.createServer()
 
+const echoServeStart = promisify(echoServer.start)
```
achingbrain: Missing the r in server: `echoServerStart`

test/interface.spec.js
hugomrdias: it was on purpose just want to make sure these pass in the CI, they were skipped for a reason. But they seem to pass now. 
ill remove the comments after CI green.

PR #1119, opened 2019-10-03
feat: add methods for listing config profiles

achingbrain:
New method:

```javascript
Promise<Array<{ name, description }>> ipfs.config.profiles.list()
```

BREAKING CHANGE:

```javascript
Promise<{oldCfg, newCfg}> ipfs.config.profile(name, opts)

// is now
Promise<{old, new}> ipfs.config.profiles.apply(name, opts)
```

Possibly contentious:

Adds `callbackify` as a dependency, see https://github.com/ipfs/js-ipfs/issues/2506 for discussion.

Depends on:

- [x] https://github.com/ipfs/interface-js-ipfs-core/pull/536

### DIFF
--- a/src/config/profiles/list.js
+++ b/src/config/profiles/list.js
@@ -1,21 +1,17 @@
 'use strict'
 
-const configure = require('../../lib/configure')
 const callbackify = require('callbackify')
+const configure = require('../../lib/configure')
 const toCamel = require('../../lib/object-to-camel')
 
 module.exports = configure(({ ky }) => {
   return callbackify.variadic(async (options) => {
     options = options || {}
 
-    const searchParams = new URLSearchParams(options.searchParams)
-    searchParams.set('stream-channels', true)
-
     const res = await ky.get('config/profile/list', {
       timeout: options.timeout,
       signal: options.signal,
-      headers: options.headers,
-      searchParams
+      headers: options.headers
     })
 
     const parsed = await res.json()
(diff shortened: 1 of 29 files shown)


src/config/profiles/list.js
```diff
@@ -0,0 +1,26 @@
...
+    options = options || {}
+
+    const searchParams = new URLSearchParams(options.searchParams)
+    searchParams.set('stream-channels', true)
+
+    const res = await ky.get('config/profile/list', {
+      timeout: options.timeout,
+      signal: options.signal,
+      headers: options.headers,
+      searchParams
```
hugomrdias: just pass an object { 'stream-channels': true } ky will do the rest, no need to manually instantiate URLSearchParams

achingbrain: I think I just copy/pasted that from `ipfs.add`..

achingbrain: Hmm, you can only set strings or numbers as values if you use a plain object (e.g. no booleans), otherwise you get:

```
Error: The `searchParams` option must be either a string, `URLSearchParams` instance or an object with string and number values
```

achingbrain: https://github.com/sindresorhus/ky/issues/182

achingbrain: Looks like they're going to fix this in `ky@0.16.0`

PR #1127, opened 2019-10-11
fix: use new ipfsd-ctl setup

hugomrdias:
Adds support for the new ipfsd-ctl tests setup.

depends on:
* [x] https://github.com/ipfs/js-ipfsd-ctl/pull/395
* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/541


test/commands.spec.js
```diff
@@ -2,30 +2,18 @@
 'use strict'
 
 const { expect } = require('interface-ipfs-core/src/utils/mocha')
-const ipfsClient = require('../src')
 const f = require('./utils/factory')
```
alanshaw: I would prefer for this to have a better name, but can be done another time.

hugomrdias:
Ready to merge

PR #1130, opened 2019-10-18
fix: increase default timeout and respect value passed to `ky.extend`

achingbrain:
Some of our operations take a really long time, if we don't set a timeout for `ky` we get the default of 10 seconds. This PR sets the timeout to `false` if one is not explicitly passed which disables it.

Nb. I had to add the default to `false` to every invocation. Looking at the code it should be enough to do it in `src/lib/configure.js` but it doesn't seem to be.  Might be a bug in `ky`, I'll dig a little deeper.

### DIFF
--- a/src/lib/configure.js
+++ b/src/lib/configure.js
@@ -27,7 +27,7 @@ module.exports = create => config => {
     // https://github.com/sindresorhus/ky/pull/153
     ky: ky.extend({
       prefixUrl: config.apiAddr + config.apiPath,
-      timeout: config.timeout || 60 * 1000,
+      timeout: config.timeout || false,
       headers: config.headers,
       hooks: {
         afterResponse: [errorHandler]
--- a/src/lib/configure.js
+++ b/src/lib/configure.js
@@ -5,6 +5,7 @@ const ky = require('ky-universal').default
 const { isBrowser, isWebWorker } = require('ipfs-utils/src/env')
 const { toUri } = require('./multiaddr')
 const errorHandler = require('./error-handler')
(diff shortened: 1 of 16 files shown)


achingbrain:
Predictably it's because we pass `undefined` for `options.timeout` which `ky` then defaults to it's own default of `10000` and not what we pass into `ky.extend` in `src/lib/configure.js`.

achingbrain:
I've changed the timeout to wait for something stupid but not to wait forever, and also added a `ignoreUndefined` option to `merge-options` so we can filter the options we pass to `ky` but also not pass `undefined` for values.

Depends on https://github.com/schnittstabil/merge-options/pull/14

src/lib/configure.js
```diff
@@ -57,3 +67,11 @@ function getDefaultApiAddr ({ protocol, host, port }) {
...
   return `${protocol || 'http'}://${host || 'localhost'}:${port || 5001}`
 }
+
+// returns the passed function wrapped in a function that ignores
+// undefined values in the passed `options` object
+function wrap (fn, defaults) {
+  return (input, options) => {
+    return fn(input, mergeOptions(defaults, options))
+  }
+}
```
schnittstabil: Suggestion:
```js
const wrap = (fn, defaults) => (input, options) => fn(input, mergeOptions(defaults, options));
```

achingbrain: I find this sort of deeply chained single-line-arrow-function style optimises readability for computers and not humans.  It'll get minified to something similar to this anyway so there's little to gain here.

PR #1145, opened 2019-11-05
feat: support name.resolve of peerid as cid

lidel:
This PR updates to `interface-ipfs-core": "^0.118.0` which includes test for resolving paths with libp2p-key represented as CIDv1 Base32.

### Open Problem

The problem is that go-ipfs 0.4.22 does not support PeerID as CID, so test added in https://github.com/ipfs/interface-js-ipfs-core/pull/553 fails:

```
  1) interface-ipfs-core tests
       .name.resolve offline
         should resolve a record from peerid as cidv1 in base32:
     Error: not a valid proquint string
```

<del>@alanshaw  not sure how to handle this. Wait for go-ipfs v0.5.0? Fixup on the fly in the meantime?</del>

<del>A hacky way to support CIDs when API is go-ipfs 0.4.22 or older is to convert toBase58btc before sending. See below.</del>

For now we skip cidv1b32 test, added todo to unskip when go-ipfs ships with native support.

closes https://github.com/ipfs/js-ipfs-http-client/pull/1147

### DIFF
--- a/src/name/resolve.js
+++ b/src/name/resolve.js
@@ -1,6 +1,8 @@
 'use strict'
 
 const promisify = require('promisify-es6')
+const CID = require('cids')
+const multihash = require('multihashes')
 
 const transform = function (res, callback) {
   callback(null, res.Path)
@@ -11,6 +13,19 @@ module.exports = (send) => {
     if (typeof (opts) === 'function') {
       callback = opts
       opts = {}
+    }
+
+    // normalize PeerIDs to Base58btc, so it works with go-ipfs <=0.4.22
+    if (typeof (args) === 'string') {
+      try {
+        const path = args.split('/')
+        if (path.length > 2) {
+          path[2] = multihash.toB58String(new CID(path[2]).multihash)
+          args = path.join('/')
+        }
+      } catch (err) {
+        // noop
+      }
     }
 
     send.andTransform({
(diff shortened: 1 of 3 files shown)


src/name/resolve.js
```diff
@@ -13,6 +15,19 @@ module.exports = (send) => {
...
+      try {
+        const path = args.split('/')
+        if (path.length > 2) {
+          path[2] = multihash.toB58String(new CID(path[2]).multihash)
+          args = path.join('/')
+        }
+      } catch (err) {
+        // noop
+      }
+    }
```
lidel: is there a better way? :/

alanshaw: All we need to do is stringify `args` if it is a CID. We will skip the `interface-ipfs-core` test here until go-ipfs supports it.

alanshaw:
Please would you add the updated `interface-ipfs-core` dep here and skip the relevant test?

src/name/resolve.js
lidel: We don't stringify CIDs anywhere else, removed this hack and added TODO to unskip mentioned test when go-ipfs implements this.

lidel:
@alanshaw done

src/name/resolve.js
alanshaw: Ah interesting, I guess the querystring module calls `toString` on the CID then...

PR #1148, opened 2019-11-07
fix: repo stat for non-numeric string values

PedroMiguelSS:
`new Big()` cannot be applied to non-numeric `string` variables, which can be the case when `human` option is `true`.
E.g.: `10GB`

This PR fixes that problem.

⚠️ It needs the following PR merged and released before this one can be:
- [ ] https://github.com/ipfs/interface-js-ipfs-core/pull/554

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "browser-process-platform": "~0.1.1",
     "cross-env": "^6.0.0",
     "go-ipfs-dep": "^0.4.22",
-    "interface-ipfs-core": "^0.119.0",
+    "interface-ipfs-core": "github:ipfs/interface-js-ipfs-core#test/add-human-option-test-repo-stat",
     "ipfsd-ctl": "^0.47.1",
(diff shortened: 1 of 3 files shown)


alanshaw:
@PedroMiguelSS please add a test to https://github.com/ipfs/interface-js-ipfs-core/blob/master/src/repo/stat.js so we can verify this works. Shout if you need a hand with anything.

PedroMiguelSS:
Done https://github.com/ipfs/interface-js-ipfs-core/pull/554 @alanshaw.

PedroMiguelSS:
Can I get a second review here, please?

Please note that this PR depends on https://github.com/ipfs/interface-js-ipfs-core/pull/554.

achingbrain:
LGTM, just needs the git url removing from the `package.json`

PedroMiguelSS:
Closed in favor of https://github.com/ipfs/js-ipfs/pull/2630.

PR #1150, opened 2019-11-08
refactor: move files to root level

alanshaw:
Moves remaining files that contain root level commands to the root of the src directory. This makes it easier for new contributors to find things.

It also renames the `files-mfs` dir to `files` to reflect the fact that it lives in the "files" namespace currently.

BREAKING CHANGE: files in `src/files-regular` have moved to `src`. The `src/files-mfs` directory has been renamed to `src/files`. If you were previously requiring files from these directories e.g. `require('ipfs-http-client/src/files-regular/add')` then please be aware that they have moved.


alanshaw:
@hugomrdias @achingbrain any chance you can review these changes?

PR #1153, opened 2019-11-12
refactor: convert block API to async await

alanshaw:
depends on:

* [x] #1150

### DIFF
--- a/src/block/put.js
+++ b/src/block/put.js
@@ -1,74 +1,58 @@
 'use strict'
 
-const promisify = require('promisify-es6')
+const FormData = require('form-data')
 const Block = require('ipfs-block')
 const CID = require('cids')
 const multihash = require('multihashes')
-const SendOneFile = require('../utils/send-one-file')
+const configure = require('../lib/configure')
 
-module.exports = (send) => {
-  const sendOneFile = SendOneFile(send, 'block/put')
-
-  return promisify((block, options, callback) => {
-    if (typeof options === 'function') {
-      callback = options
-      options = {}
-    }
-
+module.exports = configure(({ ky }) => {
+  async function put (data, options) {
     options = options || {}
 
-    if (Array.isArray(block)) {
-      return callback(new Error('block.put accepts only one block'))
+    if (Block.isBlock(data)) {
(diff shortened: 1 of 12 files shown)


src/block/put.js
```diff
@@ -1,74 +1,60 @@
...
-    if (block.cid || options.cid) {
-      let cid
-
-      try {
-        cid = new CID(block.cid || options.cid)
-      } catch (err) {
-        return callback(err)
+    const searchParams = new URLSearchParams(options.searchParams)
+    if (options.format) searchParams.set('format', options.format)
+    if (options.format) searchParams.set('mhtype', options.mhtype)
```
hugomrdias:
```suggestion
    if (options.mhtype) searchParams.set('mhtype', options.mhtype)
```

alanshaw: Good catch, thanks!

alanshaw:
```js
const searchParams = new URLSearchParams()
```

Agree, that shouldn't happen, we should allow them to be overridden with options on every API call. Along with `timeout`, `signal` etc.

```js
```

`URLSearchParams` allows you to pass a `URLSearchParams` instance OR an object as the parameter. This allows _users_ to pass their own query string params as an object or an instance of `URLSearchParams` AND also allows _us_ to add mandatory/overridden parameters from args or other options.

```js
if (options.format !== undefined) {
  searchParams.set('format', options.format)
}

if(options.format != null)
```

Sometimes we check for `null`/`undefined` because a falsy check on booleans or numbers that might be 0 is not good enough. It's a matter of preference if we only consider `undefined` as a "value" that doesn't get added to the query string. I've been using the `!= null` because it includes `null` as well as `undefined` and 99.9% of the time you also would not want `?arg=null` which is what happens when you pass `null` to `URLSearchParams` and stringify it.

alanshaw:
@hugomrdias apologies I'm going to expand this PR to include the DAG API since it uses the block API.

achingbrain:
I think we should apply the robustness principal - that is, all the `if` statements around `URLSearchParams` are verbose but it's good to only send what we know the server will support.

PR #1154, opened 2019-11-12
refactor: convert bootstrap API to async/await

alanshaw:


alanshaw:
ping @hugomrdias @achingbrain 🙏

src/bootstrap/add.js
```diff
@@ -1,32 +1,28 @@
...
 
-module.exports = (send) => {
-  return promisify((args, opts, callback) => {
-    if (typeof opts === 'function' &&
-      !callback) {
-      callback = opts
-      opts = {}
+module.exports = configure(({ ky }) => {
+  return async (addr, options) => {
+    if (addr && typeof addr === 'object' && !Multiaddr.isMultiaddr(addr)) {
```
achingbrain: Are we trying to support an `options` object as the first argument?  If so, is this necessary?  The spec says [`addr` is not optional](https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/BOOTSTRAP.md#bootstrapadd).

src/bootstrap/rm.js
```diff
@@ -1,32 +1,28 @@
...
 
-module.exports = (send) => {
-  return promisify((args, opts, callback) => {
-    if (typeof opts === 'function' &&
-      !callback) {
-      callback = opts
-      opts = {}
+module.exports = configure(({ ky }) => {
+  return async (addr, options) => {
+    if (addr && typeof addr === 'object' && !Multiaddr.isMultiaddr(addr)) {
```
achingbrain: Same as before, `addr` is not optional.

src/bootstrap/rm.js
```diff
@@ -1,32 +1,28 @@
...
 
-    if (args && typeof args === 'object') {
-      opts = args
-      args = undefined
-    }
+    const searchParams = new URLSearchParams(options.searchParams)
+    if (addr) searchParams.set('arg', `${addr}`)
+    if (options.all != null) searchParams.set('all', options.all)
+
+    const res = await ky.post('bootstrap/rm', {
```
achingbrain: Looking at the commit history, what happened with `DELETE` here?  Our HTTP API is all `POST`, all the time so this is fine like this, but that aside I thought `DELETE` was [ok to use in browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE#Browser_compatibility)?

src/bootstrap/add.js
alanshaw: It's what the current code does. So you can pass `add(null, { default: true })` or `add({ default: true })` to add the default bootstrappers. Same sort of situation for `rm` (with `all` option).

So `addr` is required _unless_ you're passing the `default` option.

src/bootstrap/rm.js
alanshaw: I couldn't be bothered to dig in at the time but thinking about it now it's because browsers are respecting CORS and our default `ipfsd-ctl` config does not include DELETE https://github.com/ipfs/js-ipfsd-ctl/blob/7928c1eee092a9b96a59328294c84c9f065217a3/src/defaults/config.json#L5-L9

src/bootstrap/add.js
alanshaw: https://github.com/ipfs/interface-js-ipfs-core/pull/557

src/bootstrap/rm.js
hugomrdias: Do you want me to add DELETE ?
In this case i would use POST cause its more RPC than REST
i would only use DELETE if the url was https://host/bootstrap/{addr}

src/bootstrap/add.js
hugomrdias: IMO we should leave it like this in this PR.

But i agree with alex and we should make an issue to improve this in the future cause from https://github.com/ipfs/interface-js-ipfs-core/blob/master/SPEC/BOOTSTRAP.md#ipfsbootstrapaddaddr-options i would understand i could do `add({ default: true })` maybe we can simplify to: 
if first param is undefined use defaults.

src/bootstrap/rm.js
achingbrain: I'm not suggesting changing this to `DELETE`, I just wanted to know what happened 😉

src/bootstrap/add.js
achingbrain: Ah, yes, quite right.  Sigh.

alanshaw: > But i agree with alex and we should make an issue to improve this in the future

❤️ me too. I'm just trying to get everything switched to async/await first and then take a second pass at API changes.

PR #1155, opened 2019-11-12
refactor: convert config API to async await

alanshaw:
BREAKING CHANGE: Errors returned from request failures are now all [`HTTPError`](https://github.com/sindresorhus/ky/blob/c0d9d2bb07e4c122a08f019b39e9c55a4c9324f3/index.js#L117-L123)s which carry a `response` property. This is a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) that can be used to inspect _all_ information relating to the HTTP response. This means that the `err.status` or `err.statusCode` property should now be accessed via `err.response.status`.

### DIFF
--- a/src/config/set.js
+++ b/src/config/set.js
@@ -8,7 +8,7 @@ module.exports = configure(({ ky }) => {
     options = options || {}
 
     if (typeof key !== 'string') {
-      throw new Error('invalid key')
+      throw new Error('Invalid key type')
     }
 
     const searchParams = new URLSearchParams(options.searchParams)
--- a/src/config/set.js
+++ b/src/config/set.js
@@ -1,35 +1,36 @@
 'use strict'
 
-const promisify = require('promisify-es6')
+const configure = require('../lib/configure')
+const toCamel = require('../lib/object-to-camel')
 
-module.exports = (send) => {
-  return promisify((key, value, opts, callback) => {
-    if (typeof opts === 'function') {
-      callback = opts
-      opts = {}
-    }
+module.exports = configure(({ ky }) => {
+  return async (key, value, options) => {
+    options = options || {}
+
     if (typeof key !== 'string') {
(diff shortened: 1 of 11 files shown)


src/config/set.js
```diff
@@ -1,35 +1,36 @@
...
 
-module.exports = (send) => {
-  return promisify((key, value, opts, callback) => {
-    if (typeof opts === 'function') {
-      callback = opts
-      opts = {}
-    }
     if (typeof key !== 'string') {
-      return callback(new Error('Invalid key type'))
+      throw new Error('invalid key')
```
achingbrain: Why the change in message?  Is this for go compatibility or similar?

test/lib.error-handler.spec.js
```diff
@@ -21,8 +21,9 @@ describe('lib/error-handler', () => {
 
     const err = await throwsAsync(errorHandler(null, null, res))
 
+    expect(err instanceof HTTPError).to.be.true()
     expect(err.message).to.eql('boom')
-    expect(err.status).to.eql(500)
+    expect(err.response.status).to.eql(500)
```
achingbrain: Breaking change?  I guess errors aren't really part of the spec other than 'you will get an error'.

alanshaw: Yes, it is breaking. I've updated the PR desctiption with the info and will include it when merging.

src/config/set.js
alanshaw: No, I will revert.

test/lib.error-handler.spec.js
achingbrain: Great.  One small thing - the description says `err.status` has changed but it's `err.statusCode`

alanshaw: Well this test was looking for `.status` but others are looking for `.statusCode`...I've added both to the breaking change message.

PR #1156, opened 2019-11-13
refactor: convert dht API to async/await

alanshaw:


src/dht/find-provs.js
```diff
@@ -0,0 +1,40 @@
...
+
+    const res = await ky.get('dht/findprovs', {
+      timeout: options.timeout,
+      signal: options.signal,
+      headers: options.headers,
+      searchParams
+    })
+
+    for await (const message of ndjson(toIterable(res.body))) {
+      log(message)
```
achingbrain: Do we need to log the message?

Here and in all the other DHT related calls?

PR #1160, opened 2019-11-15
refactor: convert object.* and files.* API methods to async/await

alanshaw:
This also allows the `timeout` option to be passed as a human readable string, which is compatible with the `go-ipfs` global `timeout` option.

Depends on:

* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/558


src/files/flush.js
```diff
@@ -1,17 +1,27 @@
...
-      args: args
-    }, (error) => callback(error))
-  })
-}
+    options = options || {}
+
+    const searchParams = new URLSearchParams(options.searchParams)
+    searchParams.set('arg', path)
+
+    const { Cid } = await ky.post('files/flush', {
```
alanshaw: Note: needs https://github.com/ipfs/js-ipfs-mfs/issues/50 for this to work with js-ipfs also

src/lib/configure.js
```diff
@@ -27,7 +28,7 @@ module.exports = create => config => {
   // https://github.com/sindresorhus/ky/pull/153
```
hugomrdias: do we still need this ? its been merged

alanshaw: This is not the PR I thought it was, it still gets `fetch` off the global object so we can't pass a custom fetch implementation. This comment should just be removed.

PR #1173, opened 2019-11-21
refactor: async await roundup

### LINKED ISSUE #1103: Document global options timeout, signal, headers, searchParams
As per comment here https://github.com/ipfs/js-ipfs-http-client/pull/1101#discussion_r323286361

Needs to be done after all methods converted to use fetch.

alanshaw:
This is a round up of the remaining async/await PRs along with some cleanup and docs fixes.

resolves #1103
resolves https://github.com/ipfs/js-ipfs-http-client/issues/1122
resolves https://github.com/ipfs/js-ipfs-http-client/issues/1158 (hopefully!)
closes https://github.com/ipfs/js-ipfs-http-client/pull/1164
closes https://github.com/ipfs/js-ipfs-http-client/pull/1165
closes https://github.com/ipfs/js-ipfs-http-client/pull/1166
closes https://github.com/ipfs/js-ipfs-http-client/pull/1169
closes https://github.com/ipfs/js-ipfs-http-client/pull/1170
closes https://github.com/ipfs/js-ipfs-http-client/pull/1172

BREAKING CHANGE: The `log.tail` method now returns an async iterator that yields log messages. Use it like:

```js
for await (const message of ipfs.log.tail()) {
  console.log(message)
}
```

BREAKING CHANGE: The response to a call to `log.level` now returns an object that has camel cased keys. i.e. `Message` and `Error` properties have changed to `message` and `error`.

BREAKING CHANGE: Dropped support for go-ipfs <= 0.4.4 in `swarm.peers` response.

BREAKING CHANGE: The signature for `ipfs.mount` has changed from `ipfs.mount([ipfsPath], [ipnsPath])` to `ipfs.mount([options])`. Where `options` is an optional object that may contain two boolean properties `ipfsPath` and `ipnsPath`. The response object has also changed to be camel case. See https://docs.ipfs.io/reference/api/http/#api-v0-mount.

BREAKING CHANGE: Default ping `count` of 1 in client has been removed. The default ping count is now whatever the IPFS node defaults it to (currently 10). If you specifically need 1 ping message then please pass `count: 1` in options for `ipfs.ping()`.

BREAKING CHANGE: Multi parameter constructor options are no longer supported. To create a new IPFS HTTP client, pass a single parameter to the constructor. The parameter can be one of:

* String, formatted as one of:
    * Multiaddr e.g. /ip4/127.0.0.1/tcp/5001
    * URL e.g. http://127.0.0.1:5001
* [Multiaddr](https://www.npmjs.com/package/multiaddr) instance
* Object, in format of either:
    * Address and path e.g. `{ apiAddr: '/ip4/127.0.0.1/tcp/5001': apiPath: '/api/v0' }` (Note: `apiAddr` can also be a string in URL form or a Multiaddr instance)
    * Node.js style address e.g. `{ host: '127.0.0.1', port: 5001, protocol: 'http' }`


test/log.spec.js
```diff
@@ -65,7 +56,7 @@ describe('.log', function () {
 
     expect(res).to.exist()
     expect(res).to.be.an('object')
-    expect(res).to.not.have.property('Error')
-    expect(res).to.have.property('Message')
+    expect(res).to.not.have.property('error')
+    expect(res).to.have.property('message')
```
achingbrain: Did the property names change?

alanshaw: Yep! ...and documented in the breaking changes above.

PR #1174, opened 2019-11-21
refactor: async iterables

alanshaw:
TLDR;

* Remove Node.js streams and pull streams
* Remove callbacks
* Remove `peer-info` and `peer-id`

---

Now that internals are all async/await/iterables and `fetch` the next step is to bubble up that goodness to the core API, removing the multiple stream APIs and removing callback support.

I'm also proposing removing `peer-info` and `peer-id`, since these drastically increase the bundle size by pulling in `libp2p-crypto`, for which 99% of it's capability is unused. In place of `peer-id` we return a `CID`, which can easily be converted to a `PeerId` instance via:

```js
const peerId = PeerId.createFromCID(peerCid)
```

In place of `peer-info` we return an object `{ id: CID, addrs: Multiaddr[] }`, which can easily be converted to a `PeerInfo` like:

```js
const peerInfo = new PeerInfo(PeerId.createFromCID(info.id))
info.addrs.forEach(addr => peerInfo.multiaddrs.add(addr))
```

refs https://github.com/ipfs/js-ipfs/issues/1670

TODO:

* [ ] Refactor local tests
* [ ] Refactor `interface-ipfs-core` tests
* [ ] Document new APIs in `interface-ipfs-core`
* [ ] Update README with API changes
* [ ] Document upgrade path from Node.js/pull streams to async iterables


alanshaw:
History cleaned up in https://github.com/ipfs/js-ipfs-http-client/pull/1183

PR #1183, opened 2019-11-27
refactor: async iterables

alanshaw:
TLDR;

* Remove Node.js streams and pull streams
* Remove callbacks
* Remove `peer-info` and `peer-id`

---

Now that internals are all async/await/iterables and `fetch` the next step is to bubble up that goodness to the core API, removing the multiple stream APIs and removing callback support.

I'm also proposing removing `peer-info` and `peer-id`, since these drastically increase the bundle size by pulling in `libp2p-crypto`, for which 99% of it's capability is unused. In place of `peer-id` we return a `CID`, which can easily be converted to a `PeerId` instance via:

```js
const peerId = PeerId.createFromCID(peerCid)
```

In place of `peer-info` we return an object `{ id: CID, addrs: Multiaddr[] }`, which can easily be converted to a `PeerInfo` like:

```js
const peerInfo = new PeerInfo(PeerId.createFromCID(info.id))
info.addrs.forEach(addr => peerInfo.multiaddrs.add(addr))
```

refs https://github.com/ipfs/js-ipfs/issues/1670
refs https://github.com/ipfs/js-ipfs/issues/2611
refs https://github.com/ipfs/interface-js-ipfs-core/issues/394

TODO:

* [x] Refactor local tests
* [x] Remove `addFromFs` and `addFromUrl` and export `globSource` and `urlSource` instead
* [x] Refactor `interface-ipfs-core` tests
* [x] Document new APIs in `interface-ipfs-core`
* [x] Update README with API changes
* [x] [Document upgrade path from Node.js/pull streams to async iterables](https://gist.github.com/alanshaw/04b2ddc35a6fff25c040c011ac6acf26)

Depends on:

* [x] https://github.com/ipfs/js-ipfs-utils/pull/15
* [x] https://github.com/ipfs/interface-js-ipfs-core/pull/567

BREAKING CHANGE: Callbacks are no longer supported on any API methods. Please use a utility such as [`callbackify`](https://www.npmjs.com/package/callbackify) on API methods that return Promises to emulate previous behaviour.

BREAKING CHANGE: `PeerId` and `PeerInfo` classes are no longer statically exported from `ipfs-http-client` since they are no longer used internally.

BREAKING CHANGE: `pin.add` results now contain a `cid` property (a [CID instance](https://github.com/multiformats/js-cid)) instead of a string `hash` property.

BREAKING CHANGE: `pin.ls` now returns an async iterable.

BREAKING CHANGE: `pin.ls` results now contain a `cid` property (a [CID instance](https://github.com/multiformats/js-cid)) instead of a string `hash` property.

BREAKING CHANGE: `pin.rm` results now contain a `cid` property (a [CID instance](https://github.com/multiformats/js-cid)) instead of a string `hash` property.

BREAKING CHANGE: `add` now returns an async iterable.

BREAKING CHANGE: `add` results now contain a `cid` property (a [CID instance](https://github.com/multiformats/js-cid)) instead of a string `hash` property.

BREAKING CHANGE: `addReadableStream`, `addPullStream` have been removed.

BREAKING CHANGE: `ls` now returns an async iterable.

BREAKING CHANGE: `ls` results now contain a `cid` property (whose value is a [CID instance](https://github.com/multiformats/js-cid)) instead of a `hash` property.

BREAKING CHANGE: `files.ls` now returns an async iterable.

BREAKING CHANGE: `files.readPullStream` and `files.readReadableStream` have been removed.

BREAKING CHANGE: `files.read` now returns an async iterable.

BREAKING CHANGE: `files.lsPullStream` and `files.lsReadableStream` have been removed.

BREAKING CHANGE: `files.ls` now returns an async iterable.

BREAKING CHANGE: `files.ls` results now contain a `cid` property (whose value is a [CID instance](https://github.com/multiformats/js-cid)) instead of a `hash` property.

BREAKING CHANGE: `files.ls` no longer takes a `long` option (in core) - you will receive all data by default.

BREAKING CHANGE: `files.stat` result now contains a `cid` property (whose value is a [CID instance](https://github.com/multiformats/js-cid)) instead of a `hash` property.

BREAKING CHANGE: `get` now returns an async iterable. The `content` property value for objects yielded from the iterator is now an async iterable that yields [`BufferList`](https://github.com/rvagg/bl) objects.

BREAKING CHANGE: `stats.bw` now returns an async iterable.

BREAKING CHANGE: `addFromStream` has been removed. Use `add` instead.

BREAKING CHANGE: `isIPFS` is no longer exported from the client, please `npm i is-ipfs` or include the CDN script tag `<script src="https://unpkg.com/is-ipfs/dist/index.min.js"></script>` to use this utility in your applications.

BREAKING CHANGE: `addFromFs` has been removed. Please use the exported `globSource` utility and pass the result to `add`. See the [glob source documentation](https://github.com/ipfs/js-ipfs-http-client#glob-source) for more details and an example.

BREAKING CHANGE: `addFromURL` has been removed. Please use the exported `urlSource` utility and pass the result to `add`. See the [URL source documentation](https://github.com/ipfs/js-ipfs-http-client#url-source) for more details and an example.

BREAKING CHANGE: `name.resolve` now returns an async iterable. It yields increasingly more accurate resolved values as they are discovered until the best value is selected from the quorum of 16. The "best" resolved value is the last item yielded from the iterator. If you are interested only in this best value you could use `it-last` to extract it like so:

```js
const last = require('it-last')
await last(ipfs.name.resolve('/ipns/QmHash'))
```

BREAKING CHANGE: `block.rm` now returns an async iterable.

BREAKING CHANGE: `block.rm` now yields objects of `{ cid: CID, error: Error }`.

BREAKING CHANGE: `dht.findProvs`, `dht.provide`, `dht.put` and `dht.query` now all return an async iterable.

BREAKING CHANGE: `dht.findPeer`, `dht.findProvs`, `dht.provide`, `dht.put` and `dht.query` now yield/return an object `{ id: CID, addrs: Multiaddr[] }` instead of a `PeerInfo` instance(s).

BREAKING CHANGE: `refs` and `refs.local` now return an async iterable.

BREAKING CHANGE: `object.data` now returns an async iterable that yields `Buffer` objects.

BREAKING CHANGE: `ping` now returns an async iterable.

BREAKING CHANGE: `repo.gc` now returns an async iterable.

BREAKING CHANGE: `swarm.peers` now returns an array of objects with a `peer` property that is a `CID`, instead of a `PeerId` instance.

BREAKING CHANGE: `swarm.addrs` now returns an array of objects `{ id: CID, addrs: Multiaddr[] }` instead of `PeerInfo` instances.

BREAKING CHANGE: `block.stat` result now contains a `cid` property (whose value is a [CID instance](https://github.com/multiformats/js-cid)) instead of a `key` property.

BREAKING CHANGE: `bitswap.wantlist` now returns an array of [CID](https://github.com/multiformats/js-cid) instances.

BREAKING CHANGE: `bitswap.stat` result has changed - `wantlist` and `peers` values are now an array of [CID](https://github.com/multiformats/js-cid) instances.


src/name/resolve.js
```diff
@@ -1,13 +1,16 @@
...
+const toIterable = require('stream-to-it/source')
 
 module.exports = configure(({ ky }) => {
-  return async (path, options) => {
+  return async function * (path, options) {
     options = options || {}
 
     const searchParams = new URLSearchParams(options.searchParams)
     searchParams.set('arg', path)
+    searchParams.set('stream', true)
```
hugomrdias: the buffered version isn't an option anymore ?

alanshaw: This yields increasingly more accurate values as they are discovered until the best value is selected from the quorum of 16. I decided to not give the user the option to `stream` here because even if it's set to false we'd still return an async iterable that yields only 1 result.

The quorum may never be reached so it's better to give a reault than fail entirely.

If people actually want the last (and best) value they can just use `it-last`:

```js
await last(ipfs.name.resolve('/ipns/QmHash'))
```

I will update the breaking change message to include this info.

alanshaw:
I'm going to back out of the `object.data` returning an async iterable change - this doesn't make sense.

src/name/resolve.js
hugomrdias: only if the resolve is done with the dht!.
it can be done by dnslink or other Routers like ipns over dns.

we have been discussing this stream option as the way to listen to secondary routers while already having a result to work with. Allowing instant results even if stalled and when a better result arrives do something, think service worker cache. 

forcing everyone to use iterator helpers feels weird, users don't always need streaming, users need a way to do streaming when they need to.

hugomrdias: Screenshot 2019-12-12 at 16 01 48
can't this be optional like go has in the cli ? "you need stream?" cool just pass stream: true and you get a stream.

src/dht/provide.js
```diff
@@ -26,13 +25,22 @@ module.exports = configure(({ ky }) => {
     })
 
     for await (let message of ndjson(toIterable(res.body))) {
+      // 3 = QueryError
```
dirkmc: Is it possible to extract some of this boilerplate
- `for await (let message of ndjson(toIterable(res.body)))`
- `if (message.Type === 3) `
- etc
into a function / class?

src/files/stat.js
```diff
@@ -27,6 +28,12 @@ module.exports = configure(({ ky }) => {
     }).json()
 
     res.WithLocality = res.WithLocality || false
-    return toCamel(res)
+    return toCoreInterface(toCamel(res))
   }
 })
+
+function toCoreInterface (entry) {
```
dirkmc: May be worth extracting this into a shared function also

src/name/resolve.js
alanshaw: If it's not streaming by default users will just use the non-streaming APIs and be frustrated when things take a long time to resolve and they get no feedback. We need to build APIs that encourage streaming and discourage buffering. There's no downside if a result is available immeidately and it is streamed to you but if there is the possibility it needs to be sourced from multiple peers or is large or may take a while then we need to stream the result to give users feedback asap, if only to inform them of progress.

alanshaw:
> @alanshaw after sleeping on it I am bit worried about removing buffered versions of commands like `cat`.
> 
> Instead of
> 
> ```js
> const data = await ipfs.cat(cid)
> ```
> 
> user now needs to ALWAYS wrap it with a helper:
> 
> ```js
> const data = await concat(ipfs.cat(cid))
> ```
> 
> Something feels off here 🤔

If they want to buffer data into memory then yes that's what they need to do. They definitely don't ALWAYS need to do this! I'd advise against doing so anyway unless you know how much data you're buffering and that it's coming from your local repo.

If we provide them, users will just use the non-streaming APIs and be frustrated when things take a long time to yield any results and they get no feedback. For big files, it's necessary to stream data and given the majority of people won't know the size of the file they're retrieving, and are trying to obtain it from a network of poorly connected peers we should provide an API that a) encourages them to do the right thing and not buffer content into memory regardless of the size of the file they're retrieving and b) reduces time to first byte, gives them visibility over progress and allows them to differentiate between the file being unavailable or it being large.

On a related point, there's likely to be built in helpers to async iterables soon https://github.com/tc39/proposal-iterator-helpers so you'd be able to `await` for collect or reduce.

src/files/ls.js
```diff
@@ -19,7 +19,7 @@ module.exports = configure(({ ky }) => {
     searchParams.set('arg', CID.isCID(path) ? `/ipfs/${path}` : path)
     searchParams.set('stream', true)
     if (options.cidBase) searchParams.set('cid-base', options.cidBase)
-    if (options.long != null) searchParams.set('long', options.long)
+    searchParams.set('long', options.long == null ? true : options.long)
```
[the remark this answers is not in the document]
alanshaw: In JS this information is actively available and we jump through hoops to remove it from the response (even though it's not really removed, it's just set to default values - empty string, zero etc.). I've removed it from the core interface so that we always get the information as I believe this option is actually useful for CLI formatting.

PR #1186, opened 2019-11-27
feat: support UnixFSv1.5 metadata

achingbrain:
Depends on:

- [x] https://github.com/ipfs/js-ipfs-utils/pull/19
- [x] https://github.com/ipfs/interface-js-ipfs-core/pull/572


src/add/form-data.js
```diff
@@ -12,6 +12,14 @@ exports.toFormData = async input => {
   let i = 0
 
   for await (const file of files) {
+    if (file.mtime) {
+      formData.append(`file-${i}-mtime`, file.mtime)
```
alanshaw: What if this is `dir-${i}`?

src/add/form-data.browser.js
```diff
@@ -9,6 +9,14 @@ exports.toFormData = async input => {
   let i = 0
 
   for await (const file of files) {
+    if (file.mtime) {
+      formData.append('mtime', file.mtime)
```
alanshaw: Need to identify which file/dir this metadata pertains to.

src/add/form-data.js
achingbrain: The field names aren't used when parsing multipart requests - if it's a dir we detect that from the content type.

src/add/form-data.browser.js
achingbrain: This has been refactored out.

achingbrain:
Adding interface tests are blocked by https://github.com/ipfs/js-ipfs/pull/2625

src/add/form-data.browser.js
```diff
@@ -18,9 +28,13 @@ exports.toFormData = async input => {
...
 
-      formData.append(`file-${i}`, new Blob(bufs, { type: 'application/octet-stream' }), encodeURIComponent(file.path))
+      formData.append(`file-${i}`, new Blob(bufs, { type: 'application/octet-stream' }), encodeURIComponent(file.path), {
+        header: headers
+      })
     } else {
-      formData.append(`dir-${i}`, new Blob([], { type: 'application/x-directory' }), encodeURIComponent(file.path))
+      formData.append(`dir-${i}`, new Blob([], { type: 'application/x-directory' }), encodeURIComponent(file.path), {
+        header: headers
+      })
```
alanshaw: I'm not sure this is working...

```js
const fd = new FormData
fd.append('test', new Blob(['data'], { type: 'application/octet-stream' }), 'path', { header: { mtime: Date.now(), mode: 'rwx' } })

const opts =  { method: 'post', body: fd }
const r0 = new Request('/test', opts)

await r0.text()
"-----------------------------21291924801818001399752183477

Content-Disposition: form-data; name=\"test\"; filename=\"path\"

Content-Type: application/octet-stream



data

-----------------------------21291924801818001399752183477--

"
```

src/add/form-data.browser.js
```diff
@@ -9,6 +9,16 @@ exports.toFormData = async input => {
...
 
   for await (const file of files) {
+    const headers = {}
+
+    if (file.mtime) {
+      headers.mtime = file.mtime
+    }
+
+    if (file.mode) {
+      headers.mode = file.mode
```
alanshaw: Missing?

```suggestion
      headers.mode = file.mode.toString(8).padStart(4, '0')
```

achingbrain:
I'm going to merge this to stop it getting stale - interface tests are in & skipped for go-IPFS, we can address any issues in further (smaller) PRs.

PR #1190, opened 2019-12-01
Update package.json's main prop

0x6431346e:
It looks like this package does not work on Node < 10 because of the [for...await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of) feature:

```
 /home/lion/Code/aragon/aragon-cli/packages/aragon-cli/node_modules/ipfs-http-client/src/files-regular/index.js:60
      for await (const entry of get(path, options)) {
          ^^^^^
 SyntaxError: Unexpected reserved word
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:616:28)
    at Module._compile (/home/lion/Code/aragon/aragon-cli/packages/aragon-cli/node_modules/pirates/lib/index.js:99:24)
    at Module._extensions..js (module.js:663:10)
    at extensions.(anonymous function) (/home/lion/Code/aragon/aragon-cli/packages/aragon-cli/node_modules/require-precompiled/index.js:16:3)
    at newLoader (/home/lion/Code/aragon/aragon-cli/packages/aragon-cli/node_modules/pirates/lib/index.js:104:7)
    at Object.require.extensions.(anonymous function) [as .js] (/home/lion/Code/aragon/aragon-cli/packages/aragon-cli/node_modules/ava/lib/worker/dependency-tracker.js:42:4)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
```

I think the issue here is that `main` points to `src` instead of `dist`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "src",
     "dist"
   ],
-  "main": "src/index.js",
+  "main": "dist/index.js",
   "browser": {
     "glob": false,
     "fs": false,


alanshaw:
We don't do any transpilation on the CDN version either. The `engines` field in the `package.json` is wrong and we need to update it to the version of Node.js that supports `for await`, which is [`10.3.0`](https://node.green/#ES2018-features-Asynchronous-Iterators-for-await-of-loops).

@0x6431346e would you be up for sending a PR?

0x6431346e:
> We don't do any transpilation on the CDN version either. The `engines` field in the `package.json` is wrong and we need to update it to the version of Node.js that supports `for await`, which is [`10.3.0`](https://node.green/#ES2018-features-Asynchronous-Iterators-for-await-of-loops).
> 
> @0x6431346e would you be up for sending a PR?

Fair enough, what's the point of the `dist` folder and the build script though? (curious)

alanshaw:
It is for use in the browser with a script tag.

https://github.com/ipfs/js-ipfs#use-in-the-browser

PR #1224, opened 2020-01-22
Handle case where ky responses have no body with a getter for a ReadableStream

pcowgill:
Closes https://github.com/ipfs/js-ipfs-http-client/issues/1215

@hugomrdias @alanshaw Let me know what you think! The tests are passing, I added a new test case, and I've run the lint script.


pcowgill:
The basic approach Hugo and I discussed for this PR was to pass the entire `ky` response to the helper lib that converts the response to an async iterable rather than just passing the `Response.body` and hoping that part is enough.

src/lib/stream-to-async-iterable.js
```diff
@@ -1,6 +1,19 @@
...
+
+  // An env where res.body getter for ReadableStream with getReader
+  // is not supported, for example in React Native
+  if (!body) {
+    if (res.arrayBuffer) {
+      return (async function * () {
+        const arrayBuffer = await res.arrayBuffer()
+        yield arrayBuffer
+      })()
+    }
```
achingbrain: Can you please handle the case where `body` but there's no `res.arrayBuffer` - e.g. throw an error with an informative message, otherwise execution will continue and the code will try to access `body[Symbol.asyncIterator]` and fail.

src/add/index.js
```diff
@@ -16,21 +16,36 @@ module.exports = configure(({ ky }) => {
     if (options.chunker) searchParams.set('chunker', options.chunker)
     if (options.cidVersion) searchParams.set('cid-version', options.cidVersion)
     if (options.cidBase) searchParams.set('cid-base', options.cidBase)
-    if (options.enableShardingExperiment != null) searchParams.set('enable-sharding-experiment', options.enableShardingExperiment)
+    if (options.enableShardingExperiment != null) {
```
achingbrain: Were these changes necessary?  I personally like curly brackets around if statement bodies, even one liners, but this has made everything inconsistent and is outside the scope of the intended change.

src/dht/put.js
```diff
@@ -16,8 +16,12 @@ module.exports = configure(({ ky }) => {
     const searchParams = new URLSearchParams(options.searchParams)
     if (options.verbose != null) searchParams.set('verbose', options.verbose)
 
-    key = Buffer.isBuffer(key) ? encodeBufferURIComponent(key) : encodeURIComponent(key)
-    value = Buffer.isBuffer(value) ? encodeBufferURIComponent(value) : encodeURIComponent(value)
+    key = Buffer.isBuffer(key)
```
achingbrain: Similar to the curly brackets that have been added, these just look like whitespace changes - please revert.

achingbrain (changes requested):
Thanks for this PR, it mostly looks good, a few changes needed mentioned in the comments.

pcowgill:
> Thanks for this PR, it mostly looks good, a few changes needed mentioned in the comments.

@achingbrain Thanks for the quick review! I'll work on addressing the comments now.

src/lib/stream-to-async-iterable.js
pcowgill: @achingbrain do you mean where body is undefined but there's no `res.arrayBuffer`? I'll assume that's what you meant.

Good call, we should handle that case. I'll add a test case for it too.

src/add/index.js
pcowgill: No, they weren't necessary. They happened due to the default code styling settings I have globally in VS Code. If this style is important and gates PR merging, though, we should modify the linter settings to throw a warning or an error with this style. Ideally, we would get `aegir lint --fix` to automatically resolve such warnings/errors too, which may happen out of the box when we enable the warning. I'll open an issue for this.

pcowgill: Just opened the issue https://github.com/ipfs/js-ipfs-http-client/issues/1225

src/lib/stream-to-async-iterable.js
pcowgill: @achingbrain do you mean where body is undefined but there's no `res.arrayBuffer`? I'll assume that's what you meant.

Good call, we should handle that case. I'll add a test case for it too.

src/dht/put.js
pcowgill: Will do!

src/add/index.js
pcowgill: I've disabled the `Prettier` and `JavaScript Standard Style` extensions for this workspace locally in VS Code so that automatic linting changes won't make it into a PR against this repo next time, but I still think the linting settings change is worth making to make it simpler for others to make PRs in the future.

src/lib/stream-to-async-iterable.js
pcowgill: Addressed in this commit https://github.com/pcowgill/js-ipfs-http-client/commit/74c286ed08862fc7775a2f5ba51a44248faebcef

src/lib/stream-to-async-iterable.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
+
+module.exports = function toAsyncIterable (res) {
+  const { body } = res
+
+  // An env where res.body getter for ReadableStream with getReader
+  // is not supported, for example in React Native
+  if (!body) {
+    if (res.arrayBuffer) {
```
pcowgill: @achingbrain Do we want to `console.log` anything in this case, like "falling back to a hackier approach since the fetch being used here doesn't implement the streams spec"?

src/add-from-url.js
```diff
@@ -1,19 +1,19 @@
 'use strict'
 
 const kyDefault = require('ky-universal').default
-const toIterable = require('./lib/stream-to-iterable')
+const toAsyncIterable = require('./lib/stream-to-async-iterable')
 
-module.exports = (config) => {
+module.exports = config => {
```
pcowgill: @achingbrain I assume you'd like me to revert this style change too, right?

src/add/index.js
```diff
@@ -2,12 +2,12 @@
...
 const ndjson = require('iterable-ndjson')
 const configure = require('../lib/configure')
-const toIterable = require('../lib/stream-to-iterable')
+const toAsyncIterable = require('../lib/stream-to-async-iterable')
 const { toFormData } = require('./form-data')
 const toCamel = require('../lib/object-to-camel')
 
 module.exports = configure(({ ky }) => {
-  return async function * add (input, options) {
+  return async function* add(input, options) {
```
pcowgill: I'll revert this whitespace change

src/add/index.js
```diff
@@ -52,7 +52,7 @@ module.exports = configure(({ ky }) => {
   }
 })
 
-function toCoreInterface ({ name, hash, size, mode, mtime }) {
+function toCoreInterface({ name, hash, size, mode, mtime }) {
```
pcowgill: @achingbrain The aegier linter also could use a rule for this

pcowgill:
@achingbrain Okay, I think I've addressed all of your comments! Thanks again for the quick review. Could you please give this another review when you have a moment?

I reverted all of the style changes manually, although it retrospect it may have been faster to configure some local workspace settings to match the styling conventions of this project. Longer term I think [this](https://github.com/ipfs/js-ipfs-http-client/issues/1225) would be a nice improvement to make.

src/add-from-url.js
pcowgill: Done

src/add/index.js
pcowgill: Done

pcowgill:
Yikes, it looks like there are a lot of merge conflicts now that another PR was merged. I'll work on resolving them.

pcowgill:
Looks like it was this PR https://github.com/ipfs/js-ipfs-http-client/pull/1183

pcowgill:
@achingbrain @hugomrdias I don't think getting the conflicts resolved gates an interim PR review, though, in case you have time. Thanks!

pcowgill:
@achingbrain I fixed the merge conflicts. We need to decide how we want to handle this issue - https://github.com/alanshaw/stream-to-it/issues/3#issuecomment-579471374 - before merging this, now that one of the previously modified files, `lib/stream-to-async-iterable`, was extracted into [this shared package](https://github.com/alanshaw/stream-to-it/blob/master/source.js)

pcowgill:
Some tests are failing with the new logic introduced in this PR, like this one:
```
  1) .files (the MFS API part)
       .add pins by default:
     Error: Neither Response.body nor Response.arrayBuffer is defined
```
Interestingly those tests weren't failing with the earlier version of these changes against version `41.x`.

pcowgill:
The test failure was because I did a find/replace relative to an earlier version `master`, so the changes were missing some new uses of `res.body` that needed to be changed to `res` in the latest version of `master`. The tests should be passing now 🤞

pcowgill:
It looks like [the Travis failure](https://travis-ci.com/ipfs/js-ipfs-http-client/jobs/281577864) is just it hitting a timeout, which I don't think is something that my PR would increase the probability of, perhaps something that's just stochastic?

pcowgill:
As for the `codecov` warnings, it looks like `codecov` never ran for https://github.com/ipfs/js-ipfs-http-client/pull/1183, so the difference may be from that PR

alanshaw:
@pcowgill apologies I didn't get to this sooner. Did you look into adding a global afterResponse hook for react native?

It might be less invasive to add a `.body` property to the response here with an async iterable that calls `.arrayBuffer()`. Then all the rest of the code can remain the same and it would be easier to remove when RN supports `.body`.

What do you think?

pcowgill:
@alanshaw To make sure I understand the suggestion, which file are you thinking that change would land in? Thanks!

pcowgill:
@alanshaw Curious for your thoughts on the `codecov` comment too!

alanshaw:
> @alanshaw To make sure I understand the suggestion, which file are you thinking that change would land in? Thanks!

https://github.com/ipfs/js-ipfs-http-client/blob/e9aaa750beeb000f485fb31d43d2af648676dc8e/src/lib/configure.js#L40-L42

elmariachi111:
Just checked out this PR because when I run jest on my setup (ionic / Typescript / React / node 12), and just test a simple `ipfs.cat` it fails with `Cannot read property 'destroy' of undefined`. This PR's fix works 99% around that issue by supporting `res.arrayBuffer` instead of `res.body`. But I think another cause / usage is hidden in `error-handler#21` -> this line fails first on my machine. When I hack into it with

`if ((isNode || isElectronMain) && response.body) response.body.destroy()`

my jest tests are green again :)

hugomrdias:
we will be doing some changes around http requests handling soon to fix these issues, stay tuned.

pcowgill:
> we will be doing some changes around http requests handling soon to fix these issues, stay tuned.

@hugomrdias That's great! Can you share any additional details about this? Thanks!

pcowgill:
> @pcowgill apologies I didn't get to this sooner. Did you look into adding a global [afterResponse](https://github.com/sindresorhus/ky#hooksafterresponse) hook when in react native?
> 
> It might be less invasive to add a `.body` property to the response here with an async iterable that calls `.arrayBuffer()`. Then all the rest of the code can remain the same and it would be easier to remove when RN supports `.body`.
> 
> What do you think?

@alanshaw Sure, that suggestion works for me.

@hugomrdias Are you already planning on implementing the changes Alan described in the child package of the `js-ipfs` monorepo as part of the `ky` removal, or should I make a PR against the `ky` removal branch with these changes? (Assuming the `ky` removal branch will take at least a few more days to be merged.)

pcowgill:
Will there still be the equivalent of an `afterResponse` hook after removing `ky`?

hugomrdias:
There will be no hooks and would advise against making the PR that specific problem needs to be fixed outside http-client

PR #1240, opened 2020-02-17
fix: return iterables from pin add and pin rm

achingbrain:
Still need to simulate taking async iterables as `source` arguments to `pin.add` and `pin.rm`.

Depends on:

- [ ] https://github.com/ipfs/interface-js-ipfs-core/pull/594


achingbrain:
Superseded by https://github.com/ipfs/js-ipfs/pull/2771

